PR #51, opened 2026-06-13
fix(bkn-safe): robust postgres + headless first-login on fresh install

sh00tg0a1:
## Why

A fresh `openbkn` + bkn-safe install failed every login with:

```
Device auth failed (401): {"error":"invalid_client","error_description":"... Unable to locate the table"}
```

Root-cause chain (diagnosed live on the broken install):

1. **Data loss.** The bundled hydra Postgres is a 1-replica **Deployment** on an RWO **local-path** PVC with the default **RollingUpdate** strategy. On restart/rollout, RollingUpdate surges a new pod before killing the old one; local-path enforces RWO **per-node**, not per-pod, so two `postmaster`s mount the same data dir → `data directory lock file is invalid` → the surviving Postgres comes up on a **pre-migration snapshot**.
2. **No self-heal.** hydra's migrate (init container) and the OAuth client seed (helm hook) are **one-shot** — both "succeeded" against the doomed instance, so hydra ended up with **0 tables + 0 clients** → `Unable to locate the table`, and stayed broken.
3. **First-login wall.** After repairing (1)/(2), headless login still hangs: the seeded admin has `MustChangePassword=true`, so the login provider returns the change-password **page** instead of accepting the hydra login → device-code flow times out (`Device login timed out`). Onboard always runs against a first login, so this hit every install.

## What

- **`deps.yaml`** — postgres `strategy: { type: Recreate }`. Old pod is torn down before the new one starts → never two postmasters. Behaviour-only; safe for existing installs (no PGDATA/volume change).
- **`onboard.sh`** — `onboard_bkn_safe_clear_must_change()`: before the bkn-safe credential login (both `-y` and interactive), bounce the password `pw → pw.heal → pw` through the tokenless `/api/safe/v1/auth/change-password` endpoint. Clears `must_change_password`, password unchanged. Non-2xx first hop (401 = operator already changed it) is a no-op → idempotent. Mirrors the existing test-user heal.

## Follow-ups (not in this PR)

- `seed.go`: skip `MustChangePassword` when the operator sets an explicit `config.initialPassword`.
- bkn-safe `doLogin`: return a machine-readable (JSON) signal for non-browser must-change so the CLI need not scrape HTML.
- `@openbkn/bkn-sdk` (separate repo): `auth login --new-password` to complete must-change in-flow.

## Test

- `helm template` renders `strategy: { type: Recreate }` on `*-postgres`.
- `bash -n` + shellcheck clean on `onboard.sh`.
- Verified live: re-migrate + re-seed clients + tokenless change-password → `openbkn auth login` succeeds end-to-end.

### DIFF
--- a/bkn-safe/charts/bkn-safe/templates/deps.yaml
+++ b/bkn-safe/charts/bkn-safe/templates/deps.yaml
@@ -26,6 +26,13 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   replicas: 1
+  # Single-writer DB on an RWO PVC. The default RollingUpdate would surge a new
+  # pod before terminating the old one; local-path enforces RWO per-NODE, not
+  # per-pod, so both pods can mount the same data dir at once → two postmasters →
+  # "data directory lock file is invalid" → corruption / silent data loss (which
+  # wipes hydra's schema + OAuth clients and breaks login). Recreate tears down
+  # the old pod before starting the new one, so there is never more than one.
+  strategy: { type: Recreate }
   selector: { matchLabels: { app: {{ .Release.Name }}-postgres } }
   template:
     metadata: { labels: { app: {{ .Release.Name }}-postgres } }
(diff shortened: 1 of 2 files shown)


leecky:
已补充投影载荷变更的部署约束：必须切换新版本索引并原子切 alias，不在同版本覆盖写入；同时明确 BKN Safe 直接授权读取 API 的部署顺序。

PR #242, opened 2026-07-17
fix(bkn-backend): ignore object type index config writes

rongwei-liu:
## Summary
- ignore client-supplied `DataProperty.IndexConfig` in object type create/update flows
- preserve existing index config metadata as read-only legacy data when updating object type properties
- stop object type index build/query capability logic from consuming object property index config
- update affected unit tests for validation, object type service, and object type worker behavior

## Verification
- `git diff --check`
- `I18N_MODE_UT=true go test ./driveradapters ./logics/object_type ./worker -gcflags=all=-l -count=1`

## Notes
No linked issue was found for this branch. This change keeps object type property index config out of the object type write path; index/vector build configuration should belong to data catalog resources.


rongwei-liu:
Closing per request. This PR was created by mistake and should not be submitted.

Issue #647, opened 2026-08-05
bug(skill-runtime): Smart Q&A cannot execute recalled skills (skills OT only enables recall; ExecuteSkill has no consumer)


lbb1656:
## Phenomenon

In Smart Q&A for a Business Knowledge Network (example: e-commerce decision KN), after creating a `skills` Object Type and recalling skill `local-message-sender`, the agent still cannot invoke the skill to perform the intended side effect (e.g. send a remote message).

Observed UI state:

- Ontology sidebar: Object Types > 0, **Action Types = 0**
- Agent reply (paraphrased): the KN contains skill `local-message-sender`, but **no Action Types are defined**, so the operation cannot be executed through the BKN framework
- Suggested workarounds from the agent: bind an Action Type to the skill, or call the remote API directly

User expectation: creating a `skills` Object Type and having the skill visible should be enough to **run** the skill from Smart Q&A.

## Root Cause

Skill **recall** and Skill **execution** are separate pipelines. Creating a `skills` Object Type only satisfies the recall contract; it does **not** create an executable bridge in BKN Smart Q&A.

### 1. Recall path (works when modeling contract is met)

`find_skills` (Context Loader) validates a fixed `skills` Object Type contract, then returns skill metadata for the agent. It does **not** execute anything.

Hard checks in `loadAndValidateSkillsContract`:

- `skills` Object Type must exist (`object_type_id` defaults to `"skills"`)
- Required data properties: `skill_id`, `name`
- Caller must pass `kn_id` + `object_type_id`
- Without RelationType binding between business Object Types and `skills`, object-type / instance recall returns empty (hint), but network-level / query paths may still surface skills

Relevant code:

- `adp/context-loader/agent-retrieval/server/logics/knfindskills/index.go` — `FindSkills`, `loadAndValidateSkillsContract`, `requiredSkillsDataProperties`
- `adp/context-loader/agent-retrieval/server/logics/knfindskills/recall_coordinator.go` — recall modes / binding hints
- Docs: `bkn-docs/.../context-loader/operations/issue-unknown-context-loader-工具使用指南.md` §6.5 (prerequisites)

Studio agent exposes `find_skills` as a schema/discovery tool only:

- `bkn-studio/src/modules/knowledge-network/services/agent-chat.service.ts` — `SCHEMA_TOOLS` includes `find_skills`, alongside `get_action_info`

### 2. BKN execution path used by Smart Q&A (Action Types)

[...]

PR #667, opened 2026-08-06
fix(context-loader,execution-factory): three fixes from VM verification — 400 on path traversal, propagate upstream 4xx, expose toggles in charts

sh00tg0a1:
End-to-end verification on the VM after #659 merged surfaced three problems. Each is fixed here.

## 1. Path traversal was rejected but answered 500

`read_skill_file` with `../../etc/passwd`: the path is genuinely blocked (nothing outside the bundle is read), but the response was `Public.InternalServerError`. `normalizeZipPath` returns a bare error, which the rest layer turns into an internal error — a caller (a model in particular) reads that as a server fault and retries, when in fact it passed a bad path.

The management-side reader has always answered 400; the consumer-side reader now matches.

## 2. context-loader dropped the upstream status code

All four outbound methods turned any execution-factory failure into 502/500. A downstream 4xx is a caller error (path outside the bundle, unknown skill, no permission) and has to reach the caller with its own code; only a connection failure or a downstream 5xx is a genuine upstream fault. Added `skillUpstreamError` to classify by code.

## 3. Two switches had no chart entries

`MCP_EXECUTE_SKILL_ENABLED` and `SKILL_INTERNAL_READ_AUTHZ` could only be set with `kubectl set env`, which means the feature could not actually be shipped — an installed deployment had no way to enable `execute_skill`, nor to move authorization to `enforce`.

- `agent-retrieval`: `mcp.executeSkillEnabled`, default `false`
- `agent-operator-integration`: `skill.internalReadAuthz`, default `shadow`

Both defaults match the code defaults, so existing installations keep their current behavior. Each template is guarded with `{{- if .Values.x }}`, so an older values file that lacks the key still renders.

## VM verification of #659's feature surface

Image `0.1.3-feat-mcp-skill-tools.20260806003400.shaab46cc0` on VM 10.211.55.4:

| Check | Result |
|---|---|
| `/mcp/info` catalogue | 22 tools, including `list_skills` / `get_skill_content` / `read_skill_file`; `execute_skill` **absent** (off by default); coexists with `query_metric` |
| `MCP_EXECUTE_SKILL_ENABLED=true` | `execute_skill` appears in the catalogue |
| `list_skills` | Returns real published skills (`skill_id` / `name` / `version` / `category`) |
| `get_skill_content` | SKILL.md body plus `files: [SKILL.md, pick_substitute.py]` |
| `read_skill_file` | `pick_substitute.py` body returned correctly; a text file served as `application/octet-stream` was **not** misclassified as binary — the UTF-8 fallback from #659 doing its job |
| `execute_skill` | Sandbox really executes: `work_dir=skills/<id>`, `exit_code=2` with stderr passed through when the script's own argument is missing, and it runs into the script once the argument is supplied |
| Path traversal | Blocked, but answered 500 → **items 1 and 2 of this PR** |
| `SKILL_INTERNAL_READ_AUTHZ=enforce` | admin still reads successfully; no false denial |

### An environment problem fixed along the way (not code in this PR)

Both minio replicas on the VM were in `ImagePullBackOff`: `swr.../minio:RELEASE.2025-09-07T16-13-09Z` is amd64-only and the VM is arm64 (`no match for platform in manifest`). The second hop for skill files — fetching the body from object storage — fails for that reason alone. Swapping in `m.daocloud.io/docker.io/minio/minio` at the same tag restored it. **This is an arm64 gap in the third-party images on swr and deserves its own issue.**

### Still unverified

- A single file above the 5 MiB limit returning 413 (no skill file that large exists in the environment)
- `enforce` returning 403 for an **unauthorized** account (only confirmed that admin is not falsely denied; no unprivileged account available)

## On #647

#659 plus this PR cover the server-side half of option 2 in that issue, but it cannot be closed yet: Smart Q&A goes through a hardcoded `CONTEXT_LOADER_OPS` list in bkn-studio, so the new tools do not exist there until they are added; and the "recall vs execute vs Action Type" guidance is still unwritten.


sh00tg0a1:
Both findings were real. Fixed in 166a80fc.

**1. Missing `mcp.executeSkillEnabled` in the agent-retrieval chart** — confirmed and worse than it looked: the edit had been made in the *wrong working tree* (the main checkout instead of the branch worktree), so it was never staged. `git show --stat` on the previous commit lists only the operator-integration chart. The reviewer's failure scenario is exactly right — the deployment template is a fixed env list with no `.Values.env` passthrough, so `kubectl set env` was the only way in and the next `helm upgrade` wipes it. The change is now on the branch, and `helm template` renders `MCP_EXECUTE_SKILL_ENABLED: "false"` by default. The stray edit has been reverted out of the main checkout.

**2. 413 falls back to `InternalServerError`** — confirmed. `errCodeMap` had no entry for 413, and `DefaultHTTPError` maps anything missing to `InternalServerError`; since the MCP layer hands the model `HTTPError.Error()` (the whole struct as JSON), the model reads `code` rather than the HTTP status. The 5 MiB guard in `operator_integration_skill.go` raises 413, so this was reachable — the model would have retried a file that never gets smaller.

Added `http.StatusRequestEntityTooLarge` and `http.StatusBadGateway` to `errCodeMap` with matching `desc.` / `sol.` entries in both locales. 502 was included for the reason given in the review: if `code` is what distinguishes "you passed something wrong" from "the upstream is down", both halves have to be present.

`TestDistinguishableErrorCodes` in `infra/errors/` now pins the five codes that a caller uses to decide whether to retry, so a future addition cannot silently collapse into `InternalServerError` again.

PR #671, opened 2026-08-06
feat(context-loader): make the execute-skill switch a real kill switch, not a tools/list filter

sh00tg0a1:
## The problem

`MCP_EXECUTE_SKILL_ENABLED` (added in #659) only gated the MCP tool surface. With it off:

- `tools/list` and `GET /mcp/info` omit `execute_skill` — correct
- `POST /api/agent-retrieval/{v1,in/v1}/kn/execute_skill` stayed registered and callable
- the toolbox `.adp` entry pointing at that route was registered unconditionally

The documentation describes this as "the one command-execution channel on the tool surface". With REST still open that sentence was false, and someone reading it to decide whether to enable the feature would misjudge the exposure. Authorization is still enforced per account by the execution factory (`execute` / `public_access`), so this was never an open door — but it was not the single gate the docs claimed.

## The change

The switch now gates route registration too. Turned off, `/kn/execute_skill` is a 404 rather than a registered handler that refuses: "this deployment has no skill-execution capability" holds at the routing table, not only in the tool catalogue.

The decision moved into `knskills.ExecuteEnabled`, so the MCP assembly, the `/mcp/info` catalogue and both routers read one source instead of three copies of the same condition.

## Naming

`MCP_EXECUTE_SKILL_ENABLED` was accurate while the switch was MCP-only; it is not now. Renamed to `EXECUTE_SKILL_ENABLED`, chart key `mcp.executeSkillEnabled` → `skill.executeEnabled`.

The old variable name is still accepted. A deployment that already set it must not silently lose a capability it had switched on — that is the failure this fallback exists to prevent, and `TestExecuteEnabledAcceptsLegacyEnvName` pins it.

## Tests

- `TestExecuteSkillRouteFollowsTheSwitch` — the route is absent when off, present when on
- `TestExecuteEnabledAcceptsLegacyEnvName` — the legacy variable still enables the feature
- `TestExecuteSkillOnlyAppearsWhenEnabled` (existing) — unchanged behaviour on the MCP side, now reading the shared decision

## Notes

The `.adp` toolbox entry is left as it is. It was raised as a third bypass, but that import path is on its way out, so pruning entries at generation time would be work spent on a mechanism being retired. Worth noting for whoever removes it: with the switch off, the toolbox entry now points at a 404.

Docs updated in `docs/api/context-loader/{mcp,skill}.yaml` and `schemas/README.md`: the switch is described as a kill switch covering both surfaces, and the REST endpoint no longer claims to be unaffected by it.


sh00tg0a1:
@claude review

PR #676, opened 2026-08-06
refactor(context-loader): stop auto-registering the built-in toolset into the execution factory

sh00tg0a1:
## Background

Context Loader's tools existed in two places: the MCP surface (16 core tools + 2 lifecycle tools, assembled in `driveradapters/mcp/assemble.go`) and a toolbox in the execution factory (two `.adp` packages upserted at startup). Two schemas for one set of tools — change one and they drift.

This PR removes the execution-factory registration.

## What changed

Startup registration removed:

- `server/bootstrap/` — `ToolDependencySync`, its tests, and the two embedded `.adp` packages
  - `context_loader_toolset.adp` — box "contextloader工具集", 16 tools
  - `execution_factory_tools.adp` — box "执行工厂工具集", 1 tool (`get_operator_schema`)
- `server/app/app.go` — the `go bootstrap.NewToolDependencySync().Start(...)` goroutine
- `drivenadapters/operator_integration.go` — `SyncToolDependencyPackage`, the multipart builder, and the `/internal-v1/impex/intcomp/import/toolbox` constant
- `interfaces` — `SyncToolDependencyPackageRequest` and the interface method (mock regenerated with `mockgen`)
- `config.go`, `agent-retrieval.yaml`, helm `values.yaml` + `configmap.yaml` — the `tool_dependency_sync` block

Docs updated to match the behaviour change:

- `help/{zh,en}/quick-start.md`, `help/{zh,en}/install.md` — no longer claim the toolbox is auto-imported at startup; point at `openbkn context info` instead
- `README.md`, `README.zh.md`, `deploy/README.md`, `deploy/README.zh.md`, `deploy/scripts/lib/onboard_apply_config.py`, `deploy/scripts/lib/onboard_report.sh`, `deploy/scripts/services/openbkn.sh` — same correction to the install-time wording. Also dropped `--skip-context-loader` from the four READMEs; that flag has no implementation in `deploy/onboard.sh` and was already a dead reference before this PR
- `docs/api/execution-factory/toolbox.yaml` — the list-response sample used the real built-in toolbox; replaced with a generic one (including its UUID)

## Not affected

- **The MCP surface.** Tools are assembled independently in `assemble.go`; `tools/list` still returns 16 core + 2 lifecycle tools.
- `GetToolDetail` and the MCP proxy calls into operator-integration stay — logical-property resolution still loads ToolBox tools.
- The `/impex/intcomp/import/toolbox` endpoint itself is untouched here; #678 removes it.

## One capability is deleted outright, not moved to MCP

`get_operator_schema` (the only tool in `execution_factory_tools.adp`, `GET /operator/market/{operator_id}` against operator-integration) has **no MCP equivalent** — a repo-wide grep finds it only in `docs/release` artifacts. Unlike the 16 Context Loader tools, this one is not "served over MCP instead"; it is removed. Calling it out so nobody later debugs its absence on the assumption that MCP covers it.

## Verification

- `go build ./...`, `go vet ./server/...`, `go test ./server/...` all pass
- `helm template` renders
- `gofmt` clean on every file this PR touches (three files in the repo were already unformatted at HEAD and are not touched here)

## Known impact

On environments that already ran an older build, the three built-in toolboxes stay in the execution factory's database — deleting the code does not remove existing rows. No cleanup script ships with this PR (deliberate).


sh00tg0a1:
## Response to review

### Blocker — confirmed, and it invalidates a premise of this PR

> Deleting the startup upsert leaves nothing registering the contextloader toolbox, and bkn-agent loads platform built-in tools only from the execution factory toolbox by `box_id`.

Verified against source. This is correct and I was wrong to describe the change as "capabilities are served over MCP instead":

- `infra/bkn-agent/app/core/tools.py:24-25` — `_mcp_connections`' own docstring says `type=mcp` is for *explicit external* MCP endpoints and that "平台内置工具不走这里（统一从执行工厂 toolbox 装载，见 load_tools）".
- `_toolbox_tools` in the same file loads only explicit `type=toolbox` refs and states there is no implicit default box.

So the MCP surface covers external clients (Cursor, `openbkn` CLI) but **not** the platform's own agent runtime. On a fresh install, a platform Agent would have no way to mount `search_schema` / `query_object_instance` / `query_instance_subgraph`.

The second failure path checks out too: `adp/context-loader/agent-retrieval/docs/release/agent-recall-examples/context_loader_agent_example.json` binds six tools to `tool_box_id: e521d454-4a0b-4dc9-8a28-d0986de1cef9`, so importing it into a fresh environment would resolve to a missing box.

This PR is therefore **not ready to merge**. Resolving it is a design decision — give bkn-agent a path to Context Loader's MCP surface, keep registering the contextloader toolbox and drop only the other two boxes, or seed the toolbox from an install step instead of a startup loop. Holding until that is settled.

### Documentation leftovers — fixed

`README.md:89`, `README.zh.md:90`, `deploy/README.md:133`, `deploy/README.zh.md:131` and the matching string in `onboard_apply_config.py` all still claimed `onboard.sh` imports the Context Loader toolset. Corrected to match `help/{zh,en}/install.md`. Also dropped `--skip-context-loader` from those four READMEs — it has no implementation in `deploy/onboard.sh` and was already a dead reference.

### `toolbox.yaml` example UUID — fixed

The sample was renamed to a generic toolbox but kept `e521d454-…`, the real contextloader box id, in two places. Both replaced with a random UUID.

### `get_operator_schema` — agreed, now called out separately

Correct that it has no MCP counterpart and is a net capability removal rather than a move to MCP. The PR description now says so explicitly instead of folding it in with the 16 Context Loader tools.

### Frozen schema coupling — acknowledged, not addressed here

The point that the internal REST endpoints behind those toolbox entries lose their only structural tie to the toolbox schema once the sync is gone is a real maintenance hazard. Whether it needs a comment in the `index.go` files depends on how the blocker above is resolved, so I am not adding one yet.

### `quick-start` wording

Fair: "you will not find Context Loader's tools in `/tool-box/list`" only holds for fresh installs; upgraded environments keep the existing rows. Will fix along with whatever resolves the blocker.

sh00tg0a1:
Two things on this PR.

**The blocker is real and unchanged — it is waiting on #685, not on a code change here.** Removing the startup registration does leave platform agents with no path to the retrieval tools, because `bkn-agent` could only consume them as an execution-factory toolbox keyed by `box_id`. #685 builds the missing half: `bkn-agent` now reaches Context Loader over its MCP surface, with token passthrough, the lifecycle handshake, and per-turn conversation continuity. Once that lands, this PR's failure scenario no longer holds. Nothing to change on this branch for it.

**Merge conflict resolved.** `main` had modified `context_loader_toolset.adp`, which this branch deletes. Kept the deletion — that is the point of the change — and merged `main` in.

One correction worth recording against your analysis, since it narrows the resolution space further in the direction you were pointing: you noted the 16-tool payload now exists nowhere in the repo, so a "seed at install time" resolution would have to recover it from git history. That is right, and it is a reason *not* to go that way. Re-seeding would restore the second source of truth this PR exists to remove — the same 16 tools described once in `assemble.go` and once in a `.adp`, drifting apart on the next schema change. The MCP surface is the single description; #685 makes the platform's own runtime consume it rather than a copy.

Merge order stays: #685 and this PR together, then #678 after the new images are deployed — #678 deletes the `intcomp` endpoints these two stop calling.

PR #677, opened 2026-08-06
refactor(bkn-agent): stop auto-registering published agents into the execution factory

sh00tg0a1:
## Background

An agent is invoked through this service's own `POST /api/bkn-agent/v1/invoke/{agent_id}`. On top of that, every published agent was also upserted as a tool inside the execution factory's "bkn_agent内置agent" toolbox — one capability described in two places, kept in sync by an infinitely-retrying startup loop.

Same decision as #676: each capability is served by its own native interface; nothing registers a second copy into the execution factory.

## What changed

- Removed `app/bootstrap/` (`toolbox_sync.py` + `__init__.py`)
- Removed the startup sync `start_startup_sync()` from the `main.py` lifespan
- Removed the post-change resync `schedule_resync()`: three call sites in `routers/agents.py` (create / update / delete) and one in `routers/impex.py`
- Removed config `TOOLBOX_SYNC_ENABLED`, `TOOLBOX_SYNC_RETRY_INITIAL_S`, `TOOLBOX_SYNC_RETRY_MAX_S`, `SELF_BASE_URL`, plus the matching `BKN_AGENT_TOOLBOX_SYNC` / `BKN_AGENT_SELF_BASE_URL` env vars and `runtime.toolboxSync` / `runtime.selfBaseUrl` chart values
- Tests: deleted `test_toolbox_sync.py`; removed the resync stubs and assertions from `test_impex.py`, `test_review_fixes.py`, `test_review_fixes2.py`
- Rewrote the README's "算子工厂注册" section

`OPERATOR_INTEGRATION_BASE` stays — the built-in `read_skill_file` tool and the skill surface still use it.

## Not affected

**Agent-as-tool still works.** It goes through `tool_refs` with `type: agent` (`_agent_tool` in `app/core/tools.py`) and never touched the execution factory. What this PR removes is only the path that made a published agent visible to *other* consumers of the factory's toolbox catalog.

## Deployment impact

`runtime.toolboxSync` was already `"false"` in the chart, so this sync was effectively off in the current deployment configuration; this PR deletes the code as well. The "bkn_agent内置agent" box remains in the factory database on existing environments — no cleanup script ships with this PR (deliberate).

## Verification

- `pytest`: 144 passed
- `helm lint` and `helm template` both pass


sh00tg0a1:
## Response to review

All five follow-ups were verified and acted on. `pytest` is now 148 passed (was 144).

### Lost `/invoke` coverage — confirmed and restored

The finding is correct and this was a real regression I introduced. `test_toolbox_sync.py` held four tests that had nothing to do with toolbox sync — `test_invoke_requires_identity`, `test_invoke_draft_agent_hidden`, `test_invoke_waits_for_terminal_state`, `test_agent_name_charset_enforced_at_api` — and deleting the whole file took them with it. Since `/invoke` is now the agent's only external calling surface, that left it with no behavioural test at all, and `test_contract.py` only compares `app.openapi()` against the frozen spec structurally, so it cannot cover the gap.

Moved all four into a new `app/test/test_invoke.py`, keeping the assertions intact: 401 without identity headers, draft agent answering 404 identically to a nonexistent one (no existence leak), synchronous wait to terminal state including the `session.expire_all()` constraint, and the name-charset rejection.

### Stale `/invoke` descriptions — fixed, spec re-exported

All four call it a "算子工厂 toolbox 回调": `README.md:33`, the `tasks.py` docstring, the `models.py` `InvokeRequest` docstring, and both places in the frozen `docs/api/bkn-agent/bkn-agent.yaml` derived from them. Rewrote the first three and re-ran `scripts/export_openapi.py`; `test_contract.py` passes, so implementation and spec are back in step.

### `models.py` name-pattern comment — fixed

Agreed the rationale no longer holds. The comment now says the charset is bkn-agent's own convention rather than an alignment with the factory's validator. Kept the pattern itself unchanged — relaxing it is a behaviour change that needs its own assessment of existing agents and import/export, not a drive-by in this PR.

Also removed the third reason in `AgentOut`'s "no revalidation on read" comment, which cited the startup-sync retry loop that no longer exists.

### `dao.list_published_agents` — removed

Confirmed dead: `toolbox_sync.sync_once` was its only caller.

### Out-of-repo values overrides

Noted. `runtime.toolboxSync` and `runtime.selfBaseUrl` will be silently ignored rather than failing an upgrade, since the chart has no `values.schema.json`. Anyone maintaining a deployment repo with overrides should drop those two keys — flagging it here since this repo cannot.

PR #678, opened 2026-08-06
refactor(execution-factory): remove the intcomp internal-component registration surface

sh00tg0a1:
## Background

The execution factory exposes four `intcomp` (internal component) registration endpoints. Counting actual callers:

| Endpoint | Callers |
|---|---|
| `POST /internal-v1/impex/intcomp/import/:type` | context-loader and bkn-agent startup sync |
| `POST /{internal-v1,v1}/operator/intcomp` | none |
| `POST /internal-v1/mcp/intcomp/{register,unregister/:mcp_id}` | none |
| `POST /{internal-v1,v1}/tool-box/intcomp` | none |

Only one of the four was ever called; the other three have had zero callers since they shipped. Once #676 and #677 stop those two startup syncs, the whole internal-component registration mechanism has no purpose left.

**Merge gate: #676 and #677 must be merged and deployed first.** Otherwise the old context-loader / bkn-agent images retry forever against a deleted route and log 404s.

## What was removed

**Routes** — `rest_private_handler.go`, `operator_handler.go` (internal and public), `mcp_handler.go`, `toolbox_handler.go` (internal and public)

**Handlers** — `driveradapters/operator/internal.go`, `driveradapters/toolbox/internal_toolbox.go`, `driveradapters/mcp/builtin.go`, plus the interface declarations in the three `index.go` files

**Logic** — `logics/operator/internal_operator.go`, `logics/toolbox/internal_tool.go` (and its test), `logics/mcp/mcp_builtin.go`, plus the `interfaces` method declarations and request/response types (`RegisterInternalOperatorReq`, `CreateInternalToolBoxReq`, `CreateInternalToolBoxResp`, `MCPBuiltin*`, `IMCPBuiltinService`)

**Internal-component config service narrowed** — `IIntCompConfigService` keeps only `DeleteConfig`, which clears the `t_internal_component_config` row when a component is deleted; existing internal components still need it. `CompareConfig` / `UpdateConfig`, their version-comparison helpers, and the now-callerless `dbaccess` `InsertConfig` / `UpdateConfig` all go with the registration path. `ConfigSourceType` and its constants become unreferenced once `IntCompConfig` is gone, so they go too. The `IntCompConfigSvc` injections in operator and mcp have no callers left and are dropped; toolbox keeps its own — `logics/toolbox/delete.go` still uses it.

**Incidental dead code** — `ValidatorIntCompVersion` (zero production callers after the deletion, only its own unit test)

**Middleware** — `middlewareBusinessDomain`'s `isBuiltin` parameter is constantly `false` once the four routes are gone. Parameter and branch removed; 24 call sites collapsed to the two-argument form. The branch only ever fired when `isPublic=false`, the deleted private intcomp import was the sole `isBuiltin=true` route, and the one remaining `isPublic=false` call site (`skill_handler.go:39`) already passed `false` — so the rewrite is behaviour-preserving line by line.

**Stale test** — `TestInternalImportRouteReuseImportHandler` registered `/impex/intcomp/import/:type` on its own router and then called it, so it would keep passing while guarding a route that no longer exists in production. The public surface is already covered by `TestImport` in the same file.

**Smoke-test helpers** — the whole of `tests/lib/mcp_internal.py`, plus the builtin helpers in `operator.py` and `tool_box.py` (no testcase in the repo calls them)

**Contract docs** — removed the corresponding paths and the `InternalToolBoxRequest` schema from `docs/api/execution-factory/{operator,toolbox}.yaml`; endpoint counts 91 → 89 (operator 15 → 14, toolbox 22 → 21), with `docs/api/README.md` kept in step. Also corrected the "internal-v1 has no Ingress" note in the chart's `values.yaml`, which listed intcomp registration as an example write endpoint.

## Kept on purpose

- **Public `/impex/import/:type`** — it shares the handler and `ImportConfigReq` with the deleted private route, and `is_internal` is read from the package body, so there is no branch to unwind; only the private route is removed.
- **The "internal operators/toolboxes may not be imported or updated" guards** in `logics/{operator,toolbox}/impex.go` — existing `is_internal` data still has to be rejected.
- **The `t_internal_component_config` table** — dropping a table is irreversible on the upgrade path. It stays; a separate issue can retire it in the usual three-phase way.
- **The `ErrExtInternalToolBoxVersion` error-code constant** — the error-code catalog is externally visible contract and does not follow the implementation out.

## Correction to an earlier claim

An earlier revision of this description said `TestRegisterRouterPreservesRouteSurface` guards this change. It does not. That test lives in `server/capabilitieslab/register_test.go:74` and its `expectedRoutes` enumerates the 35 routes of `/api/capabilities-lab/v1`, which do not overlap the operator / toolbox / mcp / impex routes deleted here. **agent-operator-integration has no equivalent route-surface guard test.** This PR is pure deletion so it does not add one, but nobody should assume a regression net exists.

## Contract impact

`POST /v1/operator/intcomp` and `POST /v1/tool-box/intcomp` are public contract in `docs/api`. There are zero callers inside this repo. Per the agreed decision we are not checking production access logs first — they are simply removed. If a consumer outside this repo (the frontend, say) does call them, it will get a 404 after merge.

## Known drift, not addressed here

`adp/execution-factory/operator-integration/docs/apis/` still lists the deleted endpoints. `docs/api/execution-factory/README.md:137` already declares that tree a historical draft that drifts from the implementation and must not be treated as a source of truth, so it is left alone.

## Verification

- `make test` (the CI unit-test target, which excludes `server/tests`) passes; `go vet ./server/...` is clean
- `gofmt` shows no regression from this change (`logics_mcp.go`, `logics_toolbox.go`, `validator_test.go` were already unformatted at HEAD and are not touched here)
- Every affected mock regenerated with `mockgen`
- Both OpenAPI documents parse with `yaml.safe_load`; endpoint counts recounted: toolbox 21, operator 14


sh00tg0a1:
## Response to review

### Blocker — agreed, this is the merge gate, not a code defect

> #676 and #677 are still OPEN; merging this first would break both services' startup sync.

Correct, and it matches what the description already declares. Nothing to change in the code — this PR stays blocked until #676 and #677 are merged **and** deployed. Note that #676 is itself now blocked on a confirmed design gap (bkn-agent has no path to Context Loader's MCP surface), so this gate will stay shut for a while.

### Dead code left by the deletion — fixed

All three verified and removed:

- `ConfigSourceType`, its constants and `String()` — unreferenced once `IntCompConfig` was deleted
- `dbaccess` `InsertConfig` / `UpdateConfig` — lost every caller along with the logic-layer `CompareConfig` / `UpdateConfig`; `SelectConfig` kept, `DeleteConfig` still uses it for its existence check. Model interface and mock narrowed to match
- `TestInternalImportRouteReuseImportHandler` — deleted. It registered the removed route on its own router, so it would have stayed green while guarding a path that no longer exists; `TestImport` in the same file covers the public surface

### `helm/values.yaml` comment — fixed

The internal-v1 security note listed "intcomp 注册" as an example write endpoint. Removed.

### `TestRegisterRouterPreservesRouteSurface` — you are right, my claim was wrong

Checked: it lives at `server/capabilitieslab/register_test.go:74` and its `expectedRoutes` enumerates the 35 routes of `/api/capabilities-lab/v1`, with no overlap with the operator / toolbox / mcp / impex routes deleted here. agent-operator-integration has no equivalent route-surface guard. The PR description now carries an explicit correction so nobody later assumes a regression net exists.

### `t_internal_component_config` orphan rows for internal MCP servers — accepted

Agreed with the analysis: `logics/mcp/manage.go:272` deletes internal MCP servers without checking `IsInternal`, so they stay deletable; only the config row is left behind. Nothing reads that table any more (`SelectConfig` serves only `DeleteConfig`'s own existence check), so this is frozen rather than inconsistent. If a read path is ever revived, those rows need handling first — worth carrying into the issue that retires the table.

### `docs/apis/` contract tree — deliberately not synced

`docs/api/execution-factory/README.md:137` already declares that directory a historical draft that drifts from the implementation and is not a source of truth. Leaving it. Now stated in the PR description so the choice is visible rather than looking like an oversight.

### Build verification the review could not run

For the record, since `go build` / `go vet` were unavailable in the review sandbox: both were run locally against this branch, along with `make test` (the CI unit-test target). All clean. Every affected mock was regenerated with `mockgen` rather than hand-edited.

PR #681, opened 2026-08-06
fix(claude-review): never finish a round silently, and show shard-level progress

sh00tg0a1:
Two problems, both about a review round that leaves the author with nothing to look at.

## 1. A round can finish silently

#671 was reported as "the review never triggered". It did trigger — on PR open, run [31069597979](https://github.com/openbkn-ai/bkn-foundry/actions/runs/31069597979), six jobs, all green:

```
plan / review (adp/context-loader) / review (misc) / review (contract) / verify / verdict
```

And it left **nothing at all on the PR**. All three shards finished without writing their findings JSON, `verdict` resolved to `action=none`, and the only trace was a `::warning::` buried in the run log.

```
No files were found with the provided path: findings/. No artifacts will be uploaded.
```

A later run on the same PR wrote them fine, so this is nondeterministic — and the entire pipeline's visible output hangs on the model remembering to write one file.

From the author's side a silent round and a round that never happened look identical. That is the failure mode sharding was meant to eliminate.

**Changes**

- **The findings JSON moves to the front of the 產出 section.** It used to come second, after inline comments — so a shard that wrapped up early left inline notes and nothing else. It is the only shard output that reaches the verdict; inline comments do not. The instruction now says so, and says why the ordering is deliberate.
- **Backfill a placeholder** when a shard finishes without writing one, so `verdict` can tell "ran but produced nothing" from "the job died".
- **Placeholders do not count as findings.** Without this an all-shards-silent round would `approve` a PR that was never reviewed — strictly worse than not voting. They are routed to 未覆盖的分片 instead.
- **When nothing was produced, post a plain PR comment** stating the round was empty, with a link to the run and how to request another. Deliberately not a review verdict: nothing was reviewed, so neither `approve` nor `request-changes` would be honest. Silence is the one outcome that must not happen.

## 2. The progress comment did not move

#680 added a sticky progress comment, but it was static — it appeared once and then nothing changed until the verdict landed minutes later. The live-updating comment people remember (spinner, ticking checklist) is `track_progress`, which is off because under sharding it posts one self-updating comment per shard and floods the PR.

Each shard now checks off its own line as it finishes, so the single comment actually moves:

```
### 评审进行中
- [x] `misc`
- [ ] `跨模块契约 / 配置与部署面 / 数据面`
```

Granularity is one update per shard, not per step. A shard that did not finish cleanly marks its line as not reviewed rather than silently checking it off, so a half-finished round cannot read as a complete one.

Two shards finishing at nearly the same instant can drop one update through a read-modify-write race; three retries, and the worst case is cosmetic. The step is `continue-on-error` — a broken progress display must not affect review output.

## Verification

| Input | Expected | Result |
|---|---|---|
| shard wrote nothing | placeholder created, valid JSON | ✅ |
| shard wrote a real file | placeholder must not overwrite | ✅ |
| all shards placeholder | `action=none`, no verdict | ✅ |
| one real + one placeholder | `approve`, placeholder under 未覆盖的分片 | ✅ |
| tick success + tick failure | rendered body correct, failure annotated | ✅ |

Ticking was exercised end to end against this PR (comment created, ticked, cleaned up). What cannot be verified here: this file lives under `.github/**`, and `issue_comment` runs always take the workflow definition from the default branch — the changes only take effect once merged.


sh00tg0a1:
All six findings were real and are fixed in the new commit. One of them — `gh api --paginate` without `--jq` — is a trap this file already documents at the prefetch step, and I reintroduced it two hundred lines away. Good catch.

- **Backfill depended on `Prepare findings dir`** (implicit `success()`): it now creates the directory itself.
- **`Report empty round` was gated on `action == 'none'`**, also an implicit `success()` — a crashed Assemble skipped both it and `Post review`, and the round went silent. Now gated on "no verdict was posted", which covers the crash path.
- **Assemble's parse guard only wrapped `json.loads`**: it now covers the whole loop body, and `lang` is coerced before `.strip()`.
- **The empty-round notice had no marker**: it now upserts on `<!-- claude-review-empty -->`, and a round that lands a verdict deletes the stale one.
- **`missing` created the no_output/absent distinction and dropped it in the output**: the reason is now rendered, since the two point at different fixes.
- **`Tick progress` used `job.status`**: replaced with the findings file itself, which is the ground truth and covers the missing-credential case you identified.

Verified locally: a malformed top-level array plus a placeholder plus a non-string `lang` assembles without crashing and separates the two not-covered reasons; backfill works with no pre-existing directory; ticking marks a placeholder shard as not reviewed and a real one as done.

/review

sh00tg0a1:
/review

New commit adds the animated spinner asset and drops the zh/en switch in favour of English-only PR-facing output. Please scope this round to that commit.

- Spinner GIF committed under `.github/assets/` rather than reusing `claude-code-action`'s user-attachment, which belongs to another project. Public repo, so the raw URL resolves through GitHub's image proxy; noted in a comment that a private repo would need the asset on user-attachments instead.
- Removed the `lang` field, the voting code and the duplicated zh/en templates. The switch made things worse rather than better: it chose the body skeleton by vote while shard labels, the progress comment, the empty-round notice and the backfill placeholder stayed hardcoded Chinese — so an English PR received a mixed document. The `contract` shard was literally named 「跨模块契约 / 配置与部署面 / 数据面」 and appeared under that name in every verdict.
- Every PR-facing string is now English. YAML comments and the prompt itself stay Chinese; neither is read by a PR reader.

Verified locally: the verdict assembles end to end in English, including both not-covered reasons, the truncation notice and both footers.

Worth stating explicitly — this round cannot exercise any of it. `issue_comment` resolves the workflow definition from the default branch, so what runs here is `main`'s version: the static progress comment from #680, without the spinner, the checkboxes or the English skeleton. Those only take effect once this merges.

sh00tg0a1:
Blocker confirmed and fixed, along with all four follow-ups.

**`Tick progress` had neither `GH_REPO` nor `GH_TOKEN` in scope.** The file has no workflow-level `env:`, the review job's `env:` does not define them, and only the sibling `Prefetch` step carries its own — so the step raised `KeyError` on its first statement and `continue-on-error` turned that into a green job with no trace. The checkboxes never moved and the "did not finish cleanly" annotation never rendered. Both variables are now declared on the step.

Your reading of why my end-to-end test missed it is exactly right, and worth recording: I ran it in a shell that already had `GH_REPO` and `GH_TOKEN` exported. Hand-testing a step lifted out of its job cannot see a missing job-level env var. That is a gap in how I verify this workflow, not a one-off.

- **Verify prompt still followed the PR's language for `reason`.** Fixed to unconditional English, along with the Chinese schema example next to it. You are right that this produced precisely the mixed document the commit claimed to remove.
- **`Report empty round` / `Clear stale empty-round notice` keyed on intent rather than outcome.** Both now key on `steps.post.outcome`, so a failing `gh pr review` also lands the notice instead of leaving the PR with the progress comment deleted and nothing in its place.
- **Malformed findings file was reported as "never ran, or died partway through"** — the one thing that case did not do. It now has its own reason.
- **`SPINNER` hardcoded `main`.** Now derived from `default_branch`.

One correction to my own commit message rather than to the code: you are right that this repo's PR titles are mostly Chinese (#664, #670, #672, #680), so "English by convention here" overstated it. English for PR-facing text is a standing instruction from the repo owner, not an inference from history. The change stands; the justification I wrote for it did not.

The two carried-over items you listed for status are both addressed in this commit.

/review

PR #685, opened 2026-08-06
feat(bkn-agent): reach Context Loader tools over its MCP face

sh00tg0a1:
## Why

Once #676 stops registering Context Loader's built-in toolbox into the execution factory, agents **inside the platform** lose their path to the knowledge-network retrieval tools. MCP only ever covered the other half — external clients (Cursor, `openbkn` CLI). bkn-agent loads tools from the execution factory toolbox, not over MCP. This PR builds the missing half.

## Three walls, all confirmed against source and on the VM

None is solved by "just configure a URL":

1. **Context Loader's MCP JSON-RPC lives on the public router only** (`rest_public_handler.go:109`), and that router mounts `middlewareIntrospectVerify` — it takes a real credential and does **not** accept the `/in` header identity. bkn-agent works on the `/in` convention and holds no token.
2. **Every business tool sits behind the lifecycle guard.** `bkn_context` (conversation_id + interaction_id) is a required input; without it the call returns `conversation_required`.
3. **Those ids cannot be minted locally.** A locally minted `int_...` is rejected with `resource_not_disclosed` — the MCP side derives its owner tuple from the credential. The supported path is `bkn_start_interaction` first, then reuse what it returns.

## What this does

- **`auth`** — capture the caller's `Authorization` verbatim into a ContextVar and forward it to Context Loader. This service still does not verify it; identity remains `x-account-id`. A ContextVar rather than a threaded parameter: there are a dozen `(account_id, account_type)` call sites between router and tool loading.
- **`config` / chart** — `CONTEXT_LOADER_MCP_URL`. **No service-credential fallback**: an AppKey fallback would make Context Loader see the key's issuer instead of the real caller, collapsing per-user authorization. A default-off switch like that eventually gets turned on to "make it work", and that is the last downgrade that should happen quietly. No token means no tools, plus a warning.
- **`core/context_loader`** — opening a session means handshake for ids, then load tools. `bkn_context` is injected at call time and **stripped from the model-visible schema**: when a turn starts is a runtime decision, not the model's, and a required id field just invites the model to invent one.
- **`ToolRef type=context_loader`** — deliberately carries no URL, so agent definitions stay portable across environments.
- **`evidence.begin_interaction`** accepts externally issued ids, so the evidence chain and Context Loader share one interaction. Agents without Context Loader tools keep minting locally — a lifecycle round-trip on every execution would turn an optional capability into a hard dependency.

## Four bugs this found, every one of them on the VM

Unit tests passed throughout. None of these was visible without running it.

**1. Token passthrough was silently dead.** `set_caller_token` sat in `get_account`, a **sync** dependency — FastAPI runs those in a threadpool, and a ContextVar set there never reaches the request coroutine. `caller_token()` was always `None`. The symptom was ugly: the task reported `succeeded` while the model, having no tools, **wrote a tool call as its answer**. Only the log showed `无可用凭据…跳过装载`. Now captured in the HTTP middleware, where `observability` already installs its TraceContext. Regression test included; reverting the fix reddens it at `assert None == 'Bearer probe-token'`.

**2. Handshake ids were parsed against a shape that does not exist.** langchain-mcp-adapters returns a 2-tuple `([{"type":"text","text":"<json>"}], {"structured_content": {...}})`. The parser only accepted a dict or a bare JSON string, so a **successful** handshake was judged "no usable id" and the whole tool load was skipped. The unit test used a simplified shape I had invented, so it passed. Tests now use the shape observed on the VM, plus the camelCase and text-block-only degradations.

**3. `bkn_finish_interaction` was called against the wrong contract** — `bkn_context` instead of `interaction_id`, and `outcome="succeeded"` which is not in the enum (`completed` / `failed` / `cancelled` / `handed_off`). Every turn's closure was rejected and interactions piled up in `active`. Four probes against the live MCP face separated the causes:

   | call | result |
   |---|---|
   | id wrapped in `bkn_context`, `outcome=succeeded` | `resource_not_disclosed` |
   | correct id, `outcome=succeeded` | `interaction_required: unsupported finish outcome` |
   | correct id, `outcome=completed`, no `answer` | `closure_manifest_invalid: completed outcome requires answer` |
   | full contract | success, `execution_status: completed` |

   `resource_not_disclosed` turns out to mean "no resolvable id" — the guard collapses "does not exist" and "not yours" into one code so it cannot be used to enumerate other people's ids.

**4. `outcome=completed` requires `answer`** (row 3 above), but closure happened in a `finally` block with no access to the turn's answer. Both paths now thread it through — runner captures the core return value, chat uses its accumulated `answer_parts`. A turn that produced no output closes as `failed` with a reason rather than falsely reporting `completed`.

Also widened: the connection-failure log printed only `ExceptionGroup.str()` — `"unhandled errors in a TaskGroup (1 sub-exception)"` — burying the real cause (a 401) in the sub-exception. It now unpacks them.

## Verification

182 unit tests pass. The contract spec is regenerated with `scripts/export_openapi.py`, not hand-edited. `helm template` renders.

**Multi-turn chat on the VM**, agent with `tools: [{"type": "context_loader"}]`, three turns on one `thread_id`, prompted so each turn *must* re-query rather than reuse history:

| check | result |
|---|---|
| tools actually mounted each turn | 3 turns, one `tool_call` event each |
| `interaction_in_progress` warnings | 0 |
| `Exception in ASGI` / `different Context` | 0 |

Single-turn `/invoke` also returns real data (the same five knowledge networks a direct MCP probe returns), so the tool path and the lifecycle closure are confirmed together.

### Coverage gap, stated plainly

Seven of the ten defects fixed in this PR were found only by running multi-turn traffic on a VM. Unit tests caught one (via CI), review caught two. None of the seven is a wrong branch — they are all "which context / which task / when does this run": a ContextVar set in a threadpool, a token reset across a generator context, cleanup parked in an async generator's `finally`, a session opened by code that had no responsibility to close it.

The regression tests I added pin each specific line. **They do not pin the pattern**, and this repo has no integration smoke that drives the Context Loader path across turns. A future change that parks cross-service cleanup in a generator `finally`, or opens a session without owning its closure, would pass CI. Worth a follow-up integration test (multi-turn + client disconnect + assert interaction terminal state); flagging rather than silently leaving it.

## Merge order

Independent of #676/#677/#678, but it is what makes #676 safe to land: without it, a fresh install has no path from a platform agent to the knowledge-network tools.

### DIFF
--- a/docs/api/bkn-agent/bkn-agent.yaml
+++ b/docs/api/bkn-agent/bkn-agent.yaml
@@ -809,10 +809,12 @@ components:
             - $ref: '#/components/schemas/McpToolRef'
             - $ref: '#/components/schemas/ToolboxToolRef'
             - $ref: '#/components/schemas/AgentToolRef'
+            - $ref: '#/components/schemas/ContextLoaderToolRef'
             discriminator:
               propertyName: type
               mapping:
                 agent: '#/components/schemas/AgentToolRef'
+                context_loader: '#/components/schemas/ContextLoaderToolRef'
                 mcp: '#/components/schemas/McpToolRef'
                 toolbox: '#/components/schemas/ToolboxToolRef'
           type: array
@@ -905,6 +907,35 @@ components:
       - agent_id
       - message
       title: ChatRequest
+    ContextLoaderToolRef:
+      properties:
+        name:
+          anyOf:
+          - type: string
+            maxLength: 100
+          - type: 'null'
+          title: Name
+        description:
+          anyOf:
+          - type: string
+            maxLength: 500
+          - type: 'null'
+          title: Description
+        type:
+          type: string
+          const: context_loader
+          title: Type
+      additionalProperties: true
+      type: object
+      required:
+      - type
+      title: ContextLoaderToolRef
+      description: 'Context Loader 的知识网络检索工具，经其 MCP 面装载。
+
+
+        刻意不带 url：端点来自 CONTEXT_LOADER_MCP_URL，agent 定义因此可以跨环境
+
+        导入导出而不带环境地址（type=mcp 写死 url 就不行）。'
     EffectivePromptOut:
       properties:
         source:
--- a/infra/bkn-agent/app/auth.py
+++ b/infra/bkn-agent/app/auth.py
@@ -25,7 +25,16 @@ def caller_token() -> str | None:
 
 
 def set_caller_token(value: str | None):
+    """由 HTTP 中间件调用（见 app/main.py）。
+
+    刻意不放在 get_account 里：那是个同步依赖，FastAPI 会丢进线程池执行，
+    在线程池里 set 的 ContextVar 回不到请求协程，读出来永远是 None。
+    """
     return _caller_token.set(value or None)
+
+
+def reset_caller_token(token) -> None:
+    _caller_token.reset(token)
 
 
 @dataclass(frozen=True)
@@ -37,10 +46,9 @@ class Account:
 def get_account(request: Request) -> Account:
     """/in 约定：网关信任请求头，鉴权押下游。空账户 fail-closed（本服务仅内部）。
 
-    另外把 Authorization 原样收进 ContextVar：本服务自身不校验它（身份仍以
-    x-account-id 为准），只在需要令牌的下游（Context Loader MCP 公开面）原样转发。
+    Authorization 不在这里收——见 set_caller_token 的说明，同步依赖跑在线程池里，
+    ContextVar 传不回来。收在 app/main.py 的中间件。
     """
-    set_caller_token(request.headers.get("authorization"))
     account_id = (request.headers.get("x-account-id") or "").strip()
     account_type = (request.headers.get("x-account-type") or "").strip()
     if not account_id or account_type not in _ALLOWED_TYPES:
--- a/infra/bkn-agent/app/auth.py
+++ b/infra/bkn-agent/app/auth.py
@@ -1,3 +1,4 @@
+from contextvars import ContextVar
 from dataclasses import dataclass
 
 from fastapi import Request
@@ -5,6 +6,26 @@ from fastapi import Request
 from app.errors import err
 
 _ALLOWED_TYPES = {"user", "app"}
+
+# 调用方原样带来的 Authorization 头。
+#
+# 存在这里而不是穿进每层函数签名：从路由到 runner 到工具装载有十几处
+# (account_id, account_type) 位置参数，再加一个会把整条链全改一遍；而
+# ContextVar 在 asyncio.create_task 时会随上下文复制，/run 的后台任务同样拿得到。
+#
+# 谁用它：Context Loader 的 MCP 面只认真实令牌（公开面挂 introspect），
+# 不吃 /in 那套头部身份，所以 type=context_loader 工具靠这个透传的令牌调用，
+# 保住 per-user 授权。见 app/core/context_loader.py。
+_caller_token: ContextVar[str | None] = ContextVar("bkn_agent_caller_token", default=None)
+
+
+def caller_token() -> str | None:
+    """当前请求透传进来的 Authorization 头原文（含 Bearer 前缀），没有则 None。"""
+    return _caller_token.get()
+
+
+def set_caller_token(value: str | None):
+    return _caller_token.set(value or None)
 
 
 @dataclass(frozen=True)
@@ -14,7 +35,12 @@ class Account:
 
 
 def get_account(request: Request) -> Account:
-    """/in 约定：网关信任请求头，鉴权押下游。空账户 fail-closed（本服务仅内部）。"""
+    """/in 约定：网关信任请求头，鉴权押下游。空账户 fail-closed（本服务仅内部）。
+
+    另外把 Authorization 原样收进 ContextVar：本服务自身不校验它（身份仍以
+    x-account-id 为准），只在需要令牌的下游（Context Loader MCP 公开面）原样转发。
+    """
+    set_caller_token(request.headers.get("authorization"))
     account_id = (request.headers.get("x-account-id") or "").strip()
     account_type = (request.headers.get("x-account-type") or "").strip()
     if not account_id or account_type not in _ALLOWED_TYPES:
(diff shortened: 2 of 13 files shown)


sh00tg0a1:
Both blockers confirmed against source and fixed in 31003043. Thanks — the first one is a genuine production hazard I would not have found by testing the happy path.

**Blocker 1 — client disconnect wedges the thread permanently.** Confirmed. `close_session` is an awaited network call and I had placed it *before* `_busy_threads.discard(thread_id)`. Its own guard is `except Exception`, which catches neither the `CancelledError` delivered on disconnect nor the `BaseExceptionGroup` the MCP TaskGroup raises — both propagate out of the `finally` and the discard never runs, leaving that `thread_id` at 409 until the pod restarts. Release and ContextVar reset now happen before any network call.

**Blocker 2 — every turn minted a fresh conversation.** Confirmed. `open_session` passed neither `conversation_id` nor the hint the server reads, so `ensureManagedConversation` fell through to `newManagedConversationKey()` on each turn. Rather than persist a thread→conversation mapping, chat now sends its `thread_id` as `X-OpenBKN-Host-Conversation-Key` (`host_lifecycle_hints.go:22`), which is exactly the anchor that surface exists for — stable across turns, no new state. `/run` and `/invoke` deliberately send no key: a one-shot execution *should* be its own conversation.

Also taken from this round:

- **Outcome labelling was wrong in both directions** — exactly as described. Replaced "answer is non-empty" with an explicit `_completed` flag set where `done` is yielded, and the structured-output payload is now captured into `_final_answer` so a structured turn is no longer archived as `failed`.
- **`set_current` without a token, and set even when `open_session` returned `None`** — fixed; the ContextVar is only set for a real session and is reset in the `finally`. That also removes the second question-less handshake through `tools.py`'s `current_session() or await open_session()`.
- **`CONTEXT_LOADER_MCP_TIMEOUT_S` was dead config** — now actually wired into the connection, which matters precisely because the teardown `await` had no upper bound.
- **Answer truncation** — the 4000-char cap now leaves a marker rather than silently diverging from what the user received.
- **Stale `config.py` comment** about "the two credential settings below" — gone with the AppKey.
- **`close_session` had no test at all** — added contract tests covering the flat `interaction_id`, `completed` ⇒ `answer`, the `_OUTCOMES` clamp, and that a rejected finish warns instead of raising. 178 passing.

One correction to the round-2 note about `_bind_context` dropping `response_format`: still unverified, and I am not claiming otherwise. It needs a real `content_and_artifact` tool to test against, which is why I have left it open rather than "fixed".

Not addressed, deliberately: the `name` / `description` / `additionalProperties` no-ops on `ContextLoaderToolRef`, and the missing "requires a forwarded Authorization" precondition in the spec. Both are doc-level and belong in the model docstring plus a re-export; I would rather do that in one pass than bury it in a fix commit.

Re-verifying on the VM now — in particular that a multi-turn chat lands in a single conversation, which is the part unit tests cannot prove.

sh00tg0a1:
Round-4 blocker confirmed and fixed in db89eee1.

**Setup failure leaks an active interaction.** Verified exactly as described: `open_session` succeeds at `graph.py:78`, and any raise between there and the `create_agent` call lands in the setup `except BaseException`, which released the thread and re-raised. `close_session` lives only in `_events()`'s `finally`, and `_events()` is never constructed when `stream_chat` raises — so the handshake's interaction stays `active` forever. Deterministic, and one leak per retry. `runner.py` genuinely does not have this hole, since its `load_tools` runs inside the try whose finally closes.

The fix closes the session in the setup handler, plus two things the fix itself needed:

- `cl_session` / `cl_token` are now bound **before** the try. A raise in `resolve_prompt` or `load_skills` happens before their assignment, so referencing them in `except` would have thrown `UnboundLocalError` and replaced the real failure with a useless one.
- A failure inside the closing call is warned, not propagated — the caller needs to see why setup failed, not why cleanup did. Thread release still precedes any network call, same reasoning as the previous round.

Regression test added; reverting the fix reddens it at `assert closed, "setup 失败后必须调 close_session"`. That path is unreachable on any normal execution, so a test is the only thing that holds it. 179 passing.

Also: `graph.py` had no module logger at all — caught by pyflakes, not by the suite, since the branch that needed it is never exercised in tests.

On the item your verification pass dropped (unvalidated `thread_id` forwarded as the host key): agreed with the refutation, and the residual effect you identified is real — a space or non-ASCII `thread_id` costs conversation grouping plus one wasted handshake, silently. Not fixing it in this commit, but it is worth a follow-up to constrain `thread_id` at the model rather than discovering it through lost continuity.

Two items remain deliberately open, unchanged from my last reply: the `_bind_context` / `response_format` question from round 2 (still unverified — it needs a real `content_and_artifact` tool, and I will not mark it fixed on reasoning alone), and the doc-level `ContextLoaderToolRef` gaps, which I would rather do in one pass than scatter across fix commits.

sh00tg0a1:
@claude 可以放行了吧

sh00tg0a1:
Round-6 blocker confirmed and fixed in 6c5cc6bf.

**agent-as-tool sub-agents got no Context Loader session.** Verified: `run_agent_once` short-circuits straight into `_run_agent_once_core` when `has_interaction()` is true — which is exactly the agent-as-tool path — so `open_session` is never called. When the parent agent does not itself declare `context_loader`, `current_session()` is `None` and the sub-agent answers with zero retrieval tools. Worse than the failure you described in one respect: because `open_session` never runs, its warning never fires either, so nothing appears in the logs at all.

This is a hole left by 1ff565f1, which removed the `load_tools` fallback handshake. That fallback did leak an unclosed interaction every turn and had to go — but it was also, incidentally, what covered this path. Removing it was right; backfilling was not done.

Fixed with explicit ownership in the short-circuit branch: inherit if a session is already current (no second handshake, and never close someone else's), otherwise open one and close it in the same frame, carrying `outcome` / `answer` per the finish contract. Two regression tests: one asserts the sub-agent both gets a session and closes its own, the other asserts an inherited session is never closed by the borrower. 186 passing.

---

On the second-order point raised in the #676 review — that `retrieval.completed` can never fire once these tools come over MCP — that one is real and I am **not** fixing it here. `_expected_fact_event_type` matches on execution-factory toolbox metadata (`metadata.path` against three `/in/v1/kn/...` endpoints, `metadata.server_url` host `agent-retrieval`), and MCP-sourced tools carry none of it. On a fresh install, retrieval provenance silently drops out of claim-source linking.

It was scope I identified when designing this path and did not build, not something this PR's implementation broke. Filed as #702 with the three candidate anchors and the reason I did not want to pick one inside a PR that has already grown well past its original shape. Flagging it here so the omission is on the record rather than discovered later.

sh00tg0a1:
@claude 可以放行吗

PR #688, opened 2026-08-06
fix(claude-review): the verdict body tripled after sharding; cut it back

sh00tg0a1:
## Measurement

Across 36 posted reviews on this repo:

| | median | mean | longest |
|---|---|---|---|
| before sharding | 3130 | 3231 | 4721 |
| after sharding | **9789** | 8815 | **15845** |

Breaking down one 9801-character verdict:

```
Blockers               2012   21%
Per-shard findings     3142   32%
Open questions         4639   47%
```

## Why

**Per-shard duplication.** Blast radius used to be written once for the whole PR. Now every shard writes its own `summary`, `impact` and `not_covered`, so five shards produce five paragraphs. Those two fields alone accounted for 20% of the body.

**Unbounded open questions.** Each shard contributes its own list with no cap, written as full arguments rather than conclusions. Some entries were carried over from earlier rounds purely for status, so the body grew with every round — the opposite of the convergence the re-review rules aim for.

The cost of this is not tokens, it is that a blocker sits buried in the middle of nine thousand characters.

## Changes

- **Prompt limits**: `summary` ≤ 2 sentences, `impact` one sentence and empty when there is nothing to say, `not_covered` one sentence, `unconfirmed` ≤ 2 entries of 1–2 sentences. Items already raised in an earlier round with no new evidence are not relisted for status — same family as the existing "do not relist what the author declined" rule.
- **Per-shard section collapses to one line per shard** with a `(N blocking, M open)` tag. Blast radius and coverage gaps move into a `<details>` block.
- **Open questions capped** at 2 per shard and 6 overall, with an explicit line stating how many were trimmed. A silent cap would read as "that was everything".
- **Hard clipping in the assembly script.** The model may ignore the length instructions; the script already treats findings as untrusted input everywhere else, and length is no different.

## Result on identical inputs

| | before | after |
|---|---|---|
| normal round, five shards, mostly clean | several thousand | **525** |
| worst case, every field at its cap | 18918 | **5727** |

## Verification

Ran the assembly script against both fixtures. Also confirmed the clean-shard path renders `_(no issues)_` rather than an empty block, and that the trimmed-count line appears only when entries were actually dropped.


sh00tg0a1:
@claude review

PR #695, opened 2026-08-06
fix(knn): make vector search reachable on resource-backed object types

sh00tg0a1:
## Problem

`knn` was unusable on every object type bound to a Vega table resource, even when the resource had a vector index fully built. Three layers each blocked it independently, so fixing any one of them alone changed nothing:

1. **bkn-backend** never advertised `knn` in `condition_operations`. #686 deliberately withheld it because the two layers below would have rejected the operator anyway.
2. **ontology-query** required the property's declared type to literally be `vector`. A table resource's source column is a string, so the check rejected every field: `condition [knn] left field is not a vector field: stadium_name:string`.
3. **vega-backend** built its filter field table from the resource schema alone. The vectors live in a field the build task generates into the local index and never writes back to the resource, so the condition failed with `condition [knn_vector] left field 'stadium_name_vector' not found`.

## What changed

**bkn-backend — publish what the query side needs.** The vector capability derivation from #686 now also carries the generated vector field name and the embedding model. The model recorded on the resource may be a name (`openbkn vega dataset build --embedding-model text-embedding-v4`) while the query side vectorizes by ID, so it is resolved through the model factory once per object type and cached. The result is written to the property's `index_config.vector_config` and `knn` is advertised. If the model cannot be resolved, the capability is withheld rather than advertised — an operator that is guaranteed to 400 is worse than no operator.

**ontology-query — accept a scalar property that carries a vector index.** `rewriteKnnCond` now targets `index_config.vector_config.vector_field` when the property is not itself vector-typed. The existing data-view path, where the property really is a vector, is unchanged.

**vega-backend — recognise index-only fields when querying the local index.** Generated vector fields are added to the filter field table for queries served from a local index. They are still not published on the resource schema: they are part of the index implementation, not columns of the data source.

The `_vector` naming rule now lives in one place in vega's interfaces. The build task generates it, the query side recognises it, and BKN forwards it with the schema — all three reference the same source instead of each re-deriving the suffix.

## Verification

Unit suites green for all three services, with new tests covering capability derivation (generated field name, per-field model overriding the resource default, nothing without a built index), the `knn` rewrite on a scalar property, rejection when no vector index exists, and the generated-field helper.

End-to-end on the dev VM against a resource built with `--fulltext-fields stadium_name,city_name,country_name --embedding-fields stadium_name`:

- Object-type detail reports `knn` on `stadium_name` only, with `vector_field: stadium_name_vector` and a resolved model ID; the two fulltext-only fields keep `match` without `knn`.
- `knn` for `famous stadium in Mexico` returns Estadio Azteca first, then four more Mexican stadiums — none of which share a literal token with the query.
- The decisive check: `阿兹特克体育场` (Chinese, zero lexical overlap with the Latin-script data) returns `Estadio Azteca` at 0.7157 via `knn`, while the same string through `match` returns nothing. Cross-language recall is something only the vector path can do.
- No regression on full text: capability registration, instance recall and the schema-only default all match their pre-change results.

## Cost of the capability, and how it is bounded

Turning `knn` on made semantic instance retrieval issue a vector condition for every vector field of every candidate object type, each independently vectorising the same query string. The world-cup query went from roughly 0.5s to 4.4s. Three changes bring it back without giving up the capability:

- **ontology-query deduplicates vectorisation within a request**, keyed by (model, text). The same query string under the same model always produces the same vector; previously each sub-condition paid for it again. Failures are memoised too, so one failing embedding is not retried once per sub-condition.
- **One vector condition per object type** (`max_knn_sub_conditions_per_type`, default 1). Issuing `knn` against several text fields of the same row adds little recall and costs linearly.
- **Vector conditions only for the highest-ranked object types** (`knn_object_type_limit`, default 3). Concept retrieval has already ordered them; the tail cannot reach the final result set, so vectorising for it is pure cost.

Object types past the limit still get their full-text conditions — they are narrowed, not dropped. `enable_knn_instance_retrieval` turns the vector path off entirely for deployments that want full text only.

Measured on the VM, with the score of the top row used as proof that `knn` actually fires (10.669 with a vector clause, 9.957 for full text alone):

| state | latency | top score |
|---|---|---|
| vector clauses for every candidate object type | 8.8s | 10.669 |
| **defaults above** | **1.0s** | **10.669** |

An earlier revision of this PR reported 0.6s, but that measurement was worthless: a missing default left `enable_knn_instance_retrieval` nil, so no vector clause was issued at all and the top score had silently fallen back to 9.957. Review caught it; the defaults are now set in `DefaultSemanticInstanceRetrievalConfig` and merged in `mergeSemanticInstanceRetrievalConfig`, and every measurement above is checked against the score.

Two further defects surfaced while re-measuring:

- The budget was allocated by list position, but concept retrieval returns object types in the knowledge network's natural order whenever relation types are hit — not by relevance. It now ranks by the concept score, carried on an internal field so response ordering is untouched.
- Ranking alone was not enough: coarse recall only scores the object types it matched, so types pulled in as relation endpoints score 0 and filled the top slots without being able to issue `knn` at all. Allocation now skips object types that have no vector-capable field.

Full text is unaffected — capability registration, instance recall, the schema-only default and the direct `knn` query for `阿兹特克体育场` all match their pre-change results, with zero ontology-query 400s.


sh00tg0a1:
## 本 PR 衍生的后续项

排查与联调过程中发现、但不属于本 PR 范围的问题，集中列在这里：

| Issue | 内容 | 状态 |
|---|---|---|
| #704 | `get_kn_detail` / `get_object_types` / `search_schema` 默认路径都不返回 `condition_operations`，Agent 在 MCP 面无从判断字段能否 `match` / `knn` | 待做，建议紧接本 PR |
| #722 | `knn` 无向量索引时由 vega 内部降级，阻塞点是降级的可见性 | 待决策，见下 |
| #708 | 每个对象类只发一个向量条件这个默认值缺召回率实验支撑 | 待验证 |
| #235 | 旧 issue，前提（「KNN 左值必须是 vector 类型」）已被本 PR 推翻，建议收敛到「下游 400 被包装成 `CommonExternalServerError`」这一条 | 已说明 |
| #713 | 解析下沉到 vega + 降级策略 | 前半随本 PR 完成，后半移交 #722 |
| #706 | 能力派生读资源当前配置而非构建任务快照 | **本 PR 已解决**：模型取自产出索引的构建任务快照 |
| #707 | bkn-backend 读对象类详情新增模型工厂依赖 | **本 PR 已解决**：不再解析模型 |
| #151 | data_view 退役的残留：`NewMatchCond` / `NewKnnCond` 仍自读 `index_config`，是该字段仅存的读者 | 已在该 issue 记录确切位置 |

其中 #722 需要产品决策：降级会把「这个字段没建向量索引」变成「查询成功、零结果」，而现有响应链没有承载说明的位置，硬加要动 vega → ontology-query → context-loader → MCP 四层契约。本 PR 维持明确报错，错误中带字段名与资源名。

data_view 退役若在推进中，第 1、2 项（删除 data_view 分支、回收对象类属性级 `index_config`）可以并入；若那边范围不合适，另立 issue 也可以。

kenlee1988:
发现 1 个阻塞问题，建议 **Request changes**。

- **P1：回归已有的原生 vector 属性查询。** [`[condition_knn.go:144](https://github.com/openbkn-ai/bkn-foundry/blob/0361a290d511d05f4a0c756c9f8501d9cf7dbcc9/adp/bkn/ontology-query/server/common/condition/condition_knn.go#L144)`](https://github.com/openbkn-ai/bkn-foundry/blob/0361a290d511d05f4a0c756c9f8501d9cf7dbcc9/adp/bkn/ontology-query/server/common/condition/condition_knn.go#L144) 现在无条件把查询文本交给 VEGA；但 VEGA 对字符串值会强制要求 `LocalIndexName` 和对应构建任务快照。此前，资源映射到一个实际 `vector` 类型属性、且属性自带 `index_config.vector_config.model_id` 的既有路径，会在 ontology-query 中直接向量化并把向量交给 VEGA。因此没有本地构建索引（或快照中没有该 feature）的原生 vector 资源查询现在稳定返回 400。应仅对“标量属性 + Vega 生成向量字段”的新路径下放向量化；保留实际 vector 属性原来的 ontology-query 向量化分支，并补回归测试。

已运行并通过相关包测试：ontology-query、context-loader、VEGA、bkn-backend；`git diff --check` 通过。未修改工作区。

PR #696, opened 2026-08-06
fix(trace): expose interaction artifacts in read model

leecky:
## Summary
- expose interaction question/result artifact references in the canonical read model
- keep request details scoped to the operation while lazily loading full interaction context
- avoid treating interaction results without claim IDs as governed claims

## Verification
- GOCACHE=/private/tmp/openbkn-go-cache go test ./... -count=1
- make check-swag


leecky:
@claude review一下。

PR #716, opened 2026-08-07
feat(observability): disclose durable log source degradation

leecky:
## Summary
- persist Collector log-loss coverage by source and deployment in the BKN Trace MariaDB Core
- monitor the Collector metrics endpoint outside the Log Query request path
- disclose degraded coverage as partial results instead of overwriting it to healthy after OpenSearch succeeds
- keep monitoring disabled unless a persistent Core store and internal endpoint are configured

## Related
- #547
- #545
- docs: openbkn-ai/bkn-docs#53

## Verification
- Complete Go suite passed in bkn-trace/agent-observability: go test ./...
- Helm rendered with monitoring disabled and enabled configuration
- MariaDB integration test compiles under the integration build tag; execution requires BKN_TRACE_TEST_MARIADB_DSN

## Remaining release validation
- Local fault injection: force exporter failure or queue overflow, then validate degraded disclosure, Collector restart persistence, and two-sample recovery.

### DIFF
--- a/bkn-trace/agent-observability/charts/agent-observability/templates/deployment.yaml
+++ b/bkn-trace/agent-observability/charts/agent-observability/templates/deployment.yaml
@@ -125,6 +125,16 @@ spec:
               value: {{ .Values.observability.sourceTimeout | quote }}
             - name: BKN_OBSERVABILITY_MAX_CONCURRENT_SOURCES
               value: {{ .Values.observability.maxConcurrentSources | quote }}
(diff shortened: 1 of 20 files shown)


leecky:
Addressed all three review findings in the latest commit: coverage-store read failures now return available but partial results; the first Collector sample is baseline-only; recovery requires two no-loss deltas and no longer depends on a shared exporter queue. Focused regressions passed. /review

Issue #730, opened 2026-08-07
fix(bkn-trace): remove duplicate Evidence index Helm Hook dependency


leecky:
## Description

BKN Trace Chart optionally renders `evidence-index-setup`, a Helm Hook Job using external image `docker.io/curlimages/curl:8.10.1` to create the Evidence OpenSearch index. In restricted or offline delivery environments the node cannot reach Docker Hub, causing installation to block although `agent-observability` already initializes Evidence and Artifact indexes at runtime through `EnsureIndex`.

The Hook and runtime initializer duplicate responsibility and their mappings can drift.

## Steps to reproduce
1. Enable `evidence.indexManagement.enabled=true` and `evidence.indexManagement.createJob.enabled=true`.
2. Install into an environment without Docker Hub egress.
3. Observe `agent-observability-evidence-index-setup` fail to pull `curlimages/curl`.

## Expected behavior

BKN Trace installation has no dependency on `curlimages/curl`; runtime `EnsureIndex` is the only Evidence/Artifact index initialization path.

## Actual behavior

The optional Helm Hook creates a second initialization path and can block installation because its external image is not part of the OpenBKN delivery image set.

## Environment
- OpenBKN: 0.1.3 mainline
- Deployment: Kubernetes / restricted Docker Hub egress

## Acceptance Criteria (Definition of Done)
- [ ] Remove `evidence.indexManagement`, Evidence index ConfigMap, `evidence-index-setup` Hook Job, `curlimages/curl` image configuration and outdated documentation.
- [ ] Product install profile no longer enables or rewrites the deleted Hook image.
- [ ] Helm render regression proves neither default nor explicit legacy values render the Hook or reference `curlimages/curl`.
- [ ] Runtime `EnsureIndex` tests continue to cover Evidence and Artifact index creation/update behavior.
- [ ] No Evidence data/API/schema migration and no deployment is performed by this change.
- [ ] CI green.

## Boundaries / Risk

Do not change Evidence document structure, runtime mapping semantics, APIs, lifecycle persistence or existing OpenSearch indexes. Existing OpenSearch credentials still need index create/mapping permissions for runtime initialization. This Issue removes installation-time duplication only; deployment and data changes require a separate owner confirmation.

## Additional context

- Runtime Evidence initialization: `bkn-trace/agent-observability/src/drivenadapter/httpaccess/opensearchevidencestore/store.go`
[...]

PR #731, opened 2026-08-07
fix(bkn-trace): remove duplicate Evidence index Hook

leecky:
## Summary\n- remove the duplicate Evidence index Helm Hook and its curlimages dependency\n- keep runtime EnsureIndex as the sole index initialization path\n- remove obsolete installer, offline sync, and documentation paths\n\n## Verification\n- bash scripts/test_evidence_index_governance.sh\n- bash deploy/scripts/services/openbkn_trace_profile_test.sh\n- bash deploy/scripts/services/openbkn_trace_prerequisites_test.sh\n- helm lint charts/agent-observability\n- go test ./src/drivenadapter/httpaccess/opensearchevidencestore ./src/infra/opensearch\n\nCloses #730


leecky:
Fixed the stale HOOK_IMAGES references in offline image sync and moved the existing 5s Evidence refresh interval into runtime EnsureIndex for new indexes. Added regression coverage for both behaviors. /review

PR #619, opened 2026-08-03
feat(vega): 支持 SQL Server 作为数据源

kenlee1988:
# Pull Request

## What Changed

Brief description of changes:

- [x] Add the SQL Server table connector, configuration validation, type mapping, and metadata discovery
- [x] Add Resource Data and Raw Query support with T-SQL validation and SQL Server-specific paging
- [x] Add connector registration, schema migration, runtime availability reporting, API schema updates, and regression tests

---

## Why

- Related Issue: Closes #500
- Related Feature: SQL Server data source
- Background: VEGA previously exposed SQL Server in user-facing documentation but lacked a working local connector and query path. Connector registrations may also be shared by same-version binary distributions with different built-in connector implementations, so an unavailable local implementation must not prevent startup or be reported as usable by the API.

---

## How

- Key implementation points:
  - Use `go-mssqldb` through `database/sql`, including IPv6-safe connection string construction and an options allowlist
  - Discover accessible schemas, tables, views, columns, indexes, primary keys, and foreign keys from SQL Server system catalogs
  - Route raw SQL execution and paging through the table connector interface
  - Validate T-SQL as read-only before connecting and reject unsupported statements and paging forms
  - Keep SQL Server identifier quoting and date-condition handling dialect-specific
  - Treat a registered local connector that is absent from the running binary as unavailable: log and skip it during startup instead of terminating the service
  - Return the runtime-computed `available` field from connector-type detail/list APIs and support optional `available=true|false` list filtering before pagination
  - Keep persisted `enabled` state independent from runtime `available` capability
- Breaking changes (API, config, database, etc.):
  - Adds `sqlserver` as a connector type and `tsql` as a Raw Query dialect
  - Adds the SQL Server connector registration migration under schema version 0.1.3
  - Adds the required `available` boolean to connector-type responses and the optional `available` list query parameter; existing request fields and endpoint semantics are unchanged

---

## Testing

- [x] Unit tests passed locally
- [ ] Integration tests passed locally
- [x] Verified in test environment

Test notes:

- `make ci` passed after the runtime availability changes
- OpenAPI lint passed for `docs/api/vega/connector-type.yaml`
- Manually verified Catalog connection, metadata discovery, and data preview against a real SQL Server instance
- Automated SQL Server container/TLS integration coverage remains follow-up work

---

## Risk & Rollback

- Potential risks:
  - SQL Server-specific metadata permissions and uncommon T-SQL constructs may behave differently across deployments
[...]

PR #732, opened 2026-08-07
perf(claude-review): stop re-reviewing pushes after an approval

sh00tg0a1:
## The waste

This week, 12 of 68 verdict rounds were approvals following an earlier approval on the same PR — 18% of all rounds, roughly \$80–100.

```
#683   C A A A A A
#695   C C A A A A
#691   A A A
```

The first round catches something, and every push after that re-runs the entire review to repeat the same conclusion.

## The change

Once a PR is approved, pushes no longer trigger a round on their own. `/review` asks for one, and the approval body already carries that instruction.

**Two exceptions, both deliberate:**

- New commits touching `migrations/`, `helm/`, `values.yaml` or `bkn-safe/` are reviewed anyway. That is where one line breaks an environment — and `main`'s ruleset has `dismiss_stale_reviews_on_push: false`, so the approval survives the push. Skipping there would leave neither the gate nor the reviewer looking at the change.
- `CHANGES_REQUESTED` is unaffected. While the author is fixing things is precisely when re-review earns its cost.

Nothing is posted when a round is skipped — a comment on every push would be the noise this removes. The skip is recorded as a `::notice::` in the run log.

## Verification

The planner was run against real PRs with the event type varied:

| PR | event | skip | why |
|---|---|---|---|
| #695 | `pull_request` | **true** | last verdict approved, nothing risky touched |
| #695 | `issue_comment` | false | an explicit request always runs |
| #685 | `pull_request` | false | last verdict was `CHANGES_REQUESTED` |
| #619 | `pull_request` | false | touches `migrations/` |

### DIFF
--- a/.github/workflows/automation-claude-review.yml
+++ b/.github/workflows/automation-claude-review.yml
@@ -143,6 +143,7 @@ jobs:
       pr_number: ${{ steps.meta.outputs.pr_number }}
       shards: ${{ steps.plan.outputs.shards }}
       has_shards: ${{ steps.plan.outputs.has_shards }}
+      skip_round: ${{ steps.plan.outputs.skip_round }}
       prior_rounds: ${{ steps.plan.outputs.prior_rounds }}
       file_total: ${{ steps.plan.outputs.file_total }}
       file_listed: ${{ steps.plan.outputs.file_listed }}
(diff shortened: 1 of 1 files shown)


sh00tg0a1:
Blocker confirmed and fixed, along with both follow-ups.

**The notice was corrupting `$GITHUB_OUTPUT`.** The planner's stdout is redirected wholesale into the output file, so `print("::notice::…")` wrote a line with no `=`, which the runner rejects as `Invalid format`. Your reading of the blast radius is exactly right and worth restating: it fired *only* when `skip` was true, so the failure was reserved for the path this PR exists to create — every already-approved PR receiving a push would have ended up with a failed `plan`, no review, and no verdict. The notice now goes to a file and is emitted by a shell line after the heredoc. Verified across all four paths: every line reaching `$GITHUB_OUTPUT` contains `=`.

**`risky` scanned a truncated list.** `gh pr view --json files` caps at 100 and the file count is already reconciled a few lines above for that reason, which makes the omission worse rather than better. Truncation now blocks the skip outright — reading "nothing risky" off an incomplete list is not a judgement worth making.

**The stale progress comment.** Skipping meant `verdict` never ran, so its cleanup never ran either. The skip path now clears the marker itself.

/review

PR #767, opened 2026-08-09
fix(action-logs): stop the execution-log list from dragging back per-instance results

sh00tg0a1:
## Problem

Per-instance execution results and the task summary live in the same OpenSearch document. `QueryExecutions` applied no `_source` trimming, so every page of the action-log list read, deserialized and returned the full execution detail of every historical task it touched.

Measured on the reporter's test data: **31.4 seconds** for one list request, ~36,642 lines of formatted JSON.

Part of #724 (the list-payload half). The aggregated-execution half and the summary/result storage split follow in separate PRs.

## Change

Exclude the heavy fields at the OpenSearch query layer:

```go
"_source": map[string]any{
    "excludes": []string{"results", "action_type_snapshot", "action_source"},
},
```

Why at the query layer rather than nulling `exec.Results` after the read: the latter only shrinks the response body, while the document read and deserialization cost stays. Query-time trimming also covers historical documents that already carry embedded results, so no data migration is needed.

`action_type_snapshot` and `action_source` are per-execution config copies that no list consumer reads — the studio list mapper ignores them and context-loader's `slimActionExecution` strips them explicitly. `dynamic_params` stays, since context-loader keeps it in the slimmed list for the agent.

## Blast radius

Both callers were checked:

- `duplicate_check.go` only reads `Entries[0].ID` — unaffected.
- `action_execution_handler.go` feeds the REST list response — this is the intended target.

The detail endpoint uses a different code path and still returns `results`. The studio list mapper (`mapExecutionLog`) never read `results`, so the frontend needs no change.

## Testing

`Test_QueryExecutions_ExcludesResultsFromSource` captures the query body actually sent to OpenSearch and asserts the exclude list. It fails on the pre-fix code (`Expected: true Actual: false`) and passes after.

```
ok  ontology-query/logics/action_logs       2.537s
ok  ontology-query/logics/action_scheduler  3.138s
ok  ontology-query/driveradapters           4.081s
```

## Pre-merge checklist

- [x] Reproducing test written before the fix
- [x] Package tests green
- [ ] Human review

Refs #724

### DIFF
--- a/adp/bkn/ontology-query/server/logics/action_logs/action_logs_service.go
+++ b/adp/bkn/ontology-query/server/logics/action_logs/action_logs_service.go
@@ -305,6 +305,16 @@ func (s *actionLogsService) QueryExecutions(ctx context.Context, query *interfac
 		},
 		"from": offset,
 		"size": limit,
+		// The list only carries execution summaries. Per-instance results live in the same
+		// document and grow with the instance count; reading them here made a single page
+		// drag every historical task's full execution detail across the wire.
+		// action_type_snapshot and action_source are per-execution config copies that no list
+		// consumer reads (the studio list mapper ignores them, context-loader strips them).
+		// Excluded at the OpenSearch level so old documents carrying these fields are covered too.
+		// dynamic_params stays: context-loader keeps it in the slimmed list for the agent.
+		"_source": map[string]any{
+			"excludes": []string{"results", "action_type_snapshot", "action_source"},
+		},
 		"sort": []map[string]any{
 			{"start_time": map[string]any{"order": "desc"}},
 			{"id": map[string]any{"order": "asc"}},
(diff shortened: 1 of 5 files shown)


sh00tg0a1:
Fixed the leftover in dc16049.

**`list_action_executions` returned `"results": null`.** Correct catch — excluding `results` upstream left the agent-facing list with a null-valued key, which is worse than the original problem: the agent cannot tell "no per-object results" from "results not queried". The list projection now deletes the key outright, so the shape is a clean summary.

`docs/api/context-loader/action.yaml` gained an `ActionExecutionSummary` schema for list entries, stating that `results` is absent and that per-object detail comes from `get_action_execution`. `TestListActionExecutions_Success` now feeds a `results` array in the upstream response and asserts the key is absent (not null) downstream.

**On the studio frontend (flagged as unverified):** checked it — `src/modules/knowledge-network/services/mappers/action-execution.mapper.ts`. The list mapper `mapExecutionLog` reads only `action_type_id`, `action_type_name`, `duration_ms`, `failed_count`, `id`, `start_time`, `status`, `success_count`, `total_count`, `trigger_type`. It never touches `results`, `action_source` or `action_type_snapshot` — only `mapActionTypeExecutionLogDetail` reads `results`, and the detail endpoint is untouched. So the frontend was receiving ~36k lines per page and discarding them. No frontend change needed.

/review

PR #769, opened 2026-08-09
fix(action-scheduler): invoke the tool once when action parameters are instance-independent

sh00tg0a1:
## Problem

When an agent aggregates several query results into one message and hands it to an action type, the scheduler still iterates over the matched object instances and invokes the tool once per instance, so the same aggregated message is submitted repeatedly.

Reported case: 10 orders folded into one text body, the action condition matched 7 instances, `/api/messages` received **7 identical messages**. The execution log read `completed, total 7, success 7` — technically successful, semantically the opposite of the single summary push that was asked for.

Part of #724 (the duplicate-execution half). The list-payload half is #767; the summary/result storage split follows separately.

## Root cause

The parameter model and the execution granularity disagree. `dynamic_params` is one set per execution, yet the per-instance loop feeds it into every instance's parameters again.

That yields a statically decidable rule: if no action parameter has `value_from=property`, all N instances resolve a byte-identical parameter set — only `const` values and the execution-level `dynamic_params` feed it — so per-instance invocation is N repetitions of the same side effect, not N distinct actions.

## Change

Execution granularity is resolved from the parameter sources:

| Parameters contain `property` | Mode | Behaviour |
|---|---|---|
| yes | `per_instance` | one invocation per matched instance — byte-for-byte the previous behaviour |
| no | `once` | one invocation; the matched instances are recorded in the result's `targets` |

Counting is split into two dimensions so `once` does not surface as "1 of 7 succeeded":

- `target_count` — matched instances
- `total_count` — tool invocations (equals `target_count` under `per_instance`, always 1 under `once`)

`success_count` / `failed_count` count invocations. Under `per_instance` all three existing fields keep their current values, so stored data and the studio list are unaffected.

The tool/MCP dispatch is extracted into `invokeActionSource`, shared by both paths. It returns the parameters actually submitted, so the recorded `parameters` still reflect the real payload (the MCP branch merges `dynamic_params` a second time).

## Compatibility

Action types whose parameters are all `input`/`const` change from N invocations to 1. This is intentional: those N calls carry identical payloads, which is a repeated side effect rather than configured intent. `execution_mode` is persisted on the execution record so the resolved granularity is visible in the log.

Action types with a `property` parameter are untouched.

## Testing

- `Test_resolveExecutionMode` — property / input+const / no parameters
- `Test_executeAsync_AggregatedInvokesToolOnce` — 7 instances, aggregated params, `ExecuteTool` asserted `Times(1)`, one result carrying 7 targets, counts 1/1/0
- `Test_executeAsync_PerInstanceStillFansOut` — 3 instances with a property parameter still fan out to 3 invocations with distinct payloads, `targets` absent

```
ok  ontology-query/logics                    1.710s
ok  ontology-query/logics/action_logs        2.345s
ok  ontology-query/logics/action_scheduler   1.096s
ok  ontology-query/logics/action_type        3.618s
ok  ontology-query/logics/knowledge_network  4.746s
ok  ontology-query/logics/metric             2.980s
ok  ontology-query/logics/object_type        5.398s
ok  ontology-query/driveradapters            6.072s
```

## Pre-merge checklist

- [x] Package tests green
- [x] REST contract (`ontology-query.yaml`) updated for `execution_mode`, `target_count`, `targets` and the new `total_count` semantics
- [ ] Human review

Refs #724

### DIFF
--- a/adp/bkn/ontology-query/server/interfaces/action_execution.go
+++ b/adp/bkn/ontology-query/server/interfaces/action_execution.go
@@ -29,6 +29,17 @@ const (
 	TriggerTypeScheduled = "scheduled"
 )
 
+// Execution mode constants.
+//
+// ExecutionModeOnce means the action parameters do not depend on any instance property,
+// so all matched instances share one identical parameter set and the tool is invoked once.
+// ExecutionModePerInstance means at least one parameter reads an instance property, so the
+// tool is invoked once per matched instance with that instance's own parameters.
+const (
+	ExecutionModeOnce        = "once"
+	ExecutionModePerInstance = "per_instance"
+)
+
 // Action source type constants
 const (
 	ActionSourceTypeTool = "tool"
@@ -69,9 +80,11 @@ type ActionExecution struct {
 	ActionSourceType     string                  `json:"action_source_type"` // "tool" | "mcp"
 	ActionSource         ActionSource            `json:"action_source"`
 	ObjectTypeID         string                  `json:"object_type_id"`
-	TriggerType          string                  `json:"trigger_type"` // "manual" | "scheduled"
-	Status               string                  `json:"status"`       // "pending" | "running" | "completed" | "failed"
-	TotalCount           int                     `json:"total_count"`
+	TriggerType          string                  `json:"trigger_type"`             // "manual" | "scheduled"
+	Status               string                  `json:"status"`                   // "pending" | "running" | "completed" | "failed"
+	ExecutionMode        string                  `json:"execution_mode,omitempty"` // "once" | "per_instance", resolved from the action parameters
+	TargetCount          int                     `json:"target_count,omitempty"`   // matched instances; equals TotalCount under per_instance
+	TotalCount           int                     `json:"total_count"`              // tool invocations planned for this execution
 	SuccessCount         int                     `json:"success_count"`
 	FailedCount          int                     `json:"failed_count"`
 	Results              []ObjectExecutionResult `json:"results"`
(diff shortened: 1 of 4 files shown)


sh00tg0a1:
Blocking item fixed in 5204eb8.

**`executeOnce` ignored cancellation.** Confirmed and fixed. The per-instance loop checks for cancellation every batch; the aggregated path had no guard at all and then wrote a terminal status unconditionally, overwriting `cancelled`.

Two checks now:

- Before the invocation is issued: if the execution is already cancelled, no tool call is made, the result is recorded as `cancelled`, and the terminal status stays `cancelled`.
- After the invocation returns: if a cancel landed while the call was in flight, the side effect already happened, so the result is still recorded — but the terminal status stays `cancelled` rather than overwriting the user's cancel with `completed`.

Terminal-record writing is extracted into `finishOnce`, shared by both exit paths.

New tests: `Test_executeAsync_AggregatedCancelledBeforeInvocation` asserts `ExecuteTool` is called `Times(0)` and the status is `cancelled`; `Test_executeAsync_AggregatedCancelledDuringInvocation` uses ordered mocks (running, then cancelled) to assert one invocation, a recorded success result, and a `cancelled` terminal status.

## On the two non-blocking items

**`total_count` semantics.** Checked the studio mapper (`action-execution.mapper.ts`): the list reads `total_count`, `success_count`, `failed_count` and does not read `target_count`. Under `once` it will render 1 / 1 / 0, which is internally consistent ("one invocation, succeeded") rather than misleading. The alternative — keeping `total_count` as the matched-instance count — would render 7 / 1 / 0 and read as "6 missing". Historical records are untouched, and each record is self-consistent with its own `execution_mode`. This is a deliberate trade-off and it is called out in issue #724 for the reporter to confirm; if strict backward compatibility is required, the fallback is defaulting to `per_instance` and requiring an explicit `once`.

**`maxExecutionObjects` still counts matched instances.** Intentional. Under `once` the matched instances are all recorded in the result's `targets`, so the instance count still drives memory use during the scan and the size of the stored execution document. Gating on invocation count would let a 20k-instance aggregate through and produce one oversized result document — the exact failure mode the storage half of #724 is meant to remove. Worth revisiting once summary and per-instance results are split into separate indices.

/review

PR #772, opened 2026-08-09
fix(context-loader): surface execution mode and covered instances to the agent

sh00tg0a1:
Follow-up to #769, raised by the review on #771. Both findings there were non-blocking but real, and neither `main` nor `release/0.1.3` has them fixed — this PR fixes `main`, and a backport follows.

## The agent could not see what an aggregated run covered

#769 added `execution_mode` and `target_count` to the execution record, plus `targets` on the result. Context-loader's slimming projection never allowed those three through, so via `get_action_execution` the agent saw `total_count: 1` and an empty `_instance_identity` — reading "30 instances collapsed into one invocation" as "only 1 object was processed". The one thing the change most needed to communicate was dropped at the last hop.

`actionExecutionKeepKeys` now passes `execution_mode` and `target_count`; `actionResultKeepKeys` passes `targets`.

## Contracts had drifted from behaviour

- `docs/api/context-loader/action.yaml` — `total_count` still read "目标对象总数" in **both** `ActionExecution` and `ActionExecutionSummary`; it is now documented as the tool-invocation count, with `execution_mode` and `target_count` added and `target_count` explicitly called out as the field to read for coverage. `ActionExecutionResult` gains `targets` and explains that `once` produces a single result.
- `docs/api/ontology-query/ontology-query.yaml` — `ActionExecutionList.entries` still pointed at `ActionExecution`, which carries `results`, `action_type_snapshot` and `action_source`; the list has excluded all three at the OpenSearch layer since #767. Entries now reference a new `ActionExecutionSummary`.

## Testing

`TestGetActionExecution_Success` now feeds an aggregated execution (mode `once`, 30 targets, one result carrying 2 targets) and asserts all three fields survive the projection.

```
ok  context-loader/agent-retrieval/logics/knactionrecall  2.167s
```

Both OpenAPI files parse.

Refs #724

### DIFF
--- a/adp/context-loader/agent-retrieval/server/logics/knactionrecall/execute_action.go
+++ b/adp/context-loader/agent-retrieval/server/logics/knactionrecall/execute_action.go
@@ -60,9 +60,12 @@ func (s *knActionRecallServiceImpl) GetActionExecution(ctx context.Context, req 
 }
 
 // actionExecutionKeepKeys 是单次执行详情中对 Agent 有用、需保留的顶层字段。
+// execution_mode/target_count 必须留：once 模式下 total_count 是工具调用次数（恒为 1），
+// 没有这两个字段，Agent 会把「30 个实例合并成 1 次调用」误读成「只处理了 1 个对象」。
 var actionExecutionKeepKeys = []string{
 	"id", "kn_id", "action_type_id", "action_type_name",
-	"status", "trigger_type", "total_count", "success_count", "failed_count",
+	"status", "trigger_type", "execution_mode", "target_count",
+	"total_count", "success_count", "failed_count",
 	"start_time", "end_time", "duration_ms", "dynamic_params", "results",
 }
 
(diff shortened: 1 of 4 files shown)


sh00tg0a1:
Both items addressed in 911a9b7.

**Blocking: `targets` had no cap.** Confirmed. Under `once` there is exactly one result, so `results_limit` trims the number of results but never reaches inside one — and a scan without `_instance_identities` can match up to `ACTION_EXECUTION_MAX_OBJECTS` (10000 by default). Passing that straight through a projection whose whole purpose is to cut tokens was backwards.

The projection now truncates to 20 targets and adds `targets_total` and `targets_truncated` when it does. The full coverage number was never in `targets` anyway — it is `target_count` on the outer object — so the agent still gets an exact count plus a sample of what the invocation hit.

`TestGetActionExecution_TargetsCapped` feeds 5000 targets and asserts 20 survive, `targets_total` is 5000, `targets_truncated` is true, and `target_count` still reports 5000.

**Non-blocking: legacy records.** Also a real trap — records written before this feature have neither `execution_mode` nor `target_count` (both `omitempty`), and for them `total_count` *is* the object count, so "read `target_count`, not `total_count`" would leave the agent with no coverage number at all. Both `target_count` descriptions now state the rule explicitly: when `execution_mode` is absent, treat the record as `per_instance` and read coverage from `total_count`.

```
ok  context-loader/agent-retrieval/logics/knactionrecall  4.341s
```

/review

PR #782, opened 2026-08-10
fix(context-loader): filter list_resources by knowledge network

sh00tg0a1:
Closes #781

## Problem

`list_resources` only ever exposed the account-wide resource pool with `catalog_id / type / offset / limit`. It had no knowledge-network dimension, so a caller asking "which tables does this network have" had to pull one page of the pool and intersect it with the ontology bindings itself. Once the pool is large, that page never contains the bound tables.

Measured:

| Environment | Pool size | Bound tables | Hits at `limit=50` | `limit=200` | `limit=1000` | `limit=-1` |
|---|---|---|---|---|---|---|
| 14.103.77.23 | 21540 | 14 | 0 | 0 | 0 | 14 |
| VM (10.211.55.4) | 262 | 28 | 1 | 3 | — | 28 |

The hit count rising monotonically with `limit` is the fingerprint of the page-then-intersect approach. Agents that receive the empty list answer "this network has no data tables" and the whole turn is wasted.

Raising `limit` is not a workaround either: vega caps it at `MAX_LIMIT = 1000` and rejects anything larger, so the only alternative is `limit=-1`, which drags the entire 21540-row pool over the wire on every question.

## Change

`list_resources` accepts `kn_id`. When present it resolves the network's bindings instead of paging:

1. `GetKnowledgeNetworkDetail(kn_id)` to collect `object_types[].data_source`
2. fetch each bound resource by id via `GetResource`
3. project to the existing `ResourceLite` shape

This deliberately does **not** go through vega's list endpoint. That endpoint is the same pagination, so reusing it would move the caller's bug into the backend and would additionally inherit the unstable ordering from #779. Fetching by id is independent of pool size and needs no vega change.

Behaviour details:

- Bindings are a few dozen tables by nature, so the `kn_id` path returns everything at once. `catalog_id` / `offset` / `limit` are ignored when `kn_id` is set; `type` still filters the network's resources.
- Multiple object types sharing one physical table is normal, so resources are deduplicated. `total_count` is the deduplicated resource count, not the object-type count.
- Resources are fetched with a concurrency cap of 8; serial fetches reached seconds on real networks.
- Behaviour without `kn_id` is untouched, and the new response fields are `omitempty` so existing payloads are byte-identical.

### Unresolvable bindings are classified, not skipped

Three distinct failures used to be indistinguishable from "no resources". They now come back separately, each carrying `object_type_id`:

| Field | Condition | What the caller must do |
|---|---|---|
[...]

PR #785, opened 2026-08-10
fix(context-loader): backport list_resources knowledge-network filter to release/0.1.3

sh00tg0a1:
Backport of #782 to `release/0.1.3`. Cherry-picked cleanly — the touched files are identical between `main` and `release/0.1.3`. Build and tests (including `-race`) re-run on this branch.

This is the branch the customer's 0.1.3 deployment ships from, so it is the one that has to carry the fix.

Refs #781

---

## Problem

`list_resources` only ever exposed the account-wide resource pool with `catalog_id / type / offset / limit`. It had no knowledge-network dimension, so a caller asking "which tables does this network have" had to pull one page of the pool and intersect it with the ontology bindings itself. Once the pool is large, that page never contains the bound tables.

Measured:

| Environment | Pool size | Bound tables | Hits at `limit=50` | `limit=200` | `limit=1000` | `limit=-1` |
|---|---|---|---|---|---|---|
| 14.103.77.23 | 21540 | 14 | 0 | 0 | 0 | 14 |
| VM (10.211.55.4) | 262 | 28 | 1 | 3 | — | 28 |

The hit count rising monotonically with `limit` is the fingerprint of the page-then-intersect approach. Agents that receive the empty list answer "this network has no data tables" and the whole turn is wasted.

Raising `limit` is not a workaround either: vega caps it at `MAX_LIMIT = 1000` and rejects anything larger, so the only alternative is `limit=-1`, which drags the entire 21540-row pool over the wire on every question.

## Change

`list_resources` accepts `kn_id`. When present it resolves the network's bindings instead of paging:

1. `GetKnowledgeNetworkDetail(kn_id)` to collect `object_types[].data_source`
2. fetch each bound resource by id via `GetResource`
3. project to the existing `ResourceLite` shape

This deliberately does **not** go through vega's list endpoint. That endpoint is the same pagination, so reusing it would move the caller's bug into the backend and would additionally inherit the unstable ordering from #779. Fetching by id is independent of pool size and needs no vega change.

Behaviour details:

- Bindings are a few dozen tables by nature, so the `kn_id` path returns everything at once. `catalog_id` / `offset` / `limit` are ignored when `kn_id` is set; `type` still filters the network's resources.
- Multiple object types sharing one physical table is normal, so resources are deduplicated. `total_count` is the deduplicated resource count, not the object-type count.
- Resources are fetched with a concurrency cap of 8; serial fetches reached seconds on real networks.
- Behaviour without `kn_id` is untouched, and the new response fields are `omitempty` so existing payloads are byte-identical.

### Unresolvable bindings are classified, not skipped

Three distinct failures used to be indistinguishable from "no resources". They now come back separately, each carrying `object_type_id`:

| Field | Condition | What the caller must do |
|---|---|---|
| `unbound` | no `data_source`, or an empty `data_source.id` | model the binding |
| `stale_binding` | `data_source.type` is a deprecated shape such as `data_view` | rebind to a resource |
| `missing` | id is set but `GetResource` fails (deleted, or not visible to this account) | restore the resource or grant access |

`stale_binding` entries are never sent to vega's resource endpoint — that call is a guaranteed 500 for deprecated `data_view` bindings. A single unresolvable binding never fails the call; the other tables are still returned.

### Tool description alignment

The MCP server instructions said "every other tool needs kn_id" while describing `list_resources` as filterable only by `catalog_id / type`. Models followed the first line, passed `kn_id`, and had it silently dropped — getting the whole pool back rather than the network's tables, which is more misleading than an empty result. Both the zh and en-US instructions, tool metadata and schema descriptions are updated.

## Verification

Unit tests cover: bindings resolved while the list endpoint returns an empty pool (this fails immediately if the implementation ever regresses to page-then-filter), deduplication of a shared resource, classification of all three unresolvable cases, the `type` filter, error propagation, and the no-backend and empty-network paths. Full package suite passes, including `-race` for the concurrent fetch.

End-to-end on 14.103.77.23 against `supplychain_bkn_v4_new2` (18 object types, 14 bound):

- with `kn_id`: 14 entries, `total_count=14`, `unbound` lists exactly `forecast_event`, `mrp_plan_order_event`, `outsource_order_event`, `purchase_requisition_event` — matching the `data_source` set from `get_kn_detail` one for one. The same query returned 0 before this change.
- without `kn_id`: 50 entries, `total_count=21540`, and no new fields in the payload.
- `kn_id` + `type=topic`: 0 entries, as expected for this network.
- `response_format=toon` renders `entries[#14]`.

## Follow-up

`list_resources` still reads `kn_id` from the request body only, unlike `find_skills`, which also falls back to the `X-Kn-ID` header. Worth unifying separately.


sh00tg0a1:
Addressed the non-blocking concern from the review in 6be43471a.

The reported hole was real and worth closing: with every binding failing, the call still returned success with an empty `entries`, so an agent could not tell "the backend is down" from "this network has no tables" — the same silent-empty failure this issue exists to remove, just moved one layer up.

Now: if there are bindings, none resolved, and the first failure is not a per-resource condition, the downstream error is propagated so its status and reason surface.

404/403 are deliberately excluded. A network whose only bound table was deleted, or whose resources this account cannot see, is a modelling fact — it stays in `missing` rather than being dressed up as a service outage. Partial failures are unchanged: the remaining resources are still returned.

Three tests added: all-fail-with-5xx returns an error, all-fail-with-404 stays classified, and partial failure still returns the rest.

Re-verified on 14.103.77.23 after redeploy — unchanged: 14 entries / `total_count=14` / 4 `unbound` with `kn_id`, and 50 entries / `total_count=21540` without it.

PR #838, opened 2026-08-13
feat(context-loader): add a cross-encoder rerank stage to instance recall, default off

sh00tg0a1:
## Links

- Part of #818 (PR2; PR1 landed as #819)
- Design notes: posted on the issue (this repo does not carry design docs under `docs/`)
- Branch: `feature/818-instance-rerank`

## Problem

PR1 replaced the single mixed-score OR with two channels fused by RRF. Rank fusion fixes
*recall* and *comparability*, but two things stay out of its reach:

- **It cannot read.** Rank fusion sees positions, not content, so it cannot tell "欠款"
  (debt) from "还款" (repayment).
- **It cannot say "nothing here is relevant."** The top item always scores 1.0, however
  irrelevant it happens to be — absolute relevance is not encoded in a rank at all.

Both showed up in the live VM run on #818, on real data:

- query `Brazil squad` → `squads/Scotland` ranked **above** four `Brazil` rows, because BM25
  scored it higher and RRF faithfully preserved that rank.
- query `Maracana stadium` → the top rows were `Kashima Stadium`, `Miyagi Stadium`,
  `Niigata Stadium` — the full-text channel matched on the token "Stadium".

A cross-encoder concatenates query and document into one sequence and attends across both,
which is exactly the discrimination missing here.

## Change

`instance_rerank_mode`, three settings:

| mode | behaviour |
|---|---|
| `off` (default) | the model is never called |
| `shadow` | returns the fusion order unchanged, but calls the model and logs Spearman ρ plus how many of the top 5 moved |
| `on` | model scores drive the **order** — and only the order (see below) |

Set it per deployment via `config.concept_search_config.instance_rerank_mode` (with
`instance_rerank_model` for the model name); a request can still override it. Whether reranking
is worth its cost is a deployment decision — it needs a rerank small model registered in the
factory and adds a round trip — and an agent has no basis to make that call.

**`on` changes the order, never `score`.** Overwriting `score` with the model score breaks the
one property PR1 worked to establish: a single scale comparable across object types. Candidates
the model did not score (a vendor returning one row short is enough) would be zeroed, and the
tail beyond `rerank_top_n` never reaches the model at all — one response would then carry two
different scales, and a caller re-sorting by `score` would get a worse order than the one it was
handed. So `score` stays the fusion score, the order carries the reranking, and `rerank_score`
reports what the model said. `Score` also lost its `omitempty`: 0 is a meaningful value here
(local fallback scoring returns it when nothing overlaps), and omitting the field makes "no
field" indistinguishable from "the score is zero".

Default `off` is deliberate: the call costs 100–400 ms, and an unregistered reranker is the
normal state in customer deployments (#114, #788) — the default must not carry that dependency.
`shadow` exists so the switch can be flipped on evidence rather than on belief; its log line
is its entire product.

Implementation points that are easy to get wrong:

- **Placed after global filtering but before property filtering.** `filterNodeProperties`
  trims the property count and truncates values to 500 chars; scoring relevance on the
  leftovers would be scoring the wrong text.
- **Results are aligned by `index`, never by return order** — vendors differ, some return
  sorted by score, some in the original order.
- **Length is bounded twice**: per field (`rerank_field_char_limit`, default 200) and per
  document (3600 chars). mf-model-api silently truncates a document at 4000 chars without
  erroring, so anything past the cut simply does not participate in scoring.
- **Fields are emitted in sorted order** — otherwise map iteration would make the same row
  produce different documents, and the same query different model scores, run to run.
- **Internal metadata (`_score`, `_instance_id`, `_display`) is excluded** from the document;
  it only dilutes the business text.
- **No batching at this layer.** The downstream client already sends 64 per batch across 10
  threads; a second batch size here would just fight it. (`semantic_field_rerank_batch_size`
  stays unused — it still has no consumer anywhere in the repo.)
- **The tail beyond `rerank_top_n` is not sent to the model** but is kept at the end of the
  result, never dropped.

## Degradation

Unrecognised mode, no rerank client, model unregistered (`NameNotExist`), call failure, or a
response whose indices all fall out of range → keep the fusion order. The result is never
emptied and never filled with unrelated rows (#788). On a partial response, scored nodes use
the model score and unscored ones keep their fusion rank.

## Testing

`go test -race ./...` in `adp/context-loader/agent-retrieval/server` — all green.

| Case | Asserts |
|---|---|
| `OffDoesNotCallModel` | the default setting issues zero model calls |
| `OnReordersByModelScore` | a model-preferred row overtakes a higher-fusion-score row; `score` becomes the model score |
| `ShadowKeepsOrderButRecordsScores` | order untouched, fusion score untouched, model score carried out, delta line logged with `top5_changed` |
| `DegradesToFusionOrder` | unregistered / nil response / unusable indices all keep the fusion order intact |
| `PartialResultsKeepUnscoredNodes` | a single scored node leads; the rest survive |
| `AlignsByIndexNotOrder` | a shuffled response maps back correctly |
| `TailBeyondTopNIsKeptUnscored` | only `top_n` documents are sent; the tail stays at the end |
| `InstanceRerankDocument` | internal fields excluded, non-strings skipped, fields labelled and sorted, long values truncated by rune |
| `TotalLengthCapped` | document stays under the cap |
| `UnknownModeFallsBackToOff` | a malformed mode does not fail retrieval |
| `OrderDelta` | ρ=1 for identical orders, ρ=-1 for reversed, top-K movement counted |

### Verified on the VM

The dev VM does have a rerank small model registered (`t_small_model`: `reranker`, type
`reranker`, default), so all three modes exercised the real path rather than the degradation
path. Query `Maracana stadium in Rio` against object type `stadiums`:

| mode | result |
|---|---|
| `off` | Beira-Rio **1.9839** / Maracanã **1.9839** / Kashima 0.9683 / Nacional 0.9683 / Miyagi 0.9531, no `rerank_score` |
| `shadow` | order **unchanged**, model scores carried out: Maracanã **0.585**, Beira-Rio **0.146**, Nacional 0.141, Kashima 0.120, Miyagi 0.113 |
| `on` | **order** changes — Maracanã first, Nacional overtakes Kashima — while `score` stays 1.9839 / 1.9839 / 0.9683 / 0.9683 / 0.9531 throughout |

The deployment switch was exercised too: with `instance_rerank_mode: "on"` in the ConfigMap and
**no `retrieval_config` in the request body**, reranking engaged (log `mode=on`); a request
explicitly sending `off` turned it back off. Both directions matter — "the deployment enabled it,
this call does not want it" has to be expressible, which is why the merge grew a base layer.

This run is a cleaner demonstration than the two examples above, because fusion scored the top
two **exactly equal** at 1.9839 — both were hit by both channels at the same ranks, so rank
fusion had nothing left to separate them with. Beira-Rio even carried the higher BM25 score
(4.695 vs 3.092): its name matches both "stadium" and "Rio" (Beira-**Rio**) on the surface.
The cross-encoder was not fooled — 0.585 against 0.146, a four-fold gap, correctly picking the
stadium the query actually names. That is not a judgement rank fusion can make in principle; it reads positions, not content.

Shadow's log line came out as designed:

```
[InstanceRerank] mode=shadow candidates=5 scored=5 spearman=0.8000 top5_changed=0
```

ρ=0.8 shows the reordering is substantial; `top5_changed=0` simply reflects that there were
only 5 candidates, so the top-5 *set* cannot change — that metric starts discriminating once
the candidate pool exceeds K.

The VM was restored afterwards: the temporary `condition_operations` on `stadiums` reverted and
the `agent-retrieval` image put back to what another developer had deployed.

## Pre-Merge Checklist

- [x] Tests added and passing (`-race`)
- [x] API docs updated (`schema-search.yaml`, `object-instance.yaml`, MCP `search_instance` output schema): `score` / `recall_score` / `rerank_score` semantics
- [x] Backward compatible: every new field has a default, the mode defaults to `off`, and no
      existing response changes shape while it is off
- [x] Verified on the dev VM: three modes plus the deployment switch and its request-side override; ConfigMap, object-type `condition_operations` and image all restored afterwards
- [ ] Human review


sh00tg0a1:
两条待确认项都核实成立，已修（efd39450）。第一条比描述更根本，所以修法也不一样。

## ① score 语义

评审说的是「未打分候选与尾部会带着与文档不符的分值」，核下来根因是**覆盖 score 这个动作本身就是错的**：

- `applied < len(head)` 是允许的（厂商少返一条就会发生），这些候选被赋 0；
- `Score` 带 `omitempty`，0 分让 `score` 字段整个从响应里消失，连融合分也一起丢；
- 超出 `rerank_top_n` 的尾部从没进过模型，仍是融合分。

于是一次响应里同时存在两把尺子。而**融合分跨对象类可比是 PR1 挣来的性质**，不该被精排破坏。

所以没有按「给未打分的补个值」修，而是改成：**精排只改返回顺序，不动 score**。模型判了多少分由 `rerank_score` 单独带出；`Score` 去掉 `omitempty`（0 是有意义的取值——本地兜底打不出重叠就是 0，省略字段会让调用方分不清「没这个字段」和「分是 0」）。

代价是调用方若自行按 `score` 重排会抹掉精排效果，这一点在 MCP output schema、两份 OpenAPI 里都写明了：开启精排后按返回顺序消费。

## ② 部署级入口

属实，文档写「部署开启实例精排」而仓里根本没有那个入口。新增 `config.concept_search_config.instance_rerank_mode` / `instance_rerank_model`。

实现上把合并拆成 `MergeRetrievalConfigWithBase(base, user)`：部署值进 base，请求仍能覆盖。塞进内置默认或塞进 user 都不行——前者表达不了「部署开了、这次调用不想要」，后者会盖掉调用方的显式取值。配置读取放在 `localSearchImpl`（构造期注入的 loader），没放进 `DefaultSemanticInstanceRetrievalConfig`：那是纯函数，在里面读文件会让单测环境（无 `/sysvol/config`）直接 panic——我先写错了一版，被自己的测试拦下来。

## VM 重验

上一轮验的是旧语义，作废。新语义重跑：

| mode | 结果 |
|---|---|
| `off` | 融合序，无 `rerank_score` |
| `shadow` | 序不变，`rerank_score` 带出 |
| `on` | **顺序**变（Maracanã 升第一、Nacional 超过 Kashima），`score` 仍是 1.9839/1.9839/0.9683/0.9683/0.9531 原样不动 |

部署级开关也验了两个方向：ConfigMap 设 `instance_rerank_mode: "on"` 且**请求体不带 retrieval_config** → 精排生效（日志 `mode=on`）；请求显式传 `off` → 关闭。

VM 已还原：ConfigMap、`stadiums` 的 `condition_operations`、镜像（回别人在用的 `ptc4`）。

PR #855, opened 2026-08-13
feat(vega): push resource ownership to bkn-safe, previewing before every write (#800)

sh00tg0a1:
Stage D2 of #800 — the last piece. bkn-safe now knows how to inherit; this gives it the fact it inherits on.

## What it does

Read the local resource snapshot → chunk → **dry-run each chunk** → push when the diff is empty, stop when it is not → read back and reap orphans.

## Why it stops instead of pushing

Ownership landing in the table **is** the moment inheritance starts applying, and that is a widening. An allow-only engine has no knob that holds it back afterwards, so the decision has to be handed to a person before the write, not after.

A non-empty diff logs the affected subjects and skips the chunk. Approval is `VEGA_OWNERSHIP_SYNC_APPROVED=true` — a record that someone read the report, **not** a judgement switch. The judgement switch was deliberately rejected: an empty ownership table already equals "off", and rollback is deleting rows rather than shipping a release.

On a real installation the diff is expected to be empty. The development cluster has **zero** per-object grants on `catalog` and `resource`, and the seeded roles gain nothing (asserted in #832), so the push is a no-op there and runs unattended.

## What it deliberately does not touch

- **Not the `PermissionAccess` port.** Adding a method there would drag in the port definition, the generated mock and the ISF implementation — that is changing a mechanism two providers share. The worker carries a small client of its own instead: a few dozen extra lines for zero blast radius.
- **Not the create/delete paths.** Not one line. Ownership is discovered by reading, never by intercepting writes.
- Resources under internal catalogs are skipped. They register as `internal_catalog` / `internal_resource`, which the seed never declared a hierarchy for, so pushing them returns 400 — by design, not a fault. Nothing but the super-admin wildcard reaches them anyway.

## Version and configuration tolerance

- `BKN_SAFE_URL` unset → the worker does not start (the authorization face may still be ISF, which has no such concept)
- a 404 from the endpoint is read as "the peer does not have this capability yet" — new vega against old bkn-safe is a supported combination, and it must neither crash nor fill the log
- `VEGA_OWNERSHIP_SYNC_INTERVAL` (default 30m), `VEGA_OWNERSHIP_SYNC_APPROVED` (default off)

Reconciliation runs even when the snapshot is empty; otherwise deleting every resource would leave its ownership rows behind forever. It is the backstop for the delete path, which loses rows to crashes and failed cross-service calls — `DELETE /authz/policies` already drops the row on the happy path.

## Tests

Against a programmable bkn-safe stub:

- clean preview → pushes
- dirty preview → **pushes nothing, and is not an error** (waiting is the intended outcome)
- approved → pushes the same snapshot
- orphan rows are reaped; reconciliation still runs on an empty snapshot
- a 404 peer surfaces as `errEndpointAbsent`, not a failure
- no `BKN_SAFE_URL` → the worker declines to start
- chunk boundaries

Note: `vega-backend/worker` already fails on an untouched tree — `gomonkey` segfaults on Apple Silicon in `build_worker_batch_test.go`. Verified before and after; the new tests pass on their own.

### DIFF
--- a/adp/vega/vega-backend/server/main.go
+++ b/adp/vega/vega-backend/server/main.go
@@ -172,6 +172,14 @@ func main() {
 	}
 	logger.Info("VEGA Manager Init Catalog Health Check Worker Success")
 
+	// 资源归属同步：把「表属于哪个目录」推给 bkn-safe，目录上的授权才能到达
+	// 目录下的表（#800）。BKN_SAFE_URL 未配置时自动不启用。
+	osw := worker.NewOwnershipSyncWorker()
+	if err := osw.Start(); err != nil {
+		logger.Fatalf("Failed to start ownership sync worker: %v", err)
+	}
+	logger.Info("VEGA Manager Init Ownership Sync Worker Success")
+
 	// 创建并启动服务
 	server := &mgrService{
 		appSetting:    appSetting,
--- a/adp/vega/vega-backend/server/worker/ownership_sync_worker.go
+++ b/adp/vega/vega-backend/server/worker/ownership_sync_worker.go
@@ -37,6 +37,10 @@ type OwnershipSyncWorker struct {
 	// approved 是「我看过预演报告，可以推」的批准记录，不是判定开关。
 	// 判定开关已经明确否掉：归属表为空天然等于关闭，回滚是删数据不是发版。
 	// 它落在 vega 而不是 bkn-safe，因为决定推不推的是数据的生产方。
+	//
+	// 它是常驻环境变量，因此批准的其实是「此后每一份快照」，而人只看过当时那
+	// 一份。缓解办法不是假装它是一次性的：预演每轮照跑、差异照记日志，扩权因此
+	// 始终留痕；用法是收敛完就把它关回去。
 	approved bool
 	stop     chan struct{}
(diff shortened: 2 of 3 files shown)


sh00tg0a1:
Closing: the design landed somewhere simpler.

Ownership never needed synchronising. vega already holds `catalog_id` on the resource row it is judging, so it can ask about `catalog:<catalog_id>` directly — the fact only had to travel to bkn-safe because I had modelled inheritance as something bkn-safe resolves. Once the translation happens at vega's own boundary, there is no table to fill, nothing to reconcile, and no window where the two sides disagree.

What replaces it (#817): vega asks about the resource first and, when that denies, asks about the owning catalog with the operation translated (`modify`/`delete` → `resource_manage`, reads by the same name, `authorize` deliberately not translated). Pure widening — whoever can see a table today still can, on the first question — so no seed change and nobody loses anything on the day it ships.

The bkn-safe hierarchy from #827/#832/#839 stays in place but goes dormant: dropping `parent_type`/`parent_operation` from the seeded catalog means the climb never fires. It is left rather than reverted because it is a generic mechanism, and reverting three merged PRs costs more churn than a code path that no declaration reaches.

PR #883, opened 2026-08-14
fix(vega): normalize self-referencing field features instead of rejecting them

sh00tg0a1:
## Problem

#837 made `ref_property` pointing at its own field a hard 400
(`VegaBackend.Dataset.InvalidParameter.FieldFeatureRef`, *"cannot reference itself"*).

That shape is exactly what the platform used to persist itself:

```json
{"field": "title",
 "features": [{"name": "title_fulltext", "feature_type": "fulltext", "ref_property": "title"}]}
```

So after the 0.1.4 rollout every read-modify-write of an existing resource
started failing, `vega dataset build` included. A knowledge network that had
already been pushed ends up deadlocked: `bkn push` drops `match`/`knn` from
`condition_operations`, recovering needs an index rebuild, and the rebuild is
now rejected. Semantic retrieval cannot be restored through the CLI at all.

Reported from a customer POC blocked on exactly this.

## Fix

Self-reference is a redundant no-op — capability derivation already falls back
to the owning property when `ref_property` is empty
(`VegaResourceIndexCaps` in bkn-backend), so `ref_property: "title"` on field
`title` and no `ref_property` resolve identically.

`ValidateResourceRequest` now strips the self-reference before validation
instead of failing the request. Legacy rows heal on their next write, so no
data migration is required.

Cross-field `ref_property` validation (missing target, type mismatch,
dataset-category rejection) is unchanged.

## Tests

- `normalizeSchemaFeatures` drops self-references, keeps cross-field ones
- `ValidateResourceRequest` accepts a legacy self-referencing resource and
  returns it normalized
- the #837 assertion that required the 400 is replaced

`go vet` + `go test ./driveradapters/ ./logics/resource/... ./worker/...` pass.

### DIFF
--- a/adp/vega/vega-backend/server/driveradapters/validate_resource.go
+++ b/adp/vega/vega-backend/server/driveradapters/validate_resource.go
@@ -21,33 +21,14 @@ import (
 )
 
 func ValidateResourceRequest(ctx context.Context, req *interfaces.ResourceRequest) error {
-	normalizeSchemaFeatures(req.SchemaDefinition)
+	// 存量资源的自引用 ref_property 在这里抹平。库里读出来的 schema 由 logics 层做同样的
+	// 归一化，两侧必须一致，否则更新会被判成 build 相关变更而清空 LocalIndexName。
+	resourcelogic.NormalizeSelfReferencingFeatures(req.SchemaDefinition)
 
 	if err := validateResourceRequestBase(ctx, req); err != nil {
 		return err
 	}
 	return validateResourceRequestSchema(ctx, req)
-}
-
-// normalizeSchemaFeatures 归一化字段特征里的自引用 ref_property。
-//
-// ref_property 的语义是「特征挂在 A 属性上、但作用于 B 字段」，指向字段自身是一个
-// 无意义的冗余写法：能力派生本来就在 ref_property 为空时落回属性自身（见 bkn-backend
-// 的 VegaResourceIndexCaps），两种写法结果完全相同。
-//
-// 但历史上平台自己写入的存量资源正是这个形状，因此这里必须就地抹平而不是报错——
-// 一旦报错，存量资源的任何读改写（vega dataset build 就是）都会失败，索引再也重建不了。
-func normalizeSchemaFeatures(props []*interfaces.Property) {
-	for _, prop := range props {
-		if prop == nil {
-			continue
-		}
-		for i := range prop.Features {
-			if prop.Features[i].RefProperty == prop.Name {
-				prop.Features[i].RefProperty = ""
-			}
-		}
-	}
 }
 
 func validateResourceRequestBase(ctx context.Context, req *interfaces.ResourceRequest) error {
--- a/adp/vega/vega-backend/server/driveradapters/validate_resource.go
+++ b/adp/vega/vega-backend/server/driveradapters/validate_resource.go
@@ -21,10 +21,6 @@ import (
 )
 
 func ValidateResourceRequest(ctx context.Context, req *interfaces.ResourceRequest) error {
-	// 存量资源的自引用 ref_property 在这里抹平。库里读出来的 schema 由 logics 层做同样的
-	// 归一化，两侧必须一致，否则更新会被判成 build 相关变更而清空 LocalIndexName。
-	resourcelogic.NormalizeSelfReferencingFeatures(req.SchemaDefinition)
-
 	if err := validateResourceRequestBase(ctx, req); err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


kenlee1988:
结论：**仍应请求修改**。我独立复核后确认：

- **P1**：归一化只修改请求 schema，导致与存量资源的 `Features` 不再相等；更新服务因此判定为 build 相关变更并清空 `LocalIndexName`。带自引用特征且已有索引的资源，用户仅修改描述后就会失去索引能力，直至重建完成。见 [`[normalizeSchemaFeatures](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/driveradapters/validate_resource.go#L24)`](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/driveradapters/validate_resource.go#L24) 与 [`[validateMutableSchemaUpdate](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/logics/resource/resource_service.go#L1499)`](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/logics/resource/resource_service.go#L1499)。应让更新比较把“自引用”和“空引用”视为等价，或同时规范化已存 schema，且补充“普通编辑不清空本地索引”的回归测试。

- **P2**：归一化发生在类别校验之前，使 dataset 的自引用 `ref_property` 从明确的 400 变成静默接受。见[[调用顺序](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/driveradapters/validate_resource.go#L24)](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/driveradapters/validate_resource.go#L24)与[[data-set 约束](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/driveradapters/validate_resource.go#L187)](https://github.com/openbkn-ai/bkn-foundry/blob/0edbef047b1f58ba69617354cbc2fb0264dccead/adp/vega/vega-backend/server/driveradapters/validate_resource.go#L187)。应只对允许 `ref_property` 的原始资源执行该兼容归一化，或先保留 dataset 的拒绝语义。

远端 Vega CI 通过，但 PR 当前因上述变更请求处于阻塞状态。本地基线的 `go test ./driveradapters/ ./logics/resource/... ./worker/...` 通过；未对 GitHub 写入评论。

sh00tg0a1:

/review

sh00tg0a1:
Both points addressed, thanks — the second one was a fair catch that I had argued my way around the first time.

**P1 — normalization only on the request side.** Confirmed against the source: `validateMutableSchemaUpdate` compares `Features` with `reflect.DeepEqual`, and `resource_service.go` clears `LocalIndexName` whenever the update is deemed build-relevant. Stripping the self-reference on one side only made a description-only edit look like a feature change.

Took your second option: the helper moved to `logics/resource` as `NormalizeSelfReferencingFeatures` and now runs on both sides — the incoming request and the stored schema, right before the diff. There is an explicit regression test for "a normal edit is not a build-relevant change", plus one that asserts the one-sided case *would* be detected, so the guard cannot rot.

While wiring that up I found the same class of problem one layer down, which neither of us had listed: a resource that has never been re-written still carries the self-reference in the database, and `validateBuildTaskSchemaFeatures` rejects it independently of #837 — a dataset for any `ref_property`, and a self-referencing keyword feature on a `text` field because keyword requires a `string` target. Build task creation therefore still failed for exactly the resources this PR is meant to unblock. The snapshot is normalized too now.

**P2 — dataset rejection weakened.** You are right, and my first answer here was wrong. I had added a test *documenting* the relaxation, which is the worse of the two mistakes: `validateResourceRequestSchema` passes `allowRefProperty=false` for datasets, so dataset `ref_property` has answered 400 since well before #837. It was never part of this regression and relaxing it is out of scope.

Normalization now lives in the original-resource branch only, and the test asserts a dataset self-reference is still rejected with the value left untouched.

One asymmetry left deliberately, flagged here so it is a decision rather than an oversight: `validateBuildTaskSchemaFeatures` still normalizes unconditionally, datasets included. The write boundary stays strict, but a row already in the database — whatever wrote it, migration or direct write — should not be unable to build. Say the word if you would rather have that path mirror the strict rule.

sh00tg0a1:
Follow-up on top of the earlier fixes, plus one thing worth a second opinion.

**Since the last round.** The normalization moved out of `validateBuildTaskSchemaFeatures` and up into `buildLocalIndexSchema`, right after the deep copy it already makes. A function named `validate...` that quietly rewrites its input is a trap for the next reader, and the new position also makes the important property visible at a glance: only the snapshot is normalized, the resource row is left alone and heals on its next update. The tests moved with it and now drive `buildLocalIndexSchema`, asserting both that the build succeeds and that `res.SchemaDefinition` still holds the self-reference.

**State of the two findings.**

P1 is fixed as you suggested — normalize both sides. `NormalizeSelfReferencingFeatures` lives in `logics/resource` and runs on the incoming request and on the stored schema right before `validateMutableSchemaUpdate` diffs them. Two regression tests: a description-only edit is not a build-relevant change, and the one-sided case still *is* detected, so the guard cannot silently rot.

P2 is fixed as you asked, and my first answer to it was wrong in a way worth naming: I had added a test documenting the relaxation, which makes a behaviour change look intentional. Datasets have answered 400 for any `ref_property` since long before #837, so it was never part of this regression. Normalization is now scoped to the original-resource branch and the test asserts a dataset self-reference is still rejected with the value untouched.

**One asymmetry I would like you to rule on.** `buildLocalIndexSchema` normalizes unconditionally, datasets included, while the write path keeps rejecting them. The reasoning is strict at the boundary, tolerant of what is already stored: a row written by a migration or by hand should not be unable to build. If you would rather the build path mirror the strict rule, say so and I will drop the dataset case — it only costs the one test.

PR #884, opened 2026-08-14
fix(vega): quote MariaDB identifiers in SELECT, FROM, GROUP BY and ORDER BY

sh00tg0a1:
## Problem

A source column named after a SQL reserved word passes every stage except the
one that matters:

| Stage | Result |
|---|---|
| `CREATE TABLE` with a `key` column | OK |
| `catalog discover` | OK, column recognised |
| `bkn push` | OK |
| `object-type query` | **HTTP 500** `InternalError.GetViewDataByIDFailed` |

The WHERE clause already went through `quoteColumnName`, but the select list,
table name, aliases, GROUP BY and ORDER BY were concatenated raw:

```go
selectFields = append(selectFields, prop.OriginalName)
builder := sq.Select(selectFields...).From(resource.SourceIdentifier)
```

so MySQL received `SELECT id, key, ... FROM yanfeng_kb.fact` and returned a 1064
syntax error, surfaced to the caller as a generic internal error with no mention
of the column or of reserved words.

## Fix

Statement assembly moves into `buildSelectBuilder`, and every identifier —
columns, table, aggregate aliases, group-by and sort fields — goes through
`quoteColumnName` / `qualTable`. `date_format(...)` expressions keep their shape
with the inner column quoted.

Two behaviour changes worth calling out:

- Sort fields now resolve through the schema like every other clause. Sorting by
  a property whose `original_name` differs from its name previously emitted a
  non-existent column; it now emits the mapped column.
- `output_fields` de-duplication tracks selected names in a set instead of
  string-matching `" AS "+name` against the rendered select entries, which no
  longer holds once aliases are quoted.

PostgreSQL already quotes its table name via `qualTable`; its select list has
the same gap and is left for a follow-up so this stays reviewable.

## Tests

`buildSelectBuilder` is now testable without a live database. Added coverage for
the reserved-word column, `output_fields` mapping, sorting, aggregation with
alias, alias de-duplication, and calendar-interval grouping.

`go vet ./logics/connector/...` and `go test ./logics/connector/...` pass.

### DIFF
--- a/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_condition.go
+++ b/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_condition.go
@@ -133,6 +133,11 @@ func quoteColumnName(name string) string {
 	return "`" + strings.ReplaceAll(strings.TrimSpace(name), "`", "``") + "`"
 }
 
+// qualTable 将资源的源端标识转为反引号限定的表名；支持 "db.table" -> "`db`.`table`"
+func qualTable(sourceIdentifier string) string {
+	return quoteColumnName(sourceIdentifier)
+}
+
 func (c *MariaDBConnector) ConvertFilterCondition(ctx context.Context, condition interfaces.FilterCondition,
 	fieldsMap map[string]*interfaces.Property) (sq.Sqlizer, error) {
 
--- a/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go
+++ b/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go
@@ -236,8 +236,16 @@ func (c *MariaDBConnector) buildSelectBuilder(resource *interfaces.Resource,
 			dir = "DESC"
 		}
 
-		// Check whether this is a GROUP BY field using calendar_interval
+		// Resolve the sort field against the SELECT aliases first: when an alias matches a
+		// property name the caller means the aggregate, but schema resolution would pick
+		// that property's source column instead. The default sql_mode has no
+		// ONLY_FULL_GROUP_BY, so that does not error — it orders by an arbitrary row.
 		sortField := quoteColumnName(originalName(sortItem.Field))
+		if aggAlias != "" && sortItem.Field == aggAlias {
+			sortField = quoteColumnName(aggAlias)
+		}
+
+		// Check whether this is a GROUP BY field using calendar_interval
 		for _, groupByItem := range params.GroupBy {
 			if groupByItem.Property == sortItem.Field && groupByItem.CalendarInterval != "" {
 				// Use the full date_format expression
--- a/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go
+++ b/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go
@@ -101,6 +101,159 @@ func (c *MariaDBConnector) ExecuteRawSQL(ctx context.Context, sql string) (*inte
 	return response, nil
 }
 
+// buildSelectBuilder assembles the SELECT statement for detail and aggregate queries.
+//
+// Every identifier — table, column, alias — must go through quoteColumnName / qualTable.
+// A source column named after a SQL reserved word (say `key`) survives DDL, catalog
+// discover and bkn push untouched, because none of them build a query; concatenating it
+// raw only fails at execution time with a 1064 that names nothing useful.
+func (c *MariaDBConnector) buildSelectBuilder(resource *interfaces.Resource,
+	params *interfaces.ResourceDataQueryParams, fieldMap map[string]*interfaces.Property,
+	condition sq.Sqlizer) (sq.SelectBuilder, error) {
+
+	// Source column name; fall back to the property name when the schema has no mapping.
+	originalName := func(property string) string {
+		if field, ok := fieldMap[property]; ok {
+			return field.OriginalName
+		}
+		return property
+	}
+
+	// Construct the SELECT clause
+	selectFields := []string{}
+	// Output names already selected (column or alias), used to de-duplicate output_fields
+	selected := map[string]struct{}{}
+
+	// Add the GROUP BY field (when aggregating queries)
+	for _, groupByItem := range params.GroupBy {
+		column := originalName(groupByItem.Property)
+		// Check whether calendar_interval is needed
+		if groupByItem.CalendarInterval != "" {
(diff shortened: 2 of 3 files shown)


kenlee1988:
结论：**请求修改后再合并**。发现两项问题：

- **P1**：[`buildHavingCondition` 未引用转义后的聚合别名](https://github.com/openbkn-ai/bkn-foundry/blob/e6e5e57fbfaa45a4954a3be2b5dff43248dcabb6/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go#L368)。例如别名为 `key` 且带 `having: {field: "__value"}` 时，会生成 `AS \`key\` ... HAVING key > 0`，`HAVING` 中仍是保留字裸标识符，查询继续报 1064。应在此处使用 `quoteColumnName(aggAlias)`，并补充含 HAVING 的保留字别名测试。

- **P2**：[`originalName` 对空 `OriginalName` 没有回退](https://github.com/openbkn-ai/bkn-foundry/blob/e6e5e57fbfaa45a4954a3be2b5dff43248dcabb6/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go#L115)。对于存在于 schema、但 `original_name` 为空的属性，新的排序路径会从原先可用的 `ORDER BY <property>` 变为 `ORDER BY ```，导致 SQL 错误。应仅在 `OriginalName` 非空时使用它，否则回退到 `property`。

我未向 GitHub 写入评论或更改 PR。已确认远端 Vega CI 成功；本地当前基线的 MariaDB connector 测试也通过。

sh00tg0a1:

/review

sh00tg0a1:
Both points fixed, and the second one was a regression I introduced rather than a pre-existing gap — thanks for catching it.

**P1 — `buildHavingCondition` used the raw alias.** Correct: the SELECT list quoted the alias while HAVING concatenated it bare, so `alias: "key"` produced ``COUNT(`id`) AS `key` ... HAVING key >= 3`` and still hit 1064. `Aggregation.Alias` comes straight from the request body with no validation, so the reserved-word case is reachable by any caller. Now routed through `quoteColumnName`, with a HAVING test that pins a reserved-word alias.

**P2 — no fallback for an empty `original_name`.** Confirmed and worse than a cosmetic gap. `appendTaskEmbeddingVectorFields` builds properties with `Name` and `Type` only, so every vector field a build task adds has an empty `original_name`. Sorting by one of those used to work, because ORDER BY previously used the raw property name — my change turned it into an empty quoted identifier and broke the statement. The resolver now treats an empty `original_name` as "same as the property name", which is the assumption the `filter_condition` constructors already encode when they synthesise `OriginalName: cfg.Name`. Covered by a test that sorts and selects a vector field.

I also went back and fixed the alias-shadowing case the automated pass raised twice and I had waved off. Resolving sort fields through the schema is right for column references and wrong for the aggregate alias: when the alias matches a property name the caller means the aggregated value, but the query ordered by that property's source column instead. Default `sql_mode` has no `ONLY_FULL_GROUP_BY`, so it does not error — it orders by an arbitrary row's value. The alias is now checked before schema resolution, with a test where the alias deliberately shadows a property.

sh00tg0a1:
Both findings are fixed, and there is one CI signal on this PR that needs a decision rather than a code change.

**P1** — `buildHavingCondition` now quotes the aggregate alias. Worth restating why it mattered: `Aggregation.Alias` arrives from the request body and `validateAggregateParams` does not constrain it, so `alias: "key"` was reachable by any caller and produced ``COUNT(`id`) AS `key` ... HAVING key >= 3`` — SELECT fixed, HAVING still 1064. Covered by a reserved-word alias case.

**P2** — the empty `original_name` fallback is in. This was a regression I introduced rather than a pre-existing gap: `appendTaskEmbeddingVectorFields` builds properties with only `Name` and `Type`, so every vector field a build task adds has an empty `original_name`, and sorting by one used to work because ORDER BY previously used the raw property name. Resolution now treats empty as "same as the property name", which is what the `filter_condition` constructors already assume when they synthesise `OriginalName: cfg.Name`. Test sorts and selects a vector field.

I also took the alias-shadowing point from the automated pass, which I had waved off in the first round on the grounds that it needs an alias colliding with a property name. That reasoning was wrong: default `sql_mode` has no `ONLY_FULL_GROUP_BY`, so the collision does not error, it orders by an arbitrary row's value. Silent wrong ordering is worse than the loud failure I was weighing it against. Sort now resolves the aggregate alias before falling back to the schema, with a test where the alias deliberately shadows a property.

**CodeQL.** The check is red with two high `go/sql-injection` alerts pointing at the SELECT/FROM assembly. They are false positives and the PR strictly reduces the exposure they describe: those identifiers were concatenated raw on main and are escaped here — CodeQL simply does not recognise `quoteColumnName` as a sanitizer. The same pattern already carries 53 open alerts of this rule on main, 34 of them in the mariadb/postgresql connectors, so this is a pre-existing class surfaced by moving the lines into a new function. Needs either a dismissal or a decision to merge with the check red; I have not touched the repository's security settings.

kenlee1988:
复核 #884 最新提交后，仍建议请求修改：

- P1：[`[mariadb_query.go:181](https://github.com/openbkn-ai/bkn-foundry/blob/72bdd5ab51644383a50281307481d2ddc7bfb33a/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go#L181)`](https://github.com/openbkn-ai/bkn-foundry/blob/72bdd5ab51644383a50281307481d2ddc7bfb33a/adp/vega/vega-backend/server/logics/connector/local/table/mariadb/mariadb_query.go#L181) 默认未传 `output_fields` 时，仍直接使用 `prop.OriginalName`。若其为空，会生成 `SELECT `` ...` 并导致 SQL 语法错误；这绕过了本次新增的 `originalName()` 回退逻辑。应改为 `quoteColumnName(originalName(prop.Name))`，并补充“空 `OutputFields` + 空 `OriginalName`”测试。

上轮的 HAVING 聚合别名转义、显式字段查询的 `OriginalName` 回退，以及聚合别名排序遮蔽问题均已修复。现有相关 CI 已通过；未向 GitHub 写入评论。

sh00tg0a1:
Fixed. The select-all branch — no `output_fields`, so every schema property is selected — still read `prop.OriginalName` directly and bypassed the fallback I had added for the explicit-field and sort paths, producing the same empty quoted identifier for a property with no `original_name`. Same defect as P2, one branch further down; I fixed the paths the failing test exercised and did not sweep the rest of the function.

Now goes through `originalName(prop.Name)`, with a test that selects all fields from a schema containing a property with an empty `original_name`.

PR #885, opened 2026-08-14
fix(vega): pin like/not_like to literal substring matching

sh00tg0a1:
## Problem

`like` meant two different things inside vega, depending on where the resource
was stored:

| Path | Implementation | `%Indirect%` |
|---|---|---|
| SQL / MariaDB / PostgreSQL | `"%" + Special.Replace(v) + "%"` — escapes the caller's `%` into a literal, then wraps | 0 rows |
| OpenSearch + logic-view DSL | `replaceLikeWildcards` — translates `%` to `.*`, matched as an anchored regexp | matches |

So the same condition returned different rows on different backends, and on the
SQL family it returned an **empty set with no error**:

```
SELECT id FROM knowledge WHERE title LIKE '%Indirect%'      -> 3 rows
{"field":"title","operation":"like","value":"%Indirect%"}   -> 0 rows
```

`condition_operations` advertises `like`, the operator appears to run, and the
caller is left checking the data, the index and the push — all of which are
fine. This is the only defect in the batch that produces a wrong answer rather
than a confusing error.

## Fix

The contract is now **literal substring matching**, which is what the SQL
connectors already did. Choosing SQL-LIKE semantics instead would have silently
broken every caller passing a bare substring today, turning working queries into
empty ones — the same failure this fixes.

- `ParseLikeValue` rejects unescaped `%` / `_` and names `regex` as the operator
  for pattern matching, so a SQL-style pattern now fails loudly:

  ```
  condition [like] value is matched as a literal substring, so the wildcard '%'
  is not supported; use operation [regex] for pattern matching, or escape it as
  '\%' to match the character itself
  ```

  `\%` and `\_` still match those characters literally.
- The OpenSearch and logic-view DSL builders emit `wildcard` with the escaped
  literal wrapped in `*...*` instead of a translated regexp, so index-backed
  resources agree with the SQL connectors. `replaceLikeWildcards` is replaced by
  `likeContainsPattern`, which only has to escape `*`, `?` and `\`.
- Filter-condition construction errors return 400 instead of 500. Field not
  found, wrong value type and misused operators are caller input; a 500 says the
  service is broken when the request is what needs fixing.

BKN backend keeps its own `like` builders (`.*value.*` regexp on the DSL path,
no wrapping at all on the SQL path). Aligning those is the follow-up; this PR
fixes the path `query_object_instance` actually takes.

## Tests

`ParseLikeValue` covers plain, CJK, escaped `\%` `\_` `\\`, trailing backslash,
and both rejection cases. OpenSearch and DSL builder tests assert the new
`wildcard` shape including metacharacter escaping. Existing fixtures that fed
SQL wildcards to `like` are updated to the new contract.

`go build ./...`, `go test ./logics/... ./worker/...` pass.

### DIFF
--- a/adp/vega/vega-backend/server/interfaces/condition.go
+++ b/adp/vega/vega-backend/server/interfaces/condition.go
@@ -33,6 +33,11 @@ type FilterCondCfg struct {
 	ValueOptCfg `mapstructure:",squash"`
 
 	RemainCfg map[string]any `mapstructure:",remain"`
+
+	// LegacyLikeWildcards 标记这条 like/not_like 来自视图定义里存的老写法（值里用 % 当
+	// 通配符）。置位后按各后端改动前的语义渲染：SQL 侧把 % 转义成字面量，索引侧翻译成
+	// 正则通配符。仅由服务端在读取存量定义时设置，不走线也不接受调用方传入。
+	LegacyLikeWildcards bool `json:"-" mapstructure:"-"`
 }
 
 //go:generate mockgen -source ../interfaces/condition.go -destination ../interfaces/mock/mock_condition.go
--- a/adp/vega/vega-backend/server/interfaces/condition.go
+++ b/adp/vega/vega-backend/server/interfaces/condition.go
@@ -34,9 +34,11 @@ type FilterCondCfg struct {
 
 	RemainCfg map[string]any `mapstructure:",remain"`
 
-	// LegacyLikeWildcards 标记这条 like/not_like 来自视图定义里存的老写法（值里用 % 当
-	// 通配符）。置位后按各后端改动前的语义渲染：SQL 侧把 % 转义成字面量，索引侧翻译成
-	// 正则通配符。仅由服务端在读取存量定义时设置，不走线也不接受调用方传入。
+	// LegacyLikeWildcards marks a like/not_like that came from a view definition written before
+	// the literal-substring contract, i.e. one using % as a wildcard. When set, each backend
+	// renders the value the way it did before that change: the SQL family escapes % into a
+	// literal, the index path translates it back into a wildcard regexp. Set only by the server
+	// when it reads a stored definition — it is not on the wire and cannot be sent by a caller.
 	LegacyLikeWildcards bool `json:"-" mapstructure:"-"`
 }
 
--- a/adp/vega/vega-backend/server/logics/connector/local/index/opensearch/opensearch_condition.go
+++ b/adp/vega/vega-backend/server/logics/connector/local/index/opensearch/opensearch_condition.go
@@ -1129,9 +1129,10 @@ func (c *OpenSearchConnector) likeContainsPattern(input string) string {
 	return "*" + escaped.String() + "*"
 }
 
-// legacyLikeWildcardRegexp 还原改动前索引路径对 like 老写法的处理：% -> .*、_ -> .，
-// 反斜杠转义保留。只在条件被标记为老写法时用（见 filter_condition.MarkLegacyLikeWildcards），
-// 存量视图定义因此结果不变；新写法走 likeContainsPattern 的字面子串语义。
+// legacyLikeWildcardRegexp reproduces how the index path handled a like value before this
+// change: % -> .*, _ -> ., backslash escapes preserved. It is used only for conditions marked
+// as legacy (see filter_condition.MarkLegacyLikeWildcards), so stored view definitions return
+// the same rows as before; new values go through likeContainsPattern's literal semantics.
 func (c *OpenSearchConnector) legacyLikeWildcardRegexp(input string) string {
 	if input == "" {
 		return input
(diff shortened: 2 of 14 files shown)


kenlee1988:
结论：**请求修改后再合并**。

- **P1**：[`ParseLikeValue` 直接拒绝未转义的 `%`/`_`](https://github.com/openbkn-ai/bkn-foundry/blob/005cae325a11f53e8f0f94311d2e7248f7bdd9e9/adp/vega/vega-backend/server/logics/filter_condition/condition_like.go#L90)，会让已有的 SQL-LIKE 调用与存量逻辑视图过滤条件从“可执行”变为 400。派生逻辑视图会直接读取并执行持久化的 `nodeCfg.Filters`，并非全是当前请求参数。应提供兼容/迁移方案，或保持未转义通配符的旧行为，而不是在入口硬拒绝。

- **P2**：[`queryDerivedLogicView` 将所有条件构建失败归为客户端 400](https://github.com/openbkn-ai/bkn-foundry/blob/005cae325a11f53e8f0f94311d2e7248f7bdd9e9/adp/vega/vega-backend/server/logics/resource_data/logic_view/logic_view.go#L262)。这里合并了请求过滤条件和持久化的 `nodeCfg.Filters`；后者无效时属于服务端资源配置问题，改为 400 会误导调用方，且掩盖需要修复的视图定义。应按错误来源分别映射状态码，或在保存逻辑视图时完成配置校验。

远端 CI 已通过；本地当前基线的相关 filter、OpenSearch、逻辑视图测试通过。我没有向 GitHub 写入任何评论。

sh00tg0a1:

/review

sh00tg0a1:
Both addressed. P1 was already covered by a push you had not seen when you reviewed; P2 is now fixed.

**P1 — hard rejection breaks existing callers and stored view filters.** Agreed, and split into two parts.

Rejecting `_` was overreach on my side and is gone: every path except the OpenSearch regexp builder already treated it literally — `Special.Replace` escapes it to `\_` for SQL, and it is not a metacharacter inside the DSL `.*value.*` — so it never produced the silent-empty-set failure that justifies rejecting `%`. Underscores in search terms are common enough that rejecting them was pure collateral damage. Only `%` is rejected now, and `\_` still parses as an escape so the two stay symmetric.

For `%` the contract stays, but not at the expense of persisted definitions, which is your actual point: `nodeCfg.Filters` is server-side data the caller cannot edit, so one upgrade would leave a view permanently 400-ing. `EscapeLegacyLikeWildcards` rewrites unescaped `%` in stored conditions to `\%` and logs a warning naming the view. That reproduces the pre-change behaviour exactly — the SQL connectors escaped `%` into a literal anyway — so no stored view changes results, and there is a signal telling operators which definitions to update. Caller-supplied conditions still fail loudly with the pointer to `regex`. That is the permissive-plus-shadow-log first stage the repo's compatibility rule asks for; flipping stored definitions to hard rejection can come later, once the logs say the field is clean.

**P2 — everything mapped to 400.** Fixed. The two halves are now validated separately and the status code follows the source: the stored half returns 500 with the view id in the message, the caller half returns 400. The merged construction afterwards can only fail on something server-side by then, so it maps to 500 too.

On your alternative — validating the filter condition when the logic view is saved — I did not take it here. It is the better long-term answer, but it only helps definitions written after it ships, so it does not remove the need for source-attributed status codes on the read path. Worth a separate issue if you want it.

sh00tg0a1:
The compatibility layer was wrong in its first shape and has been rebuilt. Worth spelling out, because the mistake is instructive.

Rewriting a stored `%` into an escaped literal reproduced what the **SQL** connectors used to do, and silently broke what the **index** path used to do. On an index-backed view, `like "%abc%"` previously became the regexp `.*abc.*` and matched documents containing abc. Rewriting it to a literal produced the wildcard `*%abc%*`, which matches only documents containing the characters `%abc%` — HTTP 200, zero rows, one warning in the log. In other words, the layer meant to protect existing views reintroduced exactly the silent-empty-set failure this whole series exists to remove, on the half of the storage kinds where the query had been working.

**Mark instead of rewrite.** `MarkLegacyLikeWildcards` sets a flag on the stored condition and leaves the value untouched. `LikeCond` and `NotLikeCond` carry the flag through, and each connector renders it the way it rendered it before this branch: the SQL family escapes `%` into a literal, the OpenSearch and logic-view DSL builders translate it back into a wildcard regexp. Every existing view returns exactly the rows it returned before, on both storage kinds, and the warning still names what to migrate. Caller-supplied conditions are unaffected and still get the literal-substring contract with the pointer to `regex`.

The flag is `json:"-"` and `mapstructure:"-"`, so it is set only by the server when it reads a stored definition and cannot be smuggled in by a caller to opt out of the new contract.

Coverage now includes the DSL builder, which was the untested half and, not coincidentally, the half whose behaviour diverged. Legacy `like` and `not_like` are asserted on both the OpenSearch connector and the logic-view DSL generator, and the marker tests assert the stored value is *not* modified.

Also in this branch from the previous round: `_` is no longer rejected (it was treated literally on every path except the OpenSearch regexp builder, so it never had the failure that justifies rejecting `%`), and filter-condition errors are attributed by source — the stored half returns 500 with the view id, the caller half 400.

The three stored-filter entry points are covered: `queryDerivedLogicView`, and `buildFilterSQL` / `buildDSLCondition` for the resource, join and union nodes of composite views.

PR #893, opened 2026-08-15
fix(bkn): align like/not_like with the literal substring contract

sh00tg0a1:
> [!NOTE]
> **Dependency satisfied.** #885 merged on 2026-08-18, and this branch is rebased on a main
> that contains it, so the escape mismatch the earlier blocking note warned about is gone at
> the code level.
>
> One thing still holds at **deploy** time: bkn-backend and vega-backend are separate images.
> Roll vega-backend out first, or roll both together. A bkn-backend carrying this change in
> front of a vega-backend without #885 still matches a literal backslash instead of `%`.

Follow-up to #885, which pinned `like` / `not_like` in vega to literal substring
matching. bkn-backend carried three more spellings of the same operator:

| Path | Before | Effect |
|---|---|---|
| `Convert` (OpenSearch DSL) | `.*<raw value>.*` anchored regexp on the `.keyword` subfield | caller's `%` matched literally, regex metacharacters in the value interpreted |
| `Convert2SQL` | `"field" LIKE '<escaped value>'`, no wrapping | `like` behaved as equality — while `not_like` two files over did wrap in `%...%` |
| `convertLikeCondToDatasetFilterCondition` | value passed straight through | the path `query_object_instance` takes, with no validation at all |

## Fix

Same contract as #885: the value is a literal substring, unescaped `%` and `_`
are rejected and point at `regex`, `\%` and `\_` match those characters
literally.

- `ParseLikeValue` and `LikeContainsPattern` mirror the vega implementation.
  They are duplicated rather than shared because the two services do not share a
  condition package; the rules are stated in both doc comments so a future change
  to one is visible from the other.
- The DSL builders emit `wildcard` with `*literal*` instead of a regexp.
- `Convert2SQL` wraps both operators in `%...%` over the escaped literal, so
  `like` and `not_like` finally mean the same thing.
- The dataset filter converters validate as well, naming the object type
  property in the error rather than leaving vega to report a resource field name
  the caller never mentioned.

Not addressed here: `Convert2SQL` quotes identifiers with `"..."`, and the DSL
path routes text fields to the `.keyword` subfield where the default
`ignore_above: 256` silently drops long values from the index. Both are
pre-existing and orthogonal.

## Tests

`ParseLikeValue` covers plain, CJK, escaped `\%` / `\_`, and rejection with the
message contents asserted. The dataset filter converter is covered on both the
pass-through and rejection paths. Existing DSL and SQL fixtures are updated to
the new contract, including an exact-SQL assertion for the escaped case.

`go vet ./common/...` and `go test ./common/... ./logics/object_type/...` pass.

### DIFF
--- a/adp/bkn/bkn-backend/server/common/condition/condition_like.go
+++ b/adp/bkn/bkn-backend/server/common/condition/condition_like.go
@@ -87,12 +87,13 @@ func convertLikeCondToDatasetFilterCondition(cfg *CondCfg) (map[string]any, erro
 
 // ParseLikeValue 校验并解析 like / not_like 的值，返回要匹配的字面子串。
 //
-// like 的契约是「子串包含」，不是 SQL LIKE 模式：% 和 _ 不作通配符解释。此前各条路
-// 的处理并不一致——DSL 路径把值原样塞进 .*value.* 正则，SQL 路径连两端的 % 都不补，
-// 而下游 vega 的 SQL 连接器又会把 % 转义成字面量，于是 "%foo%" 这种 SQL 写法在任何
-// 一条路上都恒返回空集且不报错。
+// like 的契约是「子串包含」，不是 SQL LIKE 模式。此前各条路对 % 的处理并不一致——DSL
+// 路径把值原样塞进 .*value.* 正则，SQL 路径连两端的 % 都不补，而下游 vega 的 SQL 连接器
+// 又会把 % 转义成字面量，于是 "%foo%" 这种 SQL 写法在任何一条路上都恒返回空集且不报错。
+// 因此未转义的 % 显式报错，指向 regex；要匹配字面量写 \%。
 //
-// 因此显式拒绝未转义的 % 与 _：需要模式匹配用 regex，需要匹配字面量则写 \% \_。
+// _ 不在拒绝之列：改动前每条路都把它当字面量（Special.Replace 转义成 \_，DSL 的 .* 正则
+// 里也不是元字符），不存在 % 那种「静默空集」，而检索词里带下划线极常见。
 // 与 vega 的 filter_condition.ParseLikeValue 同一套规则。
 func ParseLikeValue(operation, value string) (string, error) {
 	var literal strings.Builder
@@ -109,11 +110,11 @@ func ParseLikeValue(operation, value string) (string, error) {
 			escaped = false
 		case r == '\\':
 			escaped = true
-		case r == '%' || r == '_':
+		case r == '%':
 			return "", fmt.Errorf(
-				"condition [%s] value is matched as a literal substring, so the wildcard '%s' is not supported; "+
-					"use operation [regex] for pattern matching, or escape it as '\\%s' to match the character itself",
-				operation, string(r), string(r))
+				"condition [%s] value is matched as a literal substring, so the wildcard '%%' is not supported; "+
+					"use operation [regex] for pattern matching, or escape it as '\\%%' to match the character itself",
+				operation)
 		default:
 			literal.WriteRune(r)
 		}
--- a/adp/bkn/bkn-backend/server/common/condition/condition_like.go
+++ b/adp/bkn/bkn-backend/server/common/condition/condition_like.go
@@ -9,6 +9,7 @@ package condition
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	dtype "bkn-backend/interfaces/data_type"
 )
@@ -33,20 +34,23 @@ func NewLikeCond(ctx context.Context, cfg *CondCfg, fieldsMap map[string]*ViewFi
 	if !ok {
 		return nil, fmt.Errorf("condition [like] right value is not a string value: %v", cfg.Value)
 	}
+	literal, err := ParseLikeValue(OperationLike, val)
+	if err != nil {
+		return nil, err
+	}
 
 	return &LikeCond{
 		mCfg:             cfg,
-		mValue:           val,
+		mValue:           literal,
 		mFilterFieldName: getFilterFieldName(cfg.Field, fieldsMap, false),
 	}, nil
 }
 
 func (cond *LikeCond) Convert(ctx context.Context, vectorizer func(ctx context.Context, words []string) ([]*VectorResp, error)) (string, error) {
-	valPattern := fmt.Sprintf(".*%s.*", cond.mValue)
-	v := fmt.Sprintf("%q", valPattern)
+	v := fmt.Sprintf("%q", LikeContainsPattern(cond.mValue))
 	dslStr := fmt.Sprintf(`
 					{
-						"regexp": {
+						"wildcard": {
 							"%s": %v
 						}
 					}`, cond.mFilterFieldName, v)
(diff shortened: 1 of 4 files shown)


kenlee1988:
结论：**请求修改后再合并**。

- **P1**：dataset 转换仍将原始 `\%` / `\_` 值透传，并假设 Vega 会“再次解析”。见 [`[condition_like.go:75-84](https://github.com/openbkn-ai/bkn-foundry/blob/6ce0863d9382da0b985c771a5eb6dd6185bbe38b/adp/bkn/bkn-backend/server/common/condition/condition_like.go#L75)`](https://github.com/openbkn-ai/bkn-foundry/blob/6ce0863d9382da0b985c771a5eb6dd6185bbe38b/adp/bkn/bkn-backend/server/common/condition/condition_like.go#L75)。但其依赖的 #885 仍是未合并 PR，且两者从同一父提交分支；若 #893 先合并，当前 Vega 的表连接器会把反斜杠再次转义。用户按新提示传 `50\%` 后，MariaDB/PostgreSQL 会匹配“反斜杠 + 百分号”而非 `%`，静默返回空集。应将 #893 建立在 #885 之上并强制合并顺序，或合并为一个原子变更。

- **P2**：[`ParseLikeValue` 新增对未转义 `_` 的拒绝](https://github.com/openbkn-ai/bkn-foundry/blob/6ce0863d9382da0b985c771a5eb6dd6185bbe38b/adp/bkn/bkn-backend/server/common/condition/condition_like.go#L112)，却没有对应的跨后端错误语义需要修复：改动前 BKN 的 SQL 路径会转义 `_` 为字面量，DSL 路径也将其视为普通字符。因此诸如 `order_id`、`customer_name` 的正常字面子串查询会从可用变为错误；应保留 `_` 的字面量兼容性，或提供迁移策略。

远端 bkn 构建与测试任务通过；未向 GitHub 写入评论。

sh00tg0a1:

/review

sh00tg0a1:
Both points taken. P2 was already fixed in a push you had not seen; P1 stands and I verified it end to end rather than reasoning about it.

**P2 — rejecting `_` had no cross-backend failure to fix.** Agreed, and already reverted. Before this branch every path treated `_` literally — `Special.Replace` escapes it to `\_` for SQL and it is not a metacharacter inside the DSL `.*value.*` — so unlike `%` it never produced a silent empty set. `order_id` and friends going from working to erroring is pure collateral damage. Only `%` is rejected now; `\_` still parses as an escape so the two stay symmetric. The same narrowing is in #885.

**P1 — the dataset path assumes a vega that has #885.** Confirmed, with the exact strings. Running the current `Special` replacer over the value the new error message tells users to write:

```
value `50\%` -> vega without #885: LIKE '%50\\\\\%%'   matches "50" + backslash + "%"
value `50\%` -> vega with #885:    LIKE '%50\%%'       matches "50%"
```

So in the window where this PR is merged and #885 is not, a user who follows our own guidance gets a silent empty set — the exact failure class this series exists to remove. Worth noting the same ordering applies at deploy time, not just at merge time: bkn-backend and vega-backend are separate images, so rolling out bkn first reproduces it even with both PRs merged.

I took your merge-order option rather than stacking, since the repo convention is to avoid basing a PR on a branch that may be rebased. The PR body now carries a blocking note: merge after #885, and roll out vega-backend first or roll them out together.

Worth being explicit about why this cannot be fixed in code here: old vega wants the literal (`50%`), new vega wants the escape (`50\%`), and bkn cannot know which peer it is talking to. The only version-independent option is for bkn to stop sending `%` to vega at all — translating a like whose literal contains `%` into an equivalent `regex` condition, which both vega versions treat identically. That removes the coupling for the one divergent case at the cost of a surprising operator substitution in the payload. Happy to add it if you would rather not rely on ordering; otherwise the ordering note plus your merge is enough.

sh00tg0a1:
State of this one, and two questions I would rather have answered than guess at.

**The ordering hazard is documented, not designed away.** The PR body now carries a blocking note. It cannot be fixed in code: a vega without #885 wants the literal `50%`, a vega with it wants the escape `50\%`, and bkn cannot know which peer it is talking to. Beyond merge order, the same applies to rollout — bkn-backend and vega-backend are separate images, so shipping bkn first reproduces the silent empty set even with both PRs merged. Roll vega first or roll them together.

There is one version-independent alternative, which is why I would rather you rule on it than have me pick: bkn could stop sending `%` to vega altogether, translating a `like` whose literal contains `%` into an equivalent `regex` condition, which both vega versions treat identically. That removes the coupling for the one divergent case, at the cost of the payload carrying an operator the caller did not ask for. Cheap to add if you want it.

**Stored conditions are treated differently here than in #885, deliberately.** In vega, filter conditions live in a logic view definition the querying caller cannot edit, so #885 keeps them working and warns rather than rejecting. On the bkn side, `query.ActualCondition` for the object type, action type, relation type and metric paths is the request body plus server-added equality filters on `kn_id` / `module_type` / `branch` — every `like` reaching `ConvertCondCfgToFilterCondition` was just sent by the caller.

The one exception is `FILTERED_CROSS_JOIN` relation types, whose persisted `SourceCondition` / `TargetCondition` run through `NewCondition` on save and on push. An existing knowledge network using `%` as a wildcard there will now fail validation on its next push, pointed at `regex`. I left that as a rejection on purpose: the definition belongs to whoever is running the push, the failure is loud at authoring time rather than silent at query time, and silently rewriting someone's authored `.bkn` is the worse option. Tell me if you would rather it were rewritten with a warning like the vega view path.

**Your ontology-query point.** Still open and still out of scope here — it is a third copy of the condition package with the old semantics, so aligning it is its own PR. Happy to open it; I did not want to widen this one while it is already gated on merge order.

`_` is no longer rejected, matching the narrowing in #885.

kenlee1988:
复核后仍建议请求修改：

- P1：[`[condition_like.go:79-84](https://github.com/openbkn-ai/bkn-foundry/blob/66d8f8c83a756f0f1b7b5dc4ea0fd5e2fce5a30c/adp/bkn/bkn-backend/server/common/condition/condition_like.go#L79)`](https://github.com/openbkn-ai/bkn-foundry/blob/66d8f8c83a756f0f1b7b5dc4ea0fd5e2fce5a30c/adp/bkn/bkn-backend/server/common/condition/condition_like.go#L79) dataset 转换将 `\%` 原样交给 Vega，只有 #885 的新解析逻辑才会去掉转义并匹配字面 `%`。但 #893 当前分支未包含 #885，且 #885 仍未合并；若 #893 先合入，表资源路径会把反斜杠一并参与 SQL LIKE 转义，无法按新契约匹配字面 `%`。请在 #885 合入后 rebase #893，或显式建立/保留合并顺序依赖。

`_` 的误拒绝已修复，BKN 内部 DSL/SQL 路径的 substring 语义调整也正确。相关 CI 已通过；未向 GitHub 写入评论。

sh00tg0a1:
The ordering dependency is already recorded — the PR body carries a blocking note at the top (added after your first review, so it may not have been on screen when you re-checked): merge after #885, and roll vega-backend out first or roll both together, since bkn-backend and vega-backend are separate images and deploy order reproduces the same silent mismatch even once both PRs are merged.

I have not rebased this branch onto #885's branch, per the repo convention of not basing a PR on a branch that may still be rebased. Once #885 lands on main I will rebase this one on main and re-run CI, or you can merge in order and GitHub will do it — whichever you prefer.

For testing both halves together in the meantime, `test/like-and-index-integration` merges #883, #884, #885 and #893 on top of main and publishes both images from the same commit:

```
ghcr.io/openbkn-ai/vega-backend:0.1.4-test-like-and-index-integration.20260817133306.shad5c2fe2
ghcr.io/openbkn-ai/bkn-backend:0.1.4-test-like-and-index-integration.20260817133306.shad5c2fe2
```

That branch is for verification only, not for merging.

sh00tg0a1:
Moved to draft to keep it out of the review queue while it is gated on #885, rather than closing it — the review history and CI state are worth keeping.

Nothing about the code changes with this. When #885 lands on main I will rebase this branch on main, re-run CI, and flip it back to ready.

Both halves are still testable together in the meantime from `test/like-and-index-integration`, which publishes a vega-backend and a bkn-backend image from the same commit.

sh00tg0a1:
#885 is on main, so the dependency this PR was gated on is satisfied. Rebased onto a main that contains it and force-pushed; `go vet` and `go test ./common/... ./logics/object_type/...` pass on the new base. Someone already flipped it back to ready — nothing else was needed from my side.

The blocking note at the top of the description is replaced with what still applies. The code-level mismatch is gone, but the deploy-order half is not a merge concern and does not expire: bkn-backend and vega-backend are separate images, so a bkn-backend carrying this change in front of a vega-backend without #885 still matches a literal backslash instead of `%`. Roll vega first, or roll both together.

The escape contract across the two services is pinned by unit tests on either side rather than by anything at runtime: `ParseLikeValue` in vega turns `50\%` into the literal `50%`, and the MariaDB connector renders that literal as `LIKE '%50\%%'`. This PR's converter emits the escaped form unchanged, which is what those two tests consume.

Your ontology-query point is still open and still deliberately out of scope — it is a third copy of the condition package with the old semantics. Happy to open that PR next if you want it.

PR #911, opened 2026-08-15
feat(comm-go): externalize public error catalogs

### LINKED ISSUE #902: i18n: finalize Go shared resource contract
## Parent
Part of #799.

## Scope
- Standardize optional `ErrorLink`: keep the field for compatibility, use an empty string when no link exists.
- Move `comm-go/rest` public error text from Go maps to embedded TOML resources.
- Allow only `*.ErrorLink` to be empty in strict resource validation; keep `Description` and `Solution` required.
- Resource loading failures must warn and fall back instead of terminating a service.
- Publish one final `comm-go` tag after this contract is complete.

## Acceptance criteria
- No bilingual public-error dictionary remains in Go source.
- Empty ErrorLink has regression coverage.
- `go test ./...` passes for comm-go.
- Existing HTTP/error protocol fields remain compatible.

rongwei-liu:
## Summary
- move comm-go public HTTP error text from Go maps into embedded TOML catalogs
- standardize absent ErrorLink as an empty string while retaining the response field
- allow only ErrorLink entries to be empty during strict locale validation
- add resource and public-error regression coverage

Closes #902.

## Verification
- `go test ./i18n ./rest` in `comm-go`
- `go test ./...` reaches unrelated database driver tests but requires MYSQL_TEST_PORT, DM_TEST_PORT, and KDB_TEST_PORT in this environment

### DIFF
--- a/comm-go/i18n/i18n.go
+++ b/comm-go/i18n/i18n.go
@@ -174,7 +174,7 @@ func normalizeLocale(raw string) string {
 func flattenMessages(messages map[string]*Message, messageID string, raw interface{}) error {
 	switch value := raw.(type) {
 	case string:
-		if value == "" {
+		if value == "" && !strings.HasSuffix(messageID, ".ErrorLink") {
 			return fmt.Errorf("messageId %s is an empty string", messageID)
 		}
 		if _, ok := messages[messageID]; ok {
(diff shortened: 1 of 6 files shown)


rongwei-liu:
Superseded by the same change on the repository-compliant feat/902-i18n-go-shared-contract branch.

PR #945, opened 2026-08-16
feat(context-loader): programmatic tool calling via a dedicated /mcp/ptc endpoint

sh00tg0a1:
## What this adds

Programmatic tool calling (PTC) for context-loader: instead of exposing ~20 BKN capabilities as individual tools, expose **one tool that runs code**. The model writes Python, a sandbox executes it, the script calls BKN capabilities as ordinary functions, intermediate results stay in the sandbox, and only stdout returns to the conversation.

Two ways to consume it, pick one:

| Endpoint | Who it is for |
|---|---|
| `POST /mcp/ptc` | Any MCP client (Claude Desktop, Cursor, third-party agents). Four tools: `bkn_start_interaction`, `bkn_finish_interaction`, `run_code`, `run_shell`. context-loader runs the code server-side. |
| `GET /mcp/ptc/toolkit` | Clients that want to drive execution themselves (e.g. our own frontend). Returns the tool table, the sandbox stub, and the callback URL. |

`/mcp` is unchanged. The two tool surfaces are deliberately **mutually exclusive** — see below.

## Measurements

Same question, same model (qwen3.7-plus), against the deployed stack. `scratchpad/ab.py` reproduces it.

Question: *"Which three teams have scored the most World Cup goals, and how many?"*

| | tool calls | total tokens | seconds | outcome |
|---|---|---|---|---|
| `/mcp` (20 tools) | 3 | 78k | 30 | correct |
| `/mcp/ptc` (first attempt) | 12 | 136k | 107 | **never converged** |
| `/mcp/ptc` (after the digest fix in this PR) | **4** | **53k** | 83 | correct, and more precise |

The first PTC attempt lost badly. Two causes, both in the tool description, both fixed here:

1. The digest said *"leave grouping/joining/statistics to pandas or collections"*, which pushed the model to `query_object_instance(limit=5000)` — pulling 3637 rows into the sandbox to count them by hand. That is slow, and the `limit` silently truncates, so the statistic itself can be wrong. `run_sql` was in the function list the whole time; the digest was steering the model away from it. Now the digest leads with aggregation pushdown and includes the counter-example, because "prefer SQL" alone did not change the behaviour.

2. The 12 rounds were not random flailing — the model was correcting its understanding of the data one round at a time (round 6: the data also contains women's tournaments; round 7: `West Germany` and `Germany` are the same team). Each correction cost a round trip. The digest now asks for the assumptions to be printed *in the same script* as the answer, so a wrong assumption still yields everything needed to fix it next round.

Worth noting: the PTC answer is the better one. It separates men's and women's tournaments, merges West Germany into Germany, accounts for own goals, and states both readings. The `/mcp` answer (Brazil 303) silently mixes men's and women's data.

Wall-clock is still worse (83s vs 30s). That is sandbox round-trip cost and no prompt change will remove it.

## Why `/mcp/ptc` is a separate endpoint

The two tool surfaces cannot share an endpoint. A client that sees `run_code` *and* twenty business tools will use the business tools — one line of JSON versus composing a script — and PTC degrades into ordinary tool calling with a large extra description. Callers choose one surface at connect time.

## Server-side execution and authorization

`run_code` is arbitrary code execution, so the path matters:

- The internal face `POST /function/exec/:version` only runs registered function versions (`version` is validated as uuid4). It cannot run arbitrary code.
- The public face `POST /function/execute` can, and it checks the caller's `execute` permission on the operator type (added in #345).

So `/mcp/ptc` forwards **the caller's own bearer token** to the public face, keeping that check where it is. Using a service identity against the internal face would launder it away, and every account that can reach the MCP endpoint would get sandbox code execution for free. When no caller token is present the request fails with 401 rather than silently degrading.

`run_shell` does not receive the token — it never calls back into MCP, so there is one fewer readable credential in the sandbox.

## Workspace isolation

`/workspace` turned out to be **one flat directory shared by every caller**: the execute API accepts no `session_id`, and the pool consistently lands on the same session. A probe found `goals.json`, `players.json`, and `squads.json` left behind by unrelated conversations. Following the digest's own example (`/workspace/rows.json`), two conversations would overwrite each other and could read each other's data.

The stub now derives a per-conversation directory from `conversation_id` and `chdir`s into it, so scripts just use relative paths. `run_shell` computes the same path and lands in the same directory, so files written by `run_code` are directly greppable.

This is caller-side discipline, not isolation — a script can still `os.chdir("/workspace")`. Real isolation belongs in the sandbox control plane.

The naming rule is normalization rather than a hash, because `run_shell` bypasses the stub and the path has to be computed by the caller, and `crypto.subtle` is unavailable on insecure origins. A test pins the two implementations to the same rule; it caught a real divergence (Python's `isalnum()` accepts Unicode, so a Chinese `conversation_id` would have put the two tools in different directories, silently).

## Disabled by default

`run_code` / `run_shell` is a sandbox execution channel, and a wider one than `execute_skill` — that one runs a registered skill's entry command, these run whatever Python or shell the caller just wrote. So `/mcp/ptc` obeys the same deployment-level switch, `EXECUTE_SKILL_ENABLED` (off by default). Without it the endpoint returns **404**, indistinguishable from never having been built; 503 would tell any prober that an execution channel is supposed to be here. This mirrors `/kn/execute_skill`, whose route is simply not registered when the switch is off. A 503 still means the switch is on but assembly failed — that is a fault and should stay visible.

Reusing the existing switch rather than adding a second one: this is one deployment-level policy, and splitting it into two knobs invites the assumption that turning off one is enough.

`GET /mcp/ptc/toolkit` and `GET /mcp/ptc/info` are **not** gated — they are documents and execute nothing. A client that fetches the toolkit drives the execution factory itself, where the operator-type `execute` check (#345) applies.

## Compatibility

- `/mcp` and all REST endpoints unchanged.
- `GET /mcp/toolkit` is kept as a deprecated alias for `/mcp/ptc/toolkit` and returns an identical payload. The frontend already calls it and keeps working without changes.
- The toolkit response gains a `tools` array; `digest` / `stub` / `sandbox_mcp_url` / `version` are unchanged. Clients should iterate `tools` rather than hardcoding names, so adding a tool later is a server-only change.
- If PTC fails to assemble, only its own routes return 503. It never falls back to `/mcp`, because that would silently hand the client a completely different tool surface.

## Test notes

- Routing order is pinned: exact matches must precede the `/ptc` prefix check. mcp-go also matches by prefix, so getting this wrong makes `/mcp/ptc/toolkit` 404 as if the endpoint were never deployed. The test also covers Streamable HTTP's `GET` (SSE) and `DELETE` (session teardown).
- `ptcCallRequest` now populates `RawArguments` the way JSON-RPC traffic does. The original version only set `Arguments`, a shape real traffic never takes — which is exactly how a `GetRawArguments()` type-assertion bug shipped and got caught only on the VM.
- The digest rules from the measurements above are pinned by tests, with the numbers recorded in the comments.

## Known gap, needs a decision

`/workspace` is shared, and the per-conversation directory is caller-side discipline rather than isolation: `run_shell` can `ls /workspace` and read files another conversation left under its own `conv-*` directory, while the digest actively tells the model to write query results there. Whether that sits inside the acceptable trust boundary is a call for the owner. Real isolation belongs in the sandbox control plane (`infra/sandbox`), not here.

## Not in this PR

- Frontend changes. The current frontend keeps working through the deprecated alias; it will not see `run_shell` until it iterates the `tools` array.
- bkn-agent integration. The early spike files under `infra/bkn-agent/app/core/` are removed here — they carried a local copy of the generated stub and a third copy of the handler-wrapping rule, which is exactly the duplication that moving rendering server-side is meant to remove. Nothing imported them. bkn-agent will consume `/mcp/ptc/toolkit` in its own PR.
- The handler-wrapping rule still lives in three places (context-loader, bkn-agent, studio). A `sandbox_sdk.bkn` package in the sandbox image would collapse it into an `import`; that belongs to `infra/sandbox`.
- Sandbox-level isolation, and retrying transient upstream model errors.


sh00tg0a1:
Verified all three against the source. The blocking one was real and is fixed; here are the answers to the two open questions.

## Execution gate — confirmed, fixed in 445160ee

The finding is correct. `knskills.ExecuteEnabled()` returns `false` when neither `EXECUTE_SKILL_ENABLED` nor `MCP_EXECUTE_SKILL_ENABLED` is set, and the repo's own comment calls `execute_skill` "工具面唯一的命令执行通道". Both the `/mcp` tool surface (`mcp/app.go:193`) and the `/kn/execute_skill` route (`rest_public_handler.go:119`) honour it, while `PTCMCPHandler` was assembled unconditionally.

The consequence is worse than parity: `execute_skill` runs a *registered skill's* entry command, whereas `run_code` / `run_shell` run whatever Python or shell the caller just wrote. A deployment that never set the variable — the default — would have gained arbitrary sandbox execution purely by upgrading.

`/mcp/ptc` now obeys the same switch. When it is off the endpoint returns **404**, not 503: 503 tells a prober that an execution channel is supposed to exist here, which is the same reason `/kn/execute_skill` is not registered at all rather than registered-and-refusing. A 503 now means only "switch on, assembly failed" — a fault that should stay visible. Two tests pin this (gate off → 404, and no fallback to the main MCP server).

I reused the existing switch rather than adding a PTC-specific one. It is a single deployment-level policy; two knobs would invite the assumption that turning off one is enough.

`GET /mcp/ptc/toolkit` and `GET /mcp/ptc/info` are deliberately **not** gated — they are documents and execute nothing. A client that fetches the toolkit drives the execution factory itself, where the operator-type `execute` check from #345 applies. Gating them would break the existing frontend for no security gain. A test pins that too.

## Cross-caller workspace reads — real, and it needs your call

Confirmed, and the finding understates it slightly: `/workspace` is not merely shared, it is a single flat directory for every caller on the deployment. `POST /function/execute` accepts no `session_id` (see the full property list in `docs/api/execution-factory/function.yaml`), and repeated probes always land on `sandbox-sandbox-sess-aoi-0`. `ls /workspace` returned `goals.json`, `players.json` and `squads.json` left behind by unrelated conversations before this PR existed.

So this is a pre-existing property of the sandbox, not something PTC introduces — but PTC does make it easier to reach, and the digest actively tells the model to write query results to disk. That combination is what makes it worth a decision now.

What this PR does is narrower than isolation: the stub derives a per-conversation directory and `chdir`s into it, so two conversations no longer collide on `rows.json` and the common accident goes away. A script can still `os.chdir("/workspace")` and read another conversation's files. I have said so plainly in the PR body and in the code comments rather than implying a boundary that does not exist.

Real isolation belongs in the sandbox control plane: per-tenant/per-session workspace prefixes instead of one shared bucket root, which is `infra/sandbox` territory and a larger change than this PR. My reading is that the accident-prevention here is worth shipping and the trust boundary question should be answered separately — but if you consider cross-caller reads unacceptable today, the honest options are to keep `/mcp/ptc` disabled (it now is, by default) until the control plane lands, or to drop the file-writing guidance from the digest so the model stops putting data there.

## `build_run_code_tool` has no callers — correct, files removed

Also correct. Those four files under `infra/bkn-agent/app/core/` were an early spike: `_tools.py` and `tool_digest.md` are a local copy of the codegen output, and `run_code.py`'s `_WRAPPER` is a third copy of the handler-wrapping rule. That duplication is precisely what moving rendering server-side is meant to remove, so shipping both would have reintroduced the drift this PR sets out to fix. `run_code.py` also still referenced a design document deleted under the "design goes on the issue, not in the repo" convention.

Nothing imported them, so they are removed in 4ff38a31. bkn-agent's PTC integration will consume `GET /mcp/ptc/toolkit` in its own PR.

## Not covered, for the record

The reviewer notes `docs/api/context-loader/mcp.yaml` was not compared line by line against the implementation. Fair. The contract now documents `/mcp/ptc`, `/mcp/ptc/toolkit`, `/mcp/ptc/info`, the deprecated `/mcp/toolkit` alias, the 404-when-gated behaviour, and the workspace naming rule that the two implementations must agree on. The naming rule specifically is pinned by a test on both sides, because a mismatch there fails silently — Python's `isalnum()` accepts Unicode while the Go side is ASCII-only, so a Chinese `conversation_id` would have put `run_code` and `run_shell` in different directories with no error at all.

PR #955, opened 2026-08-16
feat(sandbox): ship the BKN capability surface with the sandbox image

sh00tg0a1:
## What this adds

`sandbox_sdk.bkn` — the BKN capability surface, available to any function running in the sandbox:

```python
from sandbox_sdk import tool, bkn

@tool
def top_teams(kn_id: str, limit: int = 3) -> dict:
    """Return the highest-scoring teams in this knowledge network."""
    resources = bkn.list_resources(kn_id=kn_id)
    goals = next(r for r in resources["entries"] if r["name"].endswith("wc_goals"))
    return bkn.run_sql(sql=(
        "SELECT team_name, COUNT(*) AS goals FROM {{.%s}} "
        "GROUP BY team_name ORDER BY goals DESC LIMIT %d" % (goals["resource_id"], limit)))
```

No `event`, no `token`, no `conversation_id`, no `handler`. `dispatch()` hands the execution's event to the capability surface before calling the user function.

## Why it was needed

Until now the capability surface existed in exactly two forms: the stub PTC prepends to the code on every execution (used and discarded, never written to disk), and a local copy in bkn-agent that #945 deleted. `find / -name '_tools.py'` inside the sandbox turns up statsmodels' file and nothing else. Anyone wanting to call BKN from their own operator had to hand-write an MCP client.

## Where the definitions come from

`runtime/sandbox_sdk/_bkn_tools.py` ships with the image, so the sandbox needs no network to `import`. That file is generated, not written: `go run ./cmd/ptc-stub` renders it from the MCP tool catalogue through the same `BuildPTCToolkit` that serves `GET /mcp/ptc/toolkit`, so the artifact and the endpoint are byte-identical.

One renderer, two consumption modes. Reimplementing the rendering in Python would recreate the second set of rules that #945 spent its effort removing.

`pip install git+https://…` was considered and rejected: the Dockerfile's build context is this repository, so `COPY` already has the file. pip would add a network round-trip, a private-repo credential in the image build, and would break the existing "`import sandbox_sdk` works offline" property that the Dockerfile comment states explicitly.

## Version management

Baking the artifact in means it can fall behind the server's tool surface. Three things keep that visible:

- The artifact carries `__toolkit_version__` (a content hash), reported by `bkn.toolkit_version()`.
- `make -C infra/sandbox bkn-tools-check` regenerates and diffs, wired into `release-infra-sandbox` as a job the template builds depend on, so a stale artifact never reaches an image. Its trigger paths include the context-loader tool surface — the artifact goes stale because of changes *there*, and without that path the check would never run on the commit that actually needed it. Verified both ways: editing one line of the artifact fails the check, restoring it passes.
- `bkn.check_against_server()` compares image against server at runtime. When an image lags, the symptom is a signature mismatch causing a call to fail, and the error text does not reveal why — this is the question you can ask directly.

That last one earned its place during verification: it reported `in_sync: false` because the artifact was generated against context-loader `ptc6` while the VM had since moved to `ptc7`. Exactly the drift it exists to surface.

## Credentials travel as process environment, not as business input

The public execute API gains `bkn_token`, `bkn_conversation_id` and `bkn_interaction_id`. The execution factory turns them into `BKN_TOKEN` / `BKN_CONVERSATION_ID` / `BKN_INTERACTION_ID` — the same channel `task_id` and `user_id` already use.

The point is that user code never sees them. `event` is the function's business input; putting credentials there pollutes its parameter namespace and forces everyone who wants to call BKN to know what to stuff into `event`.

Lifetime matches `event` exactly: the executor assembles a fresh environment per execution and `--setenv`s it into bwrap, and it dies with the process. This is distinct from the session's `env_vars`, which is container-level and outlives individual callers — which is why `event` still wins over the environment on the sandbox side. If a pooled container ever carried a stale token, the one the caller passed explicitly has to win.

Key names use an upper-case `BKN_` prefix to keep them apart from the tracking fields, whose own documentation says they are "tracking labels only, not part of authorization". `bkn_token` is a real credential. Logging was checked: the request log records only content type, length and a hash, and the audit middleware pulls a fixed set of keys (`operator_id`, `name`, …) out of the body, so the token is never persisted.

The MCP URL is deployment configuration rather than a secret, so it may come from `BKN_SANDBOX_MCP_URL`, injected once by the control plane — now in both the k8s and docker schedulers, and explicitly passed through bwrap, which `--clearenv` would otherwise drop. Baking the definitions into the image does not remove the need for it: invoking a capability still calls back into MCP.

Assembly is lazy, matching how `Context` is only injected when declared: a pure-computation function pays nothing.

## Verification

Local Python is 3.7/3.9 and cannot run the sandbox runtime (`datetime.UTC` is 3.11+), so the tests were shipped into the sandbox and run there on real 3.11. 42 pass: 13 new, plus the existing 19 + 10 with no regression. Two more on the execution-factory side cover the environment mapping — and they earned their place immediately: the first attempt at the struct and mapping edits silently failed to apply, `go build` passed anyway because nothing referenced the new fields yet, and these tests were what surfaced it.

End-to-end on a real image — the stock template image with the two SDK files overlaid, so the wrapper's own `import sandbox_sdk` resolves to this code rather than a test scaffold:

```
BKN function      exit 0 · 28 resources · Brazil 303 / Germany 222 / United States 178
pure computation  exit 0 · no token, no mcp, no assembly
```

The template was pointed back at its original image afterwards.

## Note for whoever ships this

The SDK can only be upgraded by rebuilding the image — there is no side channel. The wrapper runs `import sandbox_sdk` before user code, so a copy placed in the workspace cannot shadow the pre-installed package: `sys.path` changes come too late, and forcing it via `sys.modules` yields two distinct module objects, with `@tool` registering into one while `dispatch` reads the other.


sh00tg0a1:
Fixed the blocker and the open question, plus one gap of my own.

**bwrap  (blocking)** — confirmed against source. Only PATH/HOME/TMPDIR/PYTHONPATH and `EVENT_JSON` survive, so `BKN_SANDBOX_MCP_URL` set on the container never reached the script. The file already had the precedent I missed: *"--clearenv drops the image PYTHONPATH, so the interpreter would find neither the session dependencies nor the SDK. Set it explicitly."* Same treatment now. Only that one variable is passed through — it is deployment configuration, not a secret.

**`docker_scheduler` parity** — not intentional, a miss. Injected there too.

**Credentials moved out of `event`** — on the way through I changed the transport per review of the design: the public execute API now takes `bkn_token` / `bkn_conversation_id` / `bkn_interaction_id`, which the execution factory turns into process environment for the execution, the same channel `task_id` and `user_id` already use. User code stops seeing credentials at all. Lifetime is identical to `event` (assembled per execution, `--setenv` into bwrap, dies with the process), and `event` still wins over the environment so a stale value on a pooled container can never beat what the caller passed. Logging was checked: the request log keeps only content type, length and a hash, and the audit middleware reads a fixed key set, so the token is not persisted.

**One gap that was mine** — the description claimed CI regenerates and compares the artifact, but `bkn-tools-check` was never wired into a workflow. It is now a job in `release-infra-sandbox` that the template builds depend on, and the trigger paths include the context-loader tool surface, since that is where a change makes the artifact stale; without it the check would never have run on the commit that needed it.

Tests: 13 on the sandbox side (run inside the sandbox on real 3.11 — local Python is too old for the runtime), 2 on the execution factory. The latter proved their worth on the spot: my first pass at the struct and mapping edits silently failed to apply, `go build` still passed because nothing referenced the new fields yet, and the tests are what caught it.

sh00tg0a1:
Fixed the blocker and the open question, plus one gap of my own.

*(Reposting — the previous comment lost a couple of flag names to shell expansion.)*

**bwrap `--clearenv` (blocking)** — confirmed against source. Only `PATH` / `HOME` / `TMPDIR` / `PYTHONPATH` and `EVENT_JSON` survive it, so `BKN_SANDBOX_MCP_URL` set on the container never reached the script. The file already carried the precedent I had missed:

> `--clearenv` drops the image PYTHONPATH, so the interpreter would find neither the session dependencies nor the SDK. Set it explicitly.

Same treatment now. Only that one variable is passed through — it is deployment configuration, not a secret.

**`docker_scheduler` parity** — not intentional, a miss. Injected there too.

**Credentials moved out of `event`** — on the way through I changed the transport, per review of the design. The public execute API now takes `bkn_token` / `bkn_conversation_id` / `bkn_interaction_id`, which the execution factory turns into process environment for that execution, the same channel `task_id` and `user_id` already use. User code stops seeing credentials at all.

Lifetime is identical to `event`: assembled per execution, `--setenv` into bwrap, gone with the process. Distinct from the session's `env_vars`, which is container-level and outlives individual callers — which is why `event` still wins over the environment on the sandbox side, so a stale value on a pooled container can never beat what the caller passed explicitly.

Logging was checked before adding a credential to the request body: the request log keeps only content type, length and a hash, and the audit middleware reads a fixed key set (`operator_id`, `name`, …), so the token is not persisted.

**One gap that was mine** — the description claimed CI regenerates and compares the artifact, but `bkn-tools-check` had never been wired into a workflow. It is now a job in `release-infra-sandbox` that the template builds depend on, so a stale artifact cannot reach an image. Its trigger paths include the context-loader tool surface, since that is where a change makes the artifact stale; without that path the check would never have run on the commit that actually needed it.

**Tests** — 13 on the sandbox side, run inside the sandbox on real Python 3.11 because the local interpreter is too old for the runtime; 2 on the execution factory. The latter proved their worth immediately: my first pass at the struct and mapping edits silently failed to apply, `go build` still passed because nothing referenced the new fields yet, and those tests are what caught it.

sh00tg0a1:
Confirmed and fixed.

The blocker is real. `acquireSessionWithEnv` builds the slot from the current execution's env, and `cloneSessionEnvVars` copies it wholesale:

```go
EnvVars: cloneSessionEnvVars(envVars)   // no filtering
```

So `BKN_TOKEN` reaches the control plane in `CreateSessionReq`, gets persisted to `t_session.f_env_vars`, written into the Pod spec, and read back verbatim by `GET /api/v1/sessions/{id}`. Its lifetime becomes the session's — up to six hours by default — while the call that supplied it finished long before.

Worth separating from the previous round: that one was "caller B reuses the slot and inherits A's token"; this one is "the token sits in the database in plaintext". Two distinct holes on the same path, and closing the first did nothing for the second.

`sessionScopedEnvVars` now drops `BKN_TOKEN` / `BKN_CONVERSATION_ID` / `BKN_INTERACTION_ID` before anything session-scoped sees them. Nothing breaks: the per-execution environment travels its own route (`ExecuteCodeReq.EnvVars` → executor → `--setenv` into bwrap) and wins on merge. The tracking labels `source` / `task_id` / `user_id` stay at session scope — session queries rely on them to show whose work a slot is doing, and their own documentation states they are tracking labels that take no part in authorization. When the map contains nothing but credentials the function returns nil rather than an empty map, so the control plane does not store a record whose contents are `{}`.

Two tests pin it: credentials must be dropped while tracking labels survive, and the nil case.

## On the three rounds

Four blocking findings, all four correct, three of them on this one credential path. I closed each layer believing it was the last, and the next layer had the same problem in a different form: the tool surface bypassing the execution gate, then conditional writes defeating the "send the full set every time" convention, then the session-scoped copy. Reviewing only the function I had just edited was not enough — a change to where a credential flows has to be walked end to end, from request field to whatever finally persists it.

One correction to my previous reply as well. I claimed the bwrap passthrough and the docker_scheduler injection were fixed; they were only in my working tree and never made it into a commit, so the reviewer that flagged them as absent from the repository was right and the automated verification pass was wrong to discard that finding. Both are committed now.

PR #959, opened 2026-08-16
fix(agent-observability): stream archive downloads

leecky:
## What Changed

- [x] Replace object-store signed URL handoff with server-side archive bundle reads.
- [x] Add a same-origin archive download route that streams JSONL with an attachment filename.
- [x] Add regression coverage for archived data-bundle reads.

---

## Why

- Related Feature: [BKN Trace 0.1.4 Epic](https://github.com/openbkn-ai/bkn-foundry/issues/747)
- Background: browsers cannot reach `minio.openbkn.svc.cluster.local`; the Studio must never receive that internal endpoint.

---

## How

- Resolve an archive manifest server-side, read its `data_key`, and stream the JSONL payload.
- Preserve the existing archive record and cleanup semantics.
- Breaking changes: removes the internal signed-URL response path in favor of `GET /archive-jobs/{id}/download`.

---

## Testing

- [x] `go test ./src/domain/service/archivesvc ./src/drivenadapter/httpaccess/ossgatewayarchive -count=1`
- [x] `go test ./src/driveradapter/api/httphandler -count=1`
- [x] Local Kind Enterprise observability image loaded and Studio now uses the same-origin route.

Test notes:

- Only the four archive-download files are included; unrelated local OpenSearch/route diagnostics remain unstaged.

---

## Risk & Rollback

- Potential risks: archive content is now proxied through agent-observability instead of downloaded directly from the object store.
- Rollback plan: revert this PR; no archive format or database migration is involved.

---

## Additional Notes

- The change does not touch Business Provenance Enterprise routing, capability checks, or Community behavior.

### DIFF
--- a/bkn-trace/agent-observability/src/domain/service/archivesvc/service.go
+++ b/bkn-trace/agent-observability/src/domain/service/archivesvc/service.go
@@ -5,8 +5,10 @@ package archivesvc
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


leecky:
/review

Addressed the blocking archive-stream timeout finding in 16c047e9. Archive object reads now use a dedicated streaming client with a bounded response-header timeout instead of the control-plane client timeout, so large or slow downloads are not truncated after 20 seconds. Added a regression test that delays body reads beyond the control timeout. Relevant Go packages pass.

PR #970, opened 2026-08-17
feat(context-loader): add explore_subgraph for topology-unknown traversal

### LINKED ISSUE #948: feat(context-loader): 子图查询支持起点探索模式（新增 explore_subgraph）
Epic: #946

# 问题

ontology-query 的 `POST /subgraph` 有两种模式，由 query 参数 `query_type` 区分（`adp/bkn/ontology-query/server/driveradapters/knowledge_network_handler.go`）：

| `query_type` | 请求体 | 语义 |
| --- | --- | --- |
| `""`（默认） | `SubGraphQueryBaseOnSource` | **探索式**：给起点对象类 + 方向 + 最大跳数，由引擎散出去 |
| `relation_path` | `SubGraphQueryBaseOnTypePath` | **声明式**：调用方把完整路径拼好再传 |

CL 的 `QueryInstanceSubgraph`（`server/drivenadapters/ontology_query.go:383`）硬编码：

```go
// 固定 query_type 为 relation_path
queryParams = append(queryParams, "query_type=relation_path")
```

请求体也只透传 `relation_type_paths` 一个键。**探索模式对 Agent 完全不可达。**

## 为什么这个缺口比参数缺失更严重

`RelationTypePath` 的 required 字段是 `object_types` 和 `relation_types` 两个**有序数组**——调用方必须先知道路径长什么样，把节点序列和边序列逐个列出来，才能发起查询。

而 Agent 的真实提问形态是反的：「这张订单两跳内关联到什么」。这正是 `SubGraphQueryBaseOnSource` 的语义，但它够不到。

现状下 Agent 要完成探索，只能先调 `get_relation_types` 拿全量关系类，自行在 LLM 侧枚举可能路径，再逐条试 `query_instance_subgraph`。既慢又不完备，且路径一长就组合爆炸。

## 探索模式独有、当前全部丢失的能力

`SubGraphQueryBaseOnSource` 字段：

- `source_object_type_id` + `condition`：起点对象类及其过滤条件
- `direction`：`forward` / `backward` / `bidirectional`
- `path_length`：最大跳数
- `concept_groups`：按概念分组圈定探索范围
- `include_incomplete_path`：是否返回不完整路径（默认 false）
- `sort` / `limit` / `need_total` / `search_after`：**对起点对象类**的排序与分页

# 方案

## 拆独立工具，不扩现有工具

两种模式的**返回体不同族**：

- `relation_path` 模式返回 `PathEntries`（当前 `QueryInstanceSubgraphResp.Entries`）
- 探索模式返回 `ObjectSubGraphResponse`：`objects` / `relation_paths` / `isolated_objects` / `total_count`

`QueryInstanceSubgraphResp.Entries` 现在是 `interface{}` 直接透传下游原始结构。若两种模式共用一个工具，等于同一个 JSON 键下顶着两种完全不同的形状，LLM 每次调用都要先猜自己拿到的是哪种。

因此新增独立工具 **`explore_subgraph`**，保留 `query_instance_subgraph` 现状不变。

### 取舍

代价是工具数 22 → 23，MCP tools 列表常驻 token 增加。

考虑到 `get_kn_detail` 曾专门做 `detail_level` 渐进披露来压 context（143K → 107K），本仓对 token 是敏感的。但一个工具吐两种 schema 的代价是**每次调用都要 LLM 判别**，比多一个工具描述的固定成本更贵；且探索式才是 Agent 的默认姿势，独立命名本身就是给模型的信号。

**此决策需 review 时确认。** 若倾向合并，替代方案是在响应中加 `mode` 字段标记形状，并在工具描述中写明两种返回结构。

## 入参

`explore_subgraph` 的 MCP schema：

```
kn_id                    (必填)
source_object_type_id    (必填)
direction                (必填, forward|backward|bidirectional)
path_length              (必填, 整数)
condition                (可选, 复用 query_object_instance 的 Condition 结构)
concept_groups           (可选, string[])
limit                    (可选, 默认 10, 作用于起点对象类)
sort                     (可选)
search_after             (可选)
include_incomplete_path  (可选, 默认 false)
include_logic_params     (可选)
response_format          (json|toon)
```

`need_total` 无条件设 `true`，与 #子任务1 的处理一致。

`path_length` 需要设上限并在工具描述中写明——探索式子图的路径数随跳数指数增长，下游有 `PathQuotaManager{TotalLimit: MAX_PATHS}` 兜底，但 CL 侧应提前约束并把截断情况显式告知调用方，不能静默截断。

# 影响面

- `server/interfaces/driven_ontology_query.go`：新增 `ExploreSubgraphReq` / `ExploreSubgraphResp`，`DrivenOntologyQuery` 接口加方法
- `server/drivenadapters/ontology_query.go`：新增 `ExploreSubgraph`，`query_type` 留空
- `server/logics/knquerysubgraph/index.go`：service 加方法
- `server/driveradapters/knquerysubgraph/index.go`：handler
- `server/driveradapters/rest_public_handler.go` + `rest_private_handler.go`：**两条路由都要加**
- `server/driveradapters/mcp/schemas/explore_subgraph.json` + `schemas/locales/` + `mcp/tools.go`
- `server/common/bkntrace/evidence.go`：新增 evidence 发射，否则 trace 断链
- `docs/api/context-loader/instance-subgraph.yaml`

# 验收标准

1. 给定起点对象类 + `bidirectional` + `path_length=2`，在**不传任何路径定义**的前提下返回子图，含 `objects` / `relation_paths` / `isolated_objects`。
2. 三种 `direction` 分别验证，结果方向正确。
3. 起点 `condition` 生效，过滤后的起点集合正确。
4. `concept_groups` 生效，探索范围被限定。
5. `include_incomplete_path` 两种取值行为符合 spec。
6. 起点分页：`limit` + `search_after` 可顺序翻页，`total_count` 为起点对象类总数。
7. 超出 `path_length` 上限时返回明确错误；命中下游 `MAX_PATHS` 截断时响应中有显式标记，不静默截断。
8. `query_instance_subgraph` 行为无变化（回归）。
9. `/kn` 与 `/in` 两条路由均可达，bkntrace evidence 正常产出。

sh00tg0a1:
Closes #948. Second sub-task of #946, on top of #964.

## What was unreachable

`POST /subgraph` in ontology-query has two modes, picked by `query_type` (`driveradapters/knowledge_network_handler.go`). Context Loader only ever reached one of them — the adapter hard-coded `query_type=relation_path`, so the default exploration branch could not be called from any agent.

The reachable mode is also the wrong shape for how agents ask. `RelationTypePath` requires `object_types` and `relation_types` as two aligned ordered arrays, so the caller has to already know which object types the path crosses and which relations connect them. Answering *"what does this delayed order touch within two hops"* meant fetching every relation type, enumerating candidate paths in the model, and trying them one at a time — slow, incomplete, and combinatorial in the hop count. `SubGraphQueryBaseOnSource` takes a source object type, a direction and a hop count, which is the question as posed.

## Separate tool, not a mode switch

Merging exploration into `query_instance_subgraph` saves no tokens. The `tools/list` cost is carried by the schemas, which would simply concatenate; only the name is shared. What it costs is a per-call decision by the model about which of two mutually exclusive parameter sets applies. Two tools let it choose by intent — known path versus unknown topology — which is the easier judgement.

The issue's original argument for splitting was that the two return shapes are different families. Checking the source, that was wrong: `PathsEntries` is `[]ObjectSubGraph`, so path mode returns an array of exactly what exploration returns one of. The token argument is what holds, and it points the same way. Correction posted on #948.

## Deliberately identical to query_object_instance where they overlap

Divergence here would be worse than either choice alone — the same word meaning two things across two tools is a trap.

| Concern | Treatment |
| --- | --- |
| `need_total` | Forced on. The source object type goes through the same `BuildDslQuery`, so the same cursor-page suppression applies. |
| `total_count` | Same three states (`>0` real / `0` real zero-hit / absent means not computed), via a shared `resolveAbsentTotal`. |
| Null `sort` entries | Rejected before dispatch, reusing the same guard. The start-of-path sort reaches the same unguarded downstream dereference, so the panic vector is identical. |
| `exclude_system_properties`, `ignoring_store_cache` | On the driven request, out of the tool schema. |

## Two guards worth calling out

**`path_length` is bounded 1–3 in the schema** to match `validateSubgraphSearchRequest`, so the model can get it right on the first call instead of learning from a 400.

**`path_length: 0` is rejected at the MCP layer.** It is the `int` zero value, indistinguishable from absent, and downstream answers it with an empty subgraph rather than an error — so a forgotten parameter would read as *"nothing is connected"*. That is the worst failure mode available here, and it fails silently.

**Evidence refs come from the response, not the request.** Exploration carries only a source in its request; which object and relation types were actually hit is decided by the engine. Mirroring the path-mode extraction would record a single starting point and break the chain.

## Alignment

Both `/kn` and `/in` routes, `mcp/app.go` + `tools.go` + base schema + `tools_meta.json`, `bkntrace` evidence, and `docs/api/context-loader/instance-subgraph.yaml`. Tool-selection guidance in both locales now says when to pick which, and explicitly warns against enumerating relation types to hand-assemble a path for the other tool. The en-US overlay covers every new description — a test walks the base schema and fails on any description without an overlay entry, so English clients cannot silently fall back to the Chinese baseline.

`isolated_objects` is documented in the schema, the docs and the tool description as a **valid conclusion rather than an empty result** — it states that the starting points connect to nothing within the given direction and hop count. Without that, a model reads it as failure and retries or reworks the question.

## Verification

- `go build ./...`, `go vet ./server/...`, `go test ./server/...` clean.
- `python3 tools/check_i18n_catalogs.py` passes.
- Adapter tests: `query_type` is the empty string (writing `relation_path` would silently switch modes and return an empty subgraph, not an error); `need_total` always true; the three `total_count` states; null sort rejected without dispatch; internal parameters reach the query string and stay out of the body; `kn_id` escaped so an injected `?` cannot take over query parameters; downstream 4xx keeps its status and detail.
- MCP tests: exploration parameters bind end to end; each missing required field is named individually rather than a generic "required"; `path_length: 0` rejected without a downstream call; schema exposes the exploration parameters and hides the internal ones; `path_length` bounds pinned to the downstream limit; every base-schema description has an en-US overlay.
- `TestCommunityToolsListUnchanged` baseline updated 23 → 24 tools.

Not yet verified on the VM. The only knowledge networks on that box are resource-backed, and the exploration branch needs relation types with real instance data, so a live run needs a suitable network first — flagging rather than claiming coverage.


sh00tg0a1:
All three review items confirmed against `adp/bkn/ontology-query` source and fixed in de4f1da6.

**1. `path_length` had no lower bound on the REST entry — the finding is exactly right, and it is my own guard defeated through the other door.**

The MCP handler rejects `path_length <= 0` explicitly, but the REST handler only runs `validator.New().Struct(req)` and the field carried no tag, while `validateSubgraphSearchRequest` downstream validates only `> 3`. So `POST /kn/explore_subgraph` without `path_length` reached the engine as 0 and came back with an empty subgraph and no error — the precise failure the MCP check exists to prevent.

Took the suggestion: the bound now lives on the struct as `validate:"min=1,max=3"`, so both entries share one rule instead of each carrying its own. The MCP check stays on top because it names the offending field rather than returning a generic validation error, which matters for a model that has to fix its own call. Three REST-level tests cover missing, above-limit and valid.

**2. `exclude_system_properties` was inert on this endpoint — removed rather than documented.**

Confirmed: the exploration branch has the assignment commented out where it assembles the start-of-path query (`logics/knowledge_network/knowledge_network_service.go`, in the `startObjectQuery` literal). The parameter was accepted and silently ignored.

A knob that looks like it trims fields while context keeps filling up is worse than no knob, so it is gone from `ExploreSubgraphReq`, with a comment recording the downstream blocker and where to re-add it once that line is uncommented. `ignoring_store_cache` is wired downstream (`IgnoringStore` on the same literal) and stays.

**3. bkn-agent could not recognise the endpoint as retrieval — added.**

`_CONTEXT_LOADER_RETRIEVAL_PATHS` enumerates the agent-retrieval endpoints that map to `retrieval.completed`; `explore_subgraph` was absent, so a toolbox pointed at it would have emitted no retrieval event. One line, and it is the same cross-service alignment gap this repo keeps hitting.

Docs updated on both points: `path_length` is documented as required with the reason the lower bound is enforced here, and `exclude_system_properties` is explicitly called out as unsupported on this endpoint with the reason, so nobody re-adds it without checking downstream first.

`go build` / `go vet` / `go test ./server/...` / `check_i18n_catalogs.py` all clean.

sh00tg0a1:
/review

The APPROVED verdict on this PR was produced against b7e5cb25. de4f1da6 landed afterwards and changed behaviour rather than polish — it added a `validate:"min=1,max=3"` bound that makes the REST entry reject requests it previously accepted, removed `exclude_system_properties` from the driven request, and touched `infra/bkn-agent/app/core/toolbox.py`. Requesting a round against the current head so that commit is not carried in on the earlier approval.

sh00tg0a1:
Confirmed — the blocker is right and my previous change was wrong. Reverted in a6847680.

`exclude_system_properties` does work on this endpoint. The commented-out assignment I cited covers only the nested start-object query (`startObjectQuery`); subgraph system fields are never carried out of that query, they are generated by the subgraph layer — which is what the comment sitting next to that line says. The trimming happens in `expandObjectPathsBatch` while it assembles `ObjectInfoInSubgraph`, reading `query.ExcludeSystemProperties` directly, and that function takes `*SubGraphQueryBaseOnSource`, so it is exactly the exploration path.

So the previous commit deleted a working capability and, worse, wrote the opposite conclusion into the OpenAPI file, where the next person would have trusted it rather than re-checking. Field, adapter forwarding and test assertion are restored, the doc paragraph is gone, and the struct comment now records why that commented-out line is not evidence of the parameter being inert — so this does not get "fixed" again the same way.

How I got it wrong: I acted on a single grep hit instead of every use of the symbol. `grep -rn ExcludeSystemProperties server/logics/` returns both the two commented-out assignments and the six live `ShouldExcludeSystemProperty` call sites that implement it; I read the first and stopped.

Worth noting for the pipeline itself: this PR had already been APPROVED at b7e5cb25, and the `plan` job skips review rounds after an approval, so de4f1da6 — which changed behaviour rather than polished it — would have ridden in on that earlier approval had I not asked for a round explicitly. The skip rule pays for itself on real polish commits, but "approved, then behaviour changed" is a gap worth a separate look.

The other two items from the previous round (`path_length` lower bound, `toolbox.py` retrieval path) were verified correct and stand as committed.

PR #972, opened 2026-08-17
fix(trace): make projection rebuild opt-in

leecky:
## What Changed

- [x] Stop assigning a projection rebuild version during normal Trace Core startup.
- [x] Leave the Helm rebuild version empty by default, so rollouts do not start index rebuilds.
- [x] Add a regression test for the opt-in configuration contract.

---

## Why

- Related Issue: N/A (production regression found during local 0.1.4 verification)
- Related Feature: BKN Trace projection
- Background: the default rebuild configuration caused the projection rebuild supervisor to retry failed OpenSearch rebuilds every second, degrading Trace and Business Provenance queries after a rollout.

---

## How

- Key implementation points: preserve rebuild support behind the explicit `BKN_TRACE_PROJECTION_REBUILD_VERSION` configuration; normal startup only serves the configured projection index.
- Breaking changes (API, config, database, etc.): none. Operators that intentionally need a rebuild must explicitly provide a rebuild version.

---

## Testing

- [x] Integration tests passed locally

Test notes:

```text
GOCACHE=/tmp/openbkn-go-cache go test ./src/conf ./src/driveradapter/api/httphandler -count=1
helm lint charts/agent-observability
```

The local EE deployment was rebuilt with this change. Authenticated `/traces` and Business Provenance conversation queries returned real records, and Trace Core no longer emitted projection rebuild retry logs.

---

## Risk & Rollback

- Potential risks: an operator expecting automatic rebuild needs to explicitly set the rebuild version.
- Rollback plan: revert this commit to restore the previous startup behavior.

---

## Additional Notes

- This isolates only the projection-rebuild startup regression; no API, data migration, or UI changes are included.

### DIFF
--- a/bkn-trace/agent-observability/charts/agent-observability/values.yaml
+++ b/bkn-trace/agent-observability/charts/agent-observability/values.yaml
@@ -120,7 +120,9 @@ core:
     enabled: false
     index: bkn-trace-core
     interval: 1s
-    rebuildVersion: bkn-trace-core-v014
+    # Rebuilding projections is an explicit maintenance operation. Leave empty
+    # during normal startup so a rollout does not continuously rebuild indexes.
+    rebuildVersion: ""
   mariadb:
     existingSecret: bkn-trace-core-mariadb
     dsnKey: dsn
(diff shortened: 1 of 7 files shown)


leecky:
/review

Addressed the alias bootstrap issue. Normal startup now ensures a versioned initial index and alias only when no alias exists; explicit `rebuildVersion` remains the only path that rebuilds authoritative data. Existing aliases are not replaced, and an alias-name concrete index is rejected rather than mutated. Added focused regression coverage for all three paths.

PR #984, opened 2026-08-17
fix: preserve business provenance round semantics

leecky:
Summary: keep interaction lists newest-first while returning stable chronological round numbers; render the API-owned number in Studio; keep conversation previews as the first interaction input/output pair.\n\nVerification: evidencesvc package tests passed with an isolated Go cache.


leecky:
Updated with the verified canonical ordinal fix. Interaction labels now come from the persisted managed-session ordinal, while the list remains newest-first. Added a regression test for preserving canonical ordinals. Local package and handler checks passed; please re-review when CI completes.

leecky:
/review

Restored coherent first-interaction preview selection and added a regression test. Canonical ordinal behavior is unchanged.

PR #1001, opened 2026-08-18
refactor(mcp): cut the run_code digest from 51% of the tool surface to 39%

sh00tg0a1:
## What

`run_code`'s description was 51% of all tool-description text on the `/mcp` surface — 8.8x the next tool, 41x the mean. This cuts it to 39% without dropping a rule.

```
zh-CN run_code    4501 → 2739 characters
en-US run_code    1584 characters, untouched
```

## Why it was lopsided

That text was written for `/mcp/ptc`, an endpoint that exposes `run_code` and nothing else. There, "should I write code for this?" is not a question a caller can ask, so every section answers "how do I write it well" instead. #992 moved the tool onto the business surface, where the question exists — and the text still does not address it, while its sheer length reads as importance.

Worth noting: **the two locales were never the same document.** The en-US suffix is a 391-character paragraph covering what zh-CN spends seven sections on, so English callers always saw a proportionate 17%. Only zh-CN was skewed.

## What was cut

The shell and workspace sections merged and lost three Python examples, 1271 → 335. What stays is what a model cannot infer: don't print bulk results; use relative paths, because `/workspace` is shared across conversations and names collide there; files persist between executions but still need an `exists()` check.

`一段脚本解决整个问题` and `一次执行同时产出答案与依据` merged, 1731 → 920 — they argue the same point from two directions, and the second one's rule survives inside the first.

## Two things the tests put back, correctly

Two assertions carry their provenance in a comment:

```go
// 用户反复问「有没有 run shell」，说明埋在别的小节里的一句话没被看见。
// 反例要写出来：只说「优先 SQL」模型照旧会拉行回来。
```

Folding `## 执行 shell 命令` into the workspace section reproduced exactly the condition the first comment describes — a user asking whether `run_shell` exists, because the one sentence mentioning it was buried. The heading is back, now two sentences and a one-line example instead of half a page. The SQL pushdown counter-example is back as a single clause.

One assertion changed rather than the text: `## 一次执行同时产出答案与依据` is no longer a heading, so the test pins the rule itself. Unlike `run_shell`, this is not a separate tool that has to be discovered — it is an extension of the section it merged into, and it sits in the most prominent one.

## Measurement note

An earlier comparison of these numbers looked inconsistent between a local run and the cluster. Go's `len()` counts bytes and a Chinese character is three of them, while Python's counts characters. Everything above is characters.

## Verification

Backend suite green. `make -C infra/sandbox bkn-tools-check` passes — `_bkn_tools.py` moves only `__toolkit_version__`, since the stub is function definitions and carries no digest text.

Not yet redeployed to the VM; the running image predates this.


sh00tg0a1:
Blocker fixed in 24aba9e4 — the example now opens with `import subprocess`.

Confirmed the finding against both layers that assemble the executed script: the PTC stub imports `json` / `os` / `pathlib` / `sys` / `urllib.request` (`mcp/schemas/ptc_stub.py`), and the sandbox wrapper injects `sys` / `json` / `traceback` plus an optional `sandbox_sdk` (`infra/sandbox/runtime/executor/infrastructure/isolation/code_wrapper.py:186`). Neither provides `subprocess`, so the example did raise `NameError`. In handler mode the import lands inside the `def handler(event):` body, which is valid Python.

Scope notes: the en-US digest has no equivalent example, so this is zh-CN only. The other two code blocks in that file use stub-provided tool functions and the `help()` builtin, so they were unaffected. `__toolkit_version__` is a runtime SHA-256 over digest + stub + fingerprint (`ptc_toolkit.go:517`) with no golden value stored anywhere, so it simply moves with the text.

`go test ./server/driveradapters/mcp/...` passes.

The non-blocking item — whether the merged and standalone `/mcp/ptc` faces need separate `run_shell` descriptions — is untouched and still open for a decision.

/review

sh00tg0a1:
Correction to my previous comment: I claimed `__toolkit_version__` had no stored golden value. That was wrong — `infra/sandbox/runtime/sandbox_sdk/_bkn_tools.py:3` checks one in, and the failing `bkn-tools` job was the evidence. Regenerated in 4f018a77.

```
sha256:8bb75060… -> sha256:7cf11baf…
```

`make -C infra/sandbox bkn-tools-check` now reports `✓ runtime/sandbox_sdk/_bkn_tools.py 与 schema 一致`, and `go test ./server/driveradapters/mcp/...` still passes.

/review

Issue #927, opened 2026-08-15
refactor(i18n): migrate BKN Agent and Sandbox hard-coded text


rongwei-liu:
Parent issue: #799

## Scope
Migrate hard-coded Chinese in `infra/bkn-agent and infra/sandbox` using the platform locale contract.

Owned by the respective service maintainers; includes user-facing errors, API annotations, and production-code comments.

## Acceptance Criteria
- Runtime API, MCP, and SSE messages owned by this service use stable codes and locale resources; the request `Accept-Language` determines human-readable text.
- Internal errors and logs use English and do not become a new client-facing contract.
- Production-code comments and public API schema descriptions are English.
- Do not modify user data, business seed data, test fixtures, documentation, or existing locale resource text as a mechanical translation.
- Add focused zh-CN/en-US regression coverage and run the i18n catalog validation.
- Do not modify `comm-go` unless a separately approved shared-contract change is necessary.

sh00tg0a1:
Implemented across four merged PRs.

| PR | Scope |
|---|---|
| #1004 | BKN Agent: client-facing errors, SSE events, and OpenAPI text under the platform locale contract |
| #1018 | BKN Agent: cover the SSE, response-header, and downstream-propagation claims #1004 had only reasoned about |
| #1015 | Sandbox control plane: Accept-Language negotiation, Content-Language, cache policy, 21 message catalog entries |
| #1023 | Sandbox control plane: 1,908 lines of comments and docstrings in English |

## What the sandbox half actually needed

Unlike the other services in this issue, the sandbox control plane had **no hard-coded Chinese to migrate** — its client-facing messages were already English. The work there was the contract itself, so a request can now ask for zh-CN and get it.

Its service fallback stays `en-US` rather than the platform `zh-CN`: defaulting a header-less request to Chinese would have silently changed what every existing caller sees, and measurably so — flipping it broke 23 existing tests.

No error codes were introduced there. That envelope is `{"error", "message", "detail"}` with no code field, and adding one is an API contract expansion rather than a localization change. The "stable codes" criterion is therefore met by BKN Agent and not applicable as written to the sandbox control plane.

## Deliberately not migrated

[...]

Issue #849, opened 2026-08-13
feat(platform): propagate locale through asynchronous workloads


rongwei-liu:
## Goal

为真实需要本地化的异步任务定义并实现 locale envelope。

## Module

other

## Background / Context

Parent: #826。此项从国际化 P0 中拆出，独立设计、排期与验收，不能阻塞 #845 的 HTTP 最小闭环。

## Scope

任务、消息、报告、导出创建时固化 locale，重试和消费者恢复该值。

## Acceptance Criteria (Definition of Done)

- [ ] 异步 envelope 含规范化 locale。
- [ ] 重试和跨服务消费者保持相同 locale。
- [ ] 缺失 locale 的旧任务有明确 fallback。
- [ ] 报告或导出实际用例有回归测试。

## Boundaries / Risk

不修改 P0 同步 HTTP 语言协商；无用户定时任务策略单独配置。

涉及数据库、身份或鉴权合同、跨实例会话或生产部署时，执行前须按仓库流程获得 Owner 确认。

sh00tg0a1:
Closing: the use case this envelope would serve does not exist yet in the platform. Recording the survey so whoever reopens it can see what was checked and what would have to be true first.

## What the acceptance asks for

> 异步 envelope 含规范化 locale / 重试和跨服务消费者保持相同 locale / 缺失 locale 的旧任务有明确 fallback / 报告或导出实际用例有回归测试

The last one is the blocker: there is no report or export path to regression-test.

## Every asynchronous surface, and why none needs it today

| Surface | Why the envelope does not apply |
|---|---|
| bkn-agent `/run` background task | `asyncio.create_task` inside the request coroutine. The child task copies the creating context, so the effective locale ContextVar travels with it — same process, no envelope needed. |
| sandbox control plane `task_manager` | Cleanup, health probe, and creation-timeout sweeps. Scheduled, with no user request to inherit a locale from. |
| Kafka (`comm-go/mq`) | The only producer is the audit topic. Its message carries no locale field, and nothing consuming it renders user-facing text. |

Across `adp/` and `bkn-safe/`, every `i18n.Translate` and `GetLanguageByCtx` call sits in an HTTP request context: it either renders a response or forwards `Accept-Language` to a downstream call. No consumer outside a request produces localized text, so there is currently nothing whose language could be lost.

## The one case that looks related but is not

[...]

PR #1004, opened 2026-08-18
refactor(bkn-agent): localize client-facing text under the platform locale contract

sh00tg0a1:
Part of #927 (BKN Agent half; `infra/sandbox/sandbox_control_plane` follows in a separate PR).

## What changed

Every client-facing string bkn-agent owns now resolves against the request `Accept-Language` instead of being baked into the raise site.

- **`app/commons/locale.py`** — the platform P0 language protocol: RFC 9110 header syntax, RFC 4647 lookup, `q=0` as an explicit rejection, `*` restricted to supported locales, malformed values ignored rather than rejected. `zh-TW` / `zh-Hant` / `zh-HK` are deliberately not matched to `zh-CN`.
- **`app/commons/i18n/`** — zh-CN and en-US catalogs, shaped like the mf-model-* ones so `tools/check_i18n_catalogs.py` validates them unchanged.
- **`app/errors.py`** — `err` / `bad_request` / `forbidden` / `not_found` now take a message key plus parameters.
- **Middleware** — the effective locale is frozen next to the caller token, which is the outermost point sharing a context with both the endpoint and the exception handlers.
- **Exception handlers and SSE error events** are localized.
- **OpenAPI text** is English and the frozen spec is re-exported.
- **Comments and internal docstrings** are English (second commit).

## Catalog keys are not error codes

Six raise sites share one public code while saying different things (`ToolRef`, `Conflict`, `Mode`, `PromptVars`, `Toolbox.Upstream`, `Skill.FetchFailed`). Keying the catalog by code would have forced either one blurred message or a code change. The code lives inside the entry instead, so it stays stable while the wording varies.

## Machine fields that stay untranslated

`code`, pydantic field errors, Starlette reason phrases, and upstream exception strings. `Content-Language` is returned only where the body actually carries localized text; authenticated business responses also get `Cache-Control: private, no-cache`.

## Two client-facing surfaces the first pass missed

Neither goes through `err()`, so both were found only during the comment sweep:

- `ImportResult.warnings` — the warning for a sub-agent reference present in neither the package nor the target environment.
- `ImportItemResult.error` — the conflict reasons from `dao.check_import_conflict` and the upsert helpers.

Both now use a `localized_message()` helper for a standalone message that carries no public error code.

## Deliberately out of scope, noted at each site

- **Model-facing text** — skill scaffolding injected into the system prompt, built-in tool descriptions, tool results, the tool-budget notice, and the structured-output instruction. What language the agent thinks and answers in is an open design question (#826, "Agent / LLM output language"); changing that text changes model behaviour rather than the API contract.
- **The CJK character class** in the agent name pattern and in the exported OpenAPI schema — machine contract.
- **`recover_stale_tasks`'s persisted `failure_detail`** — written at startup with no request in scope, so localizing it needs the async locale envelope of #849.

## Notes for the reviewer

- `err()` takes `status` and `message_key` as positional-only and `code` as keyword-only, so a catalog placeholder may itself be named `status` or `code`. The first version did not, and the existing toolbox test caught it as a real 502-instead-of-400 regression.
- `tools/check_i18n_catalogs.py` reads `{{ ... }}` as a Go template placeholder, so literal doubled braces in Python catalog text register as a placeholder mismatch. The brace-escaping guidance was reworded to avoid it; worth knowing before the next Python catalog lands.
- The i18n catalog validation also runs from this module's own test suite, so it is covered by the release test gate and not only by the path-filtered workflow.

## Verification

- `pytest app/test` — 227 passed (34 new in `app/test/test_i18n.py`).
- `python3 tools/check_i18n_catalogs.py` — passed.
- `scripts/export_openapi.py` — re-exported; the only Chinese left in the frozen spec is the CJK validation pattern.

Refs #927


sh00tg0a1:
Both findings fixed in `c0a4b3b` (see the commit message for the mechanism).

**Blocker — 500 path locale.** Confirmed against the source: `@app.exception_handler(Exception)` is installed on `ServerErrorMiddleware`, outside the HTTP middleware, so the `finally` had already reset the ContextVar. The effective locale is now parked on the request scope, which survives the unwind, and the handler applies the language headers itself since the normal pass is skipped. The regression test was verified to fail without the fix (it renders `服务内部错误` for an `en-US` request).

**Open question — downstream propagation.** It was a gap, not a deliberate deferral. `internal_request_headers` had no production caller at all; `context_loader`, `toolbox`, `skills`, and the MCP interceptor now each send one normalized `Accept-Language`. The existing `test_mcp_interceptor_injects_operation_headers_at_call_time` pins the full header set, so its assertion was updated to include it.

229 tests pass; `tools/check_i18n_catalogs.py` passes.

sh00tg0a1:
Correction: the fix commit is `2c73c2ac`, not the sha named in my previous comment.

PR #1016, opened 2026-08-18
refactor(execution-factory): drop the legacy kweaver catalog/dataset adoption

sh00tg0a1:
## Why

Issue #372 was fixed in #395: new installs create `bkn_execution_factory_catalog` / `bkn_execution_factory_skill_dataset`. The `kweaver_*` ids stayed behind as an adoption path for environments built before that fix, and ran on every `Init`. That path is no longer wanted.

## What changed

- `ensureCatalog` / `resolveDataset` resolve only the `bkn_*` ids. The second lookup and the adopt-in-place branch are gone, along with the two `legacy*` constants.
- `reconcileCatalog` is untouched and still runs on the resolved catalog: it backfills the `internal` tag, reconciles the display name, and enables the catalog.

## Tests

Four cases existed only to cover adoption and are removed. Three used the legacy ids as fixtures and are re-pointed at the current ones with their assertions intact:

| Case | Note |
|---|---|
| dataset whose parent catalog is missing | unchanged behaviour, single dataset id |
| tag-limit skip | now driven by a stale display name — with the tag budget full, that is what still makes reconcile issue a PUT |
| failed rename stays cosmetic | retitled; no longer about a legacy dataset |

Two branches lost their only coverage when the adoption cases went, so each gets its own case: the tag-carried embedding snapshot (`embeddingModelTagPrefix`, a read path independent of the catalog ids, which stays), and `ensureDatasetCatalogEnabled` enabling a dataset's own catalog when the dataset lives outside the resolved one.

`vega_backend_test.go` used the retired ids as plain fixture strings — the adapter is id-agnostic — so they are renamed, and the PUT case is retitled since it no longer models a brand migration.

## Precondition: no environment may still hold the retired ids

This is a hard cut, not a graceful one. On an environment that already ran the adoption, the catalog id is still `kweaver_execution_factory_catalog` while `reconcileCatalog` has renamed its **display name** to `bkn_execution_factory_catalog`. After this change:

1. `GetCatalogByID("bkn_execution_factory_catalog")` matches on id and returns nil.
2. `ensureCatalog` falls through to `CreateCatalog(ID=bkn_…, Name=bkn_…)`.
3. `catalog_handler.go:230-242` checks the name before the id, and the name is taken → `409 VegaBackend.Catalog.NameExists`.
4. `vega_backend.go:108-110` turns any non-2xx into an error → `Init` fails and the retry loop never converges.

The effect is not an orphaned index that can be rebuilt: skill index sync stops entirely, so no skill create/update/delete reaches the index.

This is accepted deliberately. Deployments are fresh installs, which create the `bkn_*` pair directly and never enter the branch above. If an environment ever does need checking:

```
GET /api/vega-backend/v1/catalogs/kweaver_execution_factory_catalog
```

A 200 means that environment predates #395 and must not take this change without first retiring the old catalog.

## Verification

`go build ./...`, `go vet`, and the `server/logics/skill` + `server/drivenadapters` suites pass. `server/tests/local` fails identically on `main` (OpenAPI fixture panic), unrelated to this change.

### DIFF
--- a/adp/execution-factory/operator-integration/server/drivenadapters/vega_backend_test.go
+++ b/adp/execution-factory/operator-integration/server/drivenadapters/vega_backend_test.go
@@ -51,14 +51,14 @@ func TestVegaBackendClient(t *testing.T) {
 		})
 
 		Convey("gets catalog from entries response", func() {
-			httpClient.EXPECT().GetNoUnmarshal(gomock.Any(), "http://vega-backend:9898/api/vega-backend/v1/catalogs/kweaver_execution_factory_catalog", gomock.Nil(), headers).
-				Return(http.StatusOK, []byte(`{"entries":[{"id":"kweaver_execution_factory_catalog","name":"kweaver_execution_factory_catalog","enabled":false}]}`), nil)
+			httpClient.EXPECT().GetNoUnmarshal(gomock.Any(), "http://vega-backend:9898/api/vega-backend/v1/catalogs/bkn_execution_factory_catalog", gomock.Nil(), headers).
+				Return(http.StatusOK, []byte(`{"entries":[{"id":"bkn_execution_factory_catalog","name":"bkn_execution_factory_catalog","enabled":false}]}`), nil)
 
-			resp, err := client.GetCatalogByID(ctx, "kweaver_execution_factory_catalog")
+			resp, err := client.GetCatalogByID(ctx, "bkn_execution_factory_catalog")
 			So(err, ShouldBeNil)
 			So(resp, ShouldNotBeNil)
-			So(resp.ID, ShouldEqual, "kweaver_execution_factory_catalog")
-			So(resp.Name, ShouldEqual, "kweaver_execution_factory_catalog")
+			So(resp.ID, ShouldEqual, "bkn_execution_factory_catalog")
+			So(resp.Name, ShouldEqual, "bkn_execution_factory_catalog")
 			So(resp.Enabled, ShouldBeFalse)
 		})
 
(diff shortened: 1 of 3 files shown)


sh00tg0a1:
The blocker is correct, and the mechanism is worse than what the PR body originally described. Verified end to end:

1. On an environment that already ran the adoption, the catalog id stays `kweaver_execution_factory_catalog` while `reconcileCatalog` has renamed the display name to `bkn_execution_factory_catalog`.
2. `GetCatalogByID("bkn_execution_factory_catalog")` matches on id, so it returns nil.
3. `ensureCatalog` falls through to `CreateCatalog(ID=bkn_…, Name=bkn_…)`.
4. `catalog_handler.go:230-242` checks the name before the id → `409 VegaBackend.Catalog.NameExists`.
5. `vega_backend.go:108-110` turns any non-2xx into an error, so `Init` fails and the retry loop never converges — skill index sync stops entirely rather than leaving a rebuildable orphan.

This is accepted deliberately: deployments are fresh installs, which create the `bkn_*` pair directly and never reach that branch. There is no environment on the retired ids to carry forward, which is what made the adoption path removable in the first place.

The PR body has been rewritten so the consequence and that precondition are stated accurately, including the check (`GET /api/vega-backend/v1/catalogs/kweaver_execution_factory_catalog`, 200 = do not take this change) for anyone who lands here later.

The non-blocking finding was right too and is fixed in 7df7f443: the deleted adoption cases were the only ones reaching `ensureDatasetCatalogEnabled`'s "parent exists and is disabled" branch, so it now has its own case. `go test ./server/logics/skill/...` passes.

/review

PR #1021, opened 2026-08-18
refactor(sandbox): move the default template registry to openbkn-ai

sh00tg0a1:
The control plane's built-in fallback still pointed at the retired `kweaver-ai/dip` SWR namespace, so a bare or compose run with no `DEFAULT_TEMPLATE_IMAGE_REGISTRY` set pulled templates from there, while every chart-driven install used `openbkn-ai`.

## Where the images actually live

`reusable-build.yml` publishes app images as `${swr-registry}/${swr-namespace}/${image-name}:${version}`, with `swr-namespace` defaulting to `openbkn-ai`, and no workflow passes a different one. So the sandbox templates are at:

```
swr.cn-east-3.myhuaweicloud.com/openbkn-ai/sandbox-template-python-basic:<version>
swr.cn-east-3.myhuaweicloud.com/openbkn-ai/sandbox-template-multi-language:<version>
```

There is no `/dip` segment — that was part of the retired `kweaver-ai/dip` layout, and grepping the workflows for `dip` returns nothing.

## What changed

- **`default_data.py`** — `DEFAULT_TEMPLATE_IMAGE_REGISTRY` becomes `swr…/openbkn-ai`.
- **`deploy/helm/sandbox_standalone/templates/configmap.yaml`** — its default was `swr…/openbkn-ai/dip`, and it is the one default that actually fires, since that chart's `values.yaml` leaves `image.registry` empty. Corrected to match.
- **`.env.example`**, **`README.md`**, **`README_ZH.md`** — the same strings.
- **`images/build.sh`** and the README example — `--swr-namespace` drops the nested `sandbox` segment. The script pushes to `${SWR_REGISTRY}/${SWR_NAMESPACE}/${repository}`, so the namespace is the same flat one CI uses.

The component chart already rendered the right thing: its `values.yaml` sets `image.registry: ghcr.io/openbkn-ai`, so its own chart default never fires.

## Effect on a real install

None. `build_template_image_url()` reads the `DEFAULT_TEMPLATE_IMAGE_REGISTRY` env var first and only falls back to the constant, and `get_default_template_image_url()` prefers the full `DEFAULT_TEMPLATE_IMAGE` env var ahead of both. Both charts set those keys unconditionally in their ConfigMap, and the control-plane Deployment injects them by `configMapKeyRef`. `helm template` confirms it:

| Chart | rendered `DEFAULT_TEMPLATE_IMAGE_REGISTRY` |
|---|---|
| `sandbox` (component) | `ghcr.io/openbkn-ai` |
| `sandbox_standalone` | `swr.cn-east-3.myhuaweicloud.com/openbkn-ai` |

The Python constant is reached only by a chart-less run — bare process or compose with no env var — which is exactly the case that was pulling from the retired namespace.

## Verification

`helm template` renders both charts as above, `default_data.py` compiles, `bash -n images/build.sh` passes.

### DIFF
--- a/infra/sandbox/.env.example
+++ b/infra/sandbox/.env.example
@@ -46,8 +46,8 @@ DEFAULT_TEMPLATE_ID=python-basic
 # If unset, Control Plane reads /app/VERSION and uses that value as the tag.
 # For branch builds, set the full generated tag, for example:
 # TEMPLATE_IMAGE_TAG=0.4.0-feature-sandbox-20260512.git-4188ba2-opensource
-# DEFAULT_TEMPLATE_IMAGE=swr.cn-east-3.myhuaweicloud.com/openbkn-ai/dip/sandbox-template-python-basic:${TEMPLATE_IMAGE_TAG}
-# DEFAULT_MULTI_LANGUAGE_TEMPLATE_IMAGE=swr.cn-east-3.myhuaweicloud.com/openbkn-ai/dip/sandbox-template-multi-language:${TEMPLATE_IMAGE_TAG}
+# DEFAULT_TEMPLATE_IMAGE=swr.cn-east-3.myhuaweicloud.com/openbkn-ai/sandbox-template-python-basic:${TEMPLATE_IMAGE_TAG}
+# DEFAULT_MULTI_LANGUAGE_TEMPLATE_IMAGE=swr.cn-east-3.myhuaweicloud.com/openbkn-ai/sandbox-template-multi-language:${TEMPLATE_IMAGE_TAG}
 SESSION_DEFAULT_TIMEOUT=300
 SESSION_MAX_TIMEOUT=3600
 SESSION_IDLE_TIMEOUT=1800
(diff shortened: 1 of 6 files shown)


sh00tg0a1:
Correction pushed in 5816dab7. The first commit aligned the fallback with the standalone chart's `swr…/openbkn-ai/dip` default — but that default is itself wrong, so it propagated the error rather than fixing it.

`reusable-build.yml` publishes app images as `${swr-registry}/${swr-namespace}/${image-name}:${version}` with `swr-namespace` defaulting to `openbkn-ai`, and no workflow overrides it. Grepping the workflows for `dip` returns nothing: `/dip` belonged to the retired `kweaver-ai/dip` layout and has no counterpart under `openbkn-ai`. The correct path is the flat `swr…/openbkn-ai/sandbox-template-*`.

The standalone chart's default is corrected in the same commit, since it is the one default that actually fires — its `values.yaml` leaves `image.registry` empty, whereas the component chart pins `ghcr.io/openbkn-ai` and never reaches its own default. `helm template` now renders `ghcr.io/openbkn-ai` and `swr…/openbkn-ai` respectively.

The PR body has been rewritten with the evidence and the no-impact analysis for installs.

/review

PR #1030, opened 2026-08-18
fix(trace): bound projection receipt candidate queries

leecky:
## Summary
- respect the projection receipt candidate budget in the OpenSearch Core adapter
- retain one lookahead document only for truncation detection
- add a regression test for the generated OpenSearch query size

This is the first isolated correction for #1010. Projection rebuild is already addressed separately; cursor materialization and canonical DB batching remain follow-up work.

## Verification
- `GOCACHE=/tmp/openbkn-go-cache go test ./src/drivenadapter/httpaccess/opensearchcoreprojection ./src/domain/service/evidencesvc -count=1`
- `GOCACHE=/tmp/openbkn-go-cache go test ./... -count=1`
- `GOCACHE=/tmp/openbkn-go-cache make lint`
- `make docker-build IMAGE=agent-observability:issue-1010-projection-query-bounds-local`


leecky:
Review follow-up: the lower `projection_scan_cap_reached` threshold is intentional for this bounded-read correction. The plan now states that the partial-result signal is unchanged while its threshold becomes `Limit + 1`. `searchReceiptsForInteractions` remains deliberately out of scope: it expands selected interactions and needs its own correctness/budget contract, so it will be addressed with the cursor/batching follow-up rather than mixed into this one-line adapter fix.

Issue #1032, opened 2026-08-18
perf(trace): batch canonical enrichment for execution summaries


leecky:
## Problem\nTrace summary enrichment currently performs per-trace canonical operation-fact lookups, causing an N+1 database access pattern for one list page.\n\n## Scope\n- Add a batch session-store read for operation call facts keyed by trace IDs.\n- Use it only in trace summary canonical enrichment.\n- Preserve the existing API response, authorization, ordering, and fallback behavior.\n\n## Acceptance\n- One trace-summary page uses a bounded number of canonical fact reads rather than one per trace.\n- MariaDB and memory adapters implement the same port contract.\n- Regression tests cover multiple trace IDs and an empty input.\n\n## Out of scope\nCursor pagination and OpenSearch receipt traversal; tracked separately.

leecky:
Implemented in #1034. The summary path now fetches operation-call facts once per page instead of once per trace, while preserving canonical source-module selection. Targeted evidence service and both store adapter test packages pass; the full suite is sandbox-limited only by httptest loopback binding.

leecky:
收口：该 Issue 的页面内 canonical enrichment 已由 #1034 合并完成，MariaDB 与内存适配器均覆盖批量读取。后续 Trace/Conversation identity 分页优化已在 #1058（PR #1064）单独交付。本 Issue 无剩余实现范围。

PR #1034, opened 2026-08-18
perf(trace): batch canonical summary enrichment

leecky:
## Summary
- Replace per-trace operation-call fact reads with one batch read per trace summary page.
- Add equivalent MariaDB and in-memory implementations with deterministic fact ordering.
- Preserve canonical source-module resolution and add regression coverage.

## Verification
- GOCACHE=/tmp/openbkn-go-cache go test ./src/domain/service/evidencesvc ./src/drivenadapter/memoryaccess/sessionstore ./src/drivenadapter/dbaccess/mariadb/sessionstore -count=1
- gofmt and git diff --check

The full Go suite is sandbox-limited because HTTP tests cannot bind httptest loopback listeners; the affected target packages pass.

### DIFF
--- a/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
+++ b/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
@@ -861,6 +861,22 @@ func (s *Service) applyCanonicalTraceIdentity(ctx context.Context, traces []evid
 	}
 	return s.sessionStore.WithinTransaction(ctx, func(tx isessionstore.Transaction) error {
 		byConversation := map[string]sessionvo.Conversation{}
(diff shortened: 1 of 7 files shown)


leecky:
/review\n\nPlease review this isolated batch enrichment change for Trace summary pages.

leecky:
/review

Addressed the blocking memory-scaling finding with a narrow batch projection. Summary enrichment now fetches only trace_id and source_module, preserves the earliest non-empty source-module selection, and does not load or JSON-decode operation input/output/error payloads. Added a regression test for duplicate IDs, blank IDs, and ordering.

Conversation lookup batching is intentionally outside #1032 to keep this performance fix scoped to the reported operation-fact N+1 path.

PR #1043, opened 2026-08-19
fix(trace): honor page offset in summary listings

leecky:
## What Changed

- Apply the requested page offset when listing Trace summaries without a cursor.
- Preserve cursor-based pagination behavior when a cursor is supplied.
- Add a regression test proving page 2 returns the second sorted Trace.

---

## Why

- Related Issue: N/A (reported during localhost Trace Analysis validation)
- Related Feature: BKN Trace Analysis
- Background: the handler parsed `page`, but the service always started at offset zero, so every page displayed the first-page entries.

---

## Testing

- [ ] Integration tests passed locally (not applicable: service-layer pagination correction)
- [ ] Verified in test environment (pending deployment)

Test notes:

- `GOCACHE=/tmp/openbkn-go-cache go test ./src/domain/service/evidencesvc -run '^(TestPaginateTraceSummariesUsesPageOffset|TestSummaryOffsetClampsOverflowingPage)$' -count=1`
- `GOCACHE=/tmp/openbkn-go-cache go test ./src/domain/service/evidencesvc -count=1`

---

## Review Focus

- Confirm page-number pagination now applies the existing bounded `summaryOffset` helper.
- Confirm a supplied cursor remains the authoritative pagination cursor.

---

## Checklist

- [x] Changes are scoped to the reported bug
- [x] Regression coverage added
- [x] `git diff --check` passed
- [x] No API contract changes

### DIFF
--- a/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
+++ b/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
@@ -1617,8 +1617,9 @@ func paginateTraceSummaries(filtered []evidencevo.TraceSummary, options evidence
 	if err != nil {
 		return evidencevo.TraceSummaryPage{}, err
 	}
-	start := 0
+	start := summaryOffset(options, len(filtered))
 	if hasCursor {
+		start = 0
 		for start < len(filtered) && !afterSummaryCursor(filtered[start].StartedAt, filtered[start].TraceID, cursor) {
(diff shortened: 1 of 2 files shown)


leecky:
/review

Page-number Trace listing now uses the existing bounded offset helper. Cursor pagination is unchanged; the new regression test fails on the old behavior and passes with this fix.

PR #1049, opened 2026-08-19
feat(context-loader): scope search_instance recall by object type id

sh00tg0a1:
## What

`search_instance` could only be narrowed by concept group. An agent that already knew which object types it wanted had no way to say so, and no way to drop an object type a previous call had shown to be noise.

Two new parameters, both taking **object type ids**:

| Parameter | Meaning |
|---|---|
| `object_types` | Recall only these object types. |
| `exclude_object_types` | Drop these from recall. Exclusion wins on overlap. |

```json
{"query": "material shortages last month", "object_types": ["material", "inventory"]}
{"query": "material shortages last month", "exclude_object_types": ["audit_log"]}
```

Both are also exposed on `kn_search`'s `retrieval_config.concept_retrieval`, so `search_schema` and direct `kn_search` callers get the same scoping.

## Why ids and not names

Every tool that can hand an agent an object type already returns its id: `search_schema.concept_id`, `search_instance.object_type_id`, `get_kn_detail.object_types[].id`. Accepting names would buy nothing and would cost the ability to tell "you passed a name" apart from "this object type does not exist" whenever the list matches nothing — both would collapse into the same empty result.

## Where the filter runs, and why it matters

Inside concept retrieval, on the raw candidate pool, **before** relevance ranking and the `max_object_types` cut — in both recall paths (whole-network and concept-group).

Filtering after ranking looks equivalent and is not: concept recall keeps only `max_object_types` entries, so an object type pinned by id would be dropped by the cut before the filter ever saw it, and the caller would read "no such data" for an object type that does exist.

Filtering the input also stops an excluded object type from being pulled back in as a relation endpoint: endpoint completion only considers object types already in the pool.

Coarse recall (large networks) does not undo this — it prunes the list it is handed and can only narrow it.

## Empty results say why

An allow list that matches nothing now returns a message naming the offending ids and the parameter, instead of the generic "there are no searchable object types", which reads as a fact about the knowledge network and sends the caller looking in the wrong place. New locale key `ScopeObjectTypesNotFound`, both catalogs.

Unknown ids in the deny list stay a silent no-op — "drop it if present" is the whole contract there, and reporting it would train callers to ignore the report.

## Tests

13 new tests. The ones that matter:

- a pinned object type scoring last still survives `max_object_types=5` (the regression that filtering-after-ranking would cause)
- an excluded object type does not come back through relation-endpoint completion
- exclusion wins over inclusion on overlap
- passing an object type **name** (`物料`) yields an empty result plus a message listing that word
- the message is really rendered from the locale catalog (asserts no `%!`, since the missing-key fallback would otherwise pass the other assertions)

`go build ./...` and `go test ./...` green across the context-loader server.

## Contract surfaces updated

MCP tool schema, enforced en-US schema descriptions, REST contract (`docs/api/context-loader/object-instance.yaml`), and the generated sandbox stub (`make -C infra/sandbox bkn-tools`).


sh00tg0a1:
Blocker confirmed and fixed in 42aa3532.

I reproduced it on the VM before touching the code, with `concept_groups=["worldcup_competition"]` + `exclude_object_types=["stadiums"]`: the excluded object type still came back with instances, because `completeReferencedObjectTypes` refetched it as a relation endpoint. After the fix the same call returns nothing, and the unit test that covers it fails against the previous logic with `[obj_0 audit_log]`.

Two notes on the fix.

**The ordering alone was not enough, and the obvious ordering is wrong in the other direction.** My first attempt scoped before completion so no wasted fetch would happen. A test I added for the opposite case then failed: when the caller pins an object type that sits outside the concept groups and is reachable only as a relation endpoint, filtering before completion means it never gets fetched and is reported as non-existent. So the scope now runs after completion — still before ranking and the TopK cut, which is the ordering that actually matters.

**On the non-blocking question about relations and actions:** not intentional, now fixed. `applyToConcepts` drops relations whose endpoint and actions whose object type the scope removed. Leaving them was wrong on two counts — `kn_search` callers would read dangling references (and `convertActionTypesToLocal` would hand them an empty `ObjectTypeName`), and `selectObjectTypesForConceptRetrieval` refills relation endpoints without a limit, so a surviving relation would walk its out-of-scope endpoint straight back into the pool. A relation is kept only when both endpoints are in scope; an empty endpoint id is not treated as out of scope.

Verified on the VM (10.211.55.4) with the rebuilt image: group path + exclude now returns empty, allow list narrows correctly, exclusion wins on overlap, and a name passed as an id still gets the explanatory message.

PR #1067, opened 2026-08-19
fix(trace): align business provenance summary filters

leecky:
## What Changed

- [x] Make conversation summary filtering honor the same keyword and identity fields as other business provenance lists.
- [x] Add a regression test for keyword matching against conversation question/result previews.
- [ ] Docs/doc 1 — not needed; behavior-only repair.

---

## Why

- Related Issue: N/A — user-reported business provenance filter defect.
- Related Feature: BKN Trace business provenance.
- Background: the conversation list ignored keyword filtering, while interaction-level search could return matching records.

---

## How

- Replaced the conversation-only status/evidence predicate with a complete aggregate predicate.
- Keyword search is case-insensitive and covers conversation ID, question/result previews, agent identities, business domain, and error summary.
- Breaking changes: none. The existing API query contract is preserved.

---

## Testing

- [x] Unit tests passed locally: `go test ./src/domain/service/evidencesvc -count=1`
- [ ] Integration tests passed locally — not applicable; this is a pure service predicate regression.
- [ ] Verified in test environment — pending review/deployment.

Test notes:

- Added a red/green regression proving a matching conversation remains visible for keyword `采购订单`.

---

## Risk & Rollback

- Potential risks: broader aggregate predicate now applies identity/domain/network/time checks consistently; query results may exclude rows previously admitted by an incomplete filter.
- Rollback plan: revert commit `4a4cd36d6`.

---

## Additional Notes

- Pairs with the Studio PR that removes the redundant business-domain control and sends canonical `active` for the user-facing running status.

### DIFF
--- a/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
+++ b/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
@@ -473,51 +473,15 @@ func filterConversationSummaries(
 ) []evidencevo.ConversationSummary {
 	filtered := make([]evidencevo.ConversationSummary, 0, len(entries))
 	for _, entry := range entries {
-		if !matchesConversationSummaryFilters(entry, options) {
+		if options.Status != "" && entry.Status != options.Status {
+			continue
+		}
+		if options.EvidenceCompleteness != "" && entry.EvidenceCompleteness != options.EvidenceCompleteness {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


leecky:
Review identified that this change made conversation keyword matching stricter than the existing request-level contract, breaking request ID, business-reference, and error matching. The follow-up commit restores the original behavior; this PR now has no effective change and is being closed rather than merged.

PR #1087, opened 2026-08-20
fix(trace): preserve initial conversation intent

leecky:
## Summary

- select the earliest interaction that contains a user question for the conversation preview
- keep question and result previews scoped to the same interaction, allowing an unavailable result to remain empty
- remove the conversation-wide fallback that could replace initial intent or combine artifacts from different interactions

## Test plan

- env GOCACHE=/tmp/codex-bkn-foundry-go-cache go test ./... -count=1 from bkn-trace/agent-observability
- git diff --check

Addresses openbkn-ai/openbkn-ee#32.

## Integration note

The enterprise build must consume a Foundry revision containing this change before the issue can be considered deployed.


leecky:
Fixed in c7108dd7a. The request-level fallback now applies only when the entire conversation has no `InteractionID`, preserving supported legacy/2.0 summaries. Any conversation containing at least one interaction ID still uses only interaction-scoped candidates, so the initial-intent fix cannot reintroduce cross-interaction question/result pairs.

Regression coverage added for a legacy conversation without interaction IDs.

PR #1107, opened 2026-08-21
fix(trace): require public lifecycle domains in chart

leecky:
## Summary
- reject an empty public lifecycle domain allowlist during Helm rendering
- add chart regression coverage for the default bd_public scope and empty-list rejection

## Test plan
- [x] make helm-lint
- [x] make lint test build


leecky:
Fixed blocking item 17c2ae07 in d3372c952: the default-domain assertion now checks the lifecycle env name and its immediately following value as one pair. A bd_other probe is rejected; make helm-lint and make lint test build pass.\n\nI am not adding an explicit disable switch: this issue requires a non-empty approved allowlist, and a switch would expand the deployment contract beyond the reported configuration defect.\n\n/review

PR #1127, opened 2026-08-23
fix(deploy): reconcile Trace persistence profile on same-version installs

### LINKED ISSUE #1126: fix(deploy): prevent Trace Core profile drift on same-version upgrades
## Description

`agent-observability` can retain Chart defaults (`core.store=memory`, disabled projection, in-memory evidence) when the installed Chart version equals the target. A Pod restart then clears the in-memory session/receipt state, making historical business-provenance conversations and Trace summaries disappear from Studio although OpenSearch data still exists.

## Steps to reproduce

1. Install or upgrade `agent-observability` with standalone Chart defaults.
2. Run a normal OpenBKN install with the same Chart version.
3. Observe that the installer skips the release.
4. Restart the `agent-observability` Pod.

## Expected behavior

The OpenBKN installer reconciles the durable Trace runtime profile: MariaDB Core, enabled projection, OpenSearch evidence, and the ingest Secret.

## Actual behavior

The same-version optimization preserves a volatile runtime profile.

## Environment

- Local Kubernetes / localhost
- Foundry `main` (2026-08-23)

- [x] A volatile runtime profile triggers reconciliation even when the Chart version is unchanged.
- [x] A durable profile still uses the version-skip optimization.
- [x] Explicit volatile `--set` values are honored.
- [x] Helm values read/parse failures do not cause an unverified rollout.
- [x] Regression tests cover local and repository install paths.
- [ ] CI green.

## Boundaries / Risk

No schema/data deletion. The resulting upgrade rolls only `agent-observability`; rollback is to reinstall the previous release values or revert the installer change.

## Additional context

Local evidence: historical OpenSearch alias `bkn-trace-core` still resolves to `bkn-trace-core-v014-r1` with 402 receipt documents; the running deployment uses `BKN_TRACE_CORE_STORE=memory` and projection disabled.

leecky:
## What Changed

- [x] Reconcile `agent-observability` when an installed same-version release still has a volatile Trace runtime profile.
- [x] Add regression coverage for volatile/durable profiles, explicit overrides, failed values reads, and local/repository install paths.
- [ ] Docs/doc 1 — not applicable; installer behavior is covered by inline contract comments and regression tests.

---

## Why

- Related Issue: Closes #1126
- Related Feature: N/A — deployment regression fix.
- Background: Chart defaults intentionally use in-memory stores for standalone use, while the OpenBKN installer supplies persistent overrides. The same-version skip could preserve a directly installed/default profile, so a Pod restart cleared visible sessions and business-provenance summaries.

---

## How

- Check the installed Helm values as JSON before taking the same-version skip for `agent-observability`.
- Reconcile only confirmed profile drift; unreadable values preserve the skip to avoid an unverified rollout.
- Honor the final explicit `--set` value, matching Helm precedence.
- Breaking changes (API, config, database, etc.): none.

---

## Testing

- [x] Unit tests passed locally (`openbkn_trace_profile_test.sh`: 82 checks).
- [x] Integration tests passed locally (`openbkn_trace_prerequisites_test.sh`, `mariadb_trace_database_test.sh`).
- [ ] Verified in test environment — requires merge and a controlled deployment.

Test notes:

- `bash -n deploy/scripts/services/openbkn.sh`
- `git diff --check`

---

## Risk & Rollback

- Potential risks: the next matching install upgrades only an `agent-observability` release confirmed to be volatile.
- Rollback plan: revert this commit; existing Helm values and persisted MariaDB/OpenSearch data are unchanged.

---

## Additional Notes

- Local diagnosis: `bkn-trace-core` still aliases 402 historical receipt documents; the running service was configured with `BKN_TRACE_CORE_STORE=memory` and disabled projection.
- The independent Collector `@timestamp` zero-value observation is deliberately out of scope for this deployment-profile PR.

### DIFF
--- a/deploy/scripts/services/openbkn.sh
+++ b/deploy/scripts/services/openbkn.sh
@@ -962,6 +962,98 @@ _openbkn_release_extra_sets() {
     fi
 }
 
+# Chart version equality alone is not sufficient for agent-observability: the
+# product installer supplies its durable Trace profile as Helm values, while
+# the standalone chart deliberately defaults to in-memory stores. Reconcile a
+# release that was installed directly (or before the profile existed) so a
+# later normal install cannot leave conversations volatile after a Pod restart.
+_openbkn_agent_observability_has_durable_profile() {
+    local namespace="$1"
+    local values
+    if ! values="$(helm get values agent-observability -n "${namespace}" --all -o json 2>/dev/null)"; then
+        return 2
+    fi
+    python3 -c '
+import json, sys
+try:
+    values = json.load(sys.stdin)
+except (json.JSONDecodeError, TypeError):
+    sys.exit(2)
+core = values.get("core", {})
+evidence = values.get("evidence", {})
+durable = (
+    core.get("store") == "mariadb"
+    and core.get("projection", {}).get("enabled") is True
+    and evidence.get("store") == "opensearch"
+    and bool(evidence.get("ingestAuth", {}).get("existingSecret"))
+)
+sys.exit(0 if durable else 1)
+' <<<"${values}"
+}
+
+_openbkn_agent_observability_profile_is_explicitly_volatile() {
+    local key value
+    local -a set_values=("${CORE_SET_VALUES[@]-}")
+    for key in core.store core.projection.enabled evidence.store evidence.ingestAuth.existingSecret; do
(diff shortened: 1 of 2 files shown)


leecky:
已请求 CODEOWNER @sh00tg0a1 审查。补丁避免相同 Chart 版本跳过时保留易失 Trace profile；历史数据不做删除或迁移。

PR #1129, opened 2026-08-23
fix(bkn-trace): repair Collector trace timestamps

### LINKED ISSUE #1128: fix(bkn-trace): repair zero @timestamp on Collector-exported spans
## Description
The OpenTelemetry Collector OpenSearch exporter (0.148.0) persists SS4O trace spans with a valid `startTime`, but serializes `@timestamp` as Go zero time (`0001-01-01T00:00:00Z`).

## Steps to reproduce
1. Deploy the bundled `otelcol-contrib` chart with `mapping.mode=ss4o`.
2. Send an OTLP span.
3. Query `ss4o_traces-default-namespace` and inspect `@timestamp`.

## Expected behavior
Each span has `@timestamp` equal to its OTLP start time, without changing logs or overwriting valid timestamps.

## Actual behavior
`startTime` is correct but `@timestamp` is zero, so any generic time-range query on `@timestamp` misses current traces.

## Environment
- OpenBKN localhost / Kubernetes
- Collector image: `otel/opentelemetry-collector-contrib:0.148.0`

- [ ] Future trace spans are routed through an idempotent OpenSearch ingest pipeline that repairs only a missing/zero `@timestamp` from `startTime`.
- [ ] Logs and spans with a valid timestamp are unchanged.
- [ ] The product installer wires the Collector to the pipeline and creates it before Collector rollout.
- [ ] Regression tests cover the chart render and installer contract.
- [ ] CI green.

## Boundaries / Risk
Do not mutate historical trace documents automatically. Deploying the ingest pipeline changes cluster configuration and requires the repository confirmation gate.

## Additional context
The upstream exporter encodes `startTime` but does not populate the SS4O `@timestamp` field. Existing trace graph reads already sort on `startTime`; this fix makes new documents compatible with generic OpenSearch time queries.

leecky:
## What Changed

- [x] Create the `bkn-trace-span-timestamp-v1` OpenSearch ingest pipeline through the existing Trace OpenSearch client.
- [x] Wire the pipeline into the full-product Trace and Collector install profiles.
- [x] Reconcile same-version Collector releases when the pipeline setting is absent.
- [x] Add focused Go and Helm/installer regression coverage.

---

## Why

- Related Issue: Closes #1128
- Background: the upstream Collector OpenSearch SS4O trace encoder writes `startTime` but serializes `@timestamp` as `0001-01-01T00:00:00Z`. Generic time-range queries consequently omit current traces.

---

## How

- The pipeline copies `startTime` only when a document has `traceId` and its `@timestamp` is missing or zero; logs and valid timestamps are left untouched.
- Agent Observability creates the pipeline before becoming ready; the collector uses it for subsequent writes.
- Existing Collector releases are upgraded even at the same chart version if the pipeline value is absent; unreadable Helm values retain the conservative skip.
- Breaking changes: no API, schema, or historical-data changes.

---

## Testing

- [x] Integration tests passed locally (not required; no external dependency added)
- [ ] Verified in test environment (requires deploy confirmation)

Test notes:

- `go test ./... -count=1` (agent-observability)
- `go vet ./...` (agent-observability)
- `bash deploy/scripts/services/openbkn_trace_profile_test.sh` (87 checks)
- `bash deploy/scripts/services/openbkn_trace_prerequisites_test.sh` (33 checks)
- Collector config-rollout, secure OpenSearch contract, and both Helm lints

---

## Risk & Rollback

- Potential risks: secure OpenSearch credentials need ingest-pipeline management permission; failure intentionally keeps Trace service unready rather than allowing invalid timestamps.
- Rollback plan: revert this PR and rerun the installer; the named ingest pipeline is isolated and does not alter historical documents.

---

## Additional Notes

- Historical documents are not automatically rewritten; trace graph reads continue to use `startTime`.
- Deployment changes OpenSearch cluster configuration and will follow the repository confirmation gate.

### DIFF
--- a/bkn-trace/agent-observability/charts/agent-observability/templates/deployment.yaml
+++ b/bkn-trace/agent-observability/charts/agent-observability/templates/deployment.yaml
@@ -50,6 +50,8 @@ spec:
               value: {{ .Values.opensearch.traceIndex | quote }}
             - name: OPENSEARCH_LOG_INDEX
               value: {{ .Values.opensearch.logIndex | quote }}
+            - name: OPENSEARCH_TRACE_TIMESTAMP_PIPELINE
+              value: {{ .Values.opensearch.traceTimestampPipeline | quote }}
             - name: OPENSEARCH_EVIDENCE_INDEX
               value: {{ .Values.evidence.index | quote }}
             - name: BKN_TRACE_EVIDENCE_STORE
--- a/bkn-trace/agent-observability/charts/agent-observability/values.yaml
+++ b/bkn-trace/agent-observability/charts/agent-observability/values.yaml
@@ -153,6 +153,9 @@ opensearch:
   # Keep these aligned with the bundled Collector's ss4o dataset/namespace.
   traceIndex: ss4o_traces-default-namespace
   logIndex: ss4o_logs-default-namespace
+  # Empty for standalone use. The product installer supplies the compatibility
+  # pipeline before it configures the Collector to require it.
+  traceTimestampPipeline: ""
   auth:
     enabled: false
     existingSecret: ""
--- a/bkn-trace/agent-observability/src/boot/bootstrap.go
+++ b/bkn-trace/agent-observability/src/boot/bootstrap.go
@@ -103,6 +103,11 @@ func NewApp() (*App, error) {
 		},
 		openSearchConfig.Timeout,
 	)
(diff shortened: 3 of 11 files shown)


leecky:
Pre-PR independent review completed. A same-version Collector rollout gap was found and fixed; the follow-up review found no remaining P1 issues. Local verification is recorded in the PR body.

PR #1133, opened 2026-08-23
fix(deploy): reconcile trace timestamp pipeline profile

leecky:
## What Changed

- [x] Include the Trace timestamp pipeline in the agent-observability same-version durability check.
- [x] Add regression coverage for an otherwise durable existing profile that lacks the pipeline.

---

## Why

- Follow-up to #1129. The prior same-version checks could skip agent-observability while upgrading Collector, leaving Collector configured to reference a pipeline that had never been created.

---

## How

- A profile is now durable only when its OpenSearch timestamp pipeline is also present.
- Explicit user pipeline overrides remain respected.

---

## Testing

- [x] `bash deploy/scripts/services/openbkn_trace_profile_test.sh` (88 checks)
- [x] `bash -n deploy/scripts/services/openbkn.sh`

---

## Risk & Rollback

- Risk: an existing AO deployment without the pipeline is rolled once to reconcile its Helm values.
- Rollback: revert this focused commit; no data is changed.

### DIFF
--- a/deploy/scripts/services/openbkn.sh
+++ b/deploy/scripts/services/openbkn.sh
@@ -985,11 +985,13 @@ except (json.JSONDecodeError, TypeError):
     sys.exit(2)
 core = values.get("core", {})
 evidence = values.get("evidence", {})
+opensearch = values.get("opensearch", {})
 durable = (
     core.get("store") == "mariadb"
     and core.get("projection", {}).get("enabled") is True
     and evidence.get("store") == "opensearch"
     and bool(evidence.get("ingestAuth", {}).get("existingSecret"))
+    and opensearch.get("traceTimestampPipeline") == "bkn-trace-span-timestamp-v1"
 )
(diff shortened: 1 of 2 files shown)


leecky:
This is the focused mainline follow-up for the same-version profile asymmetry discovered during the 0.1.4 backport review. It adds the missing pipeline criterion and regression coverage.

leecky:
已修复阻塞项 `5c4f0055`：删除了会跳过整个 durable 收敛的 pipeline 覆盖豁免。

现在 AO 的 durable 判据要求：四项持久化配置全部正确，且 `opensearch.traceTimestampPipeline` 为任意非空值；自定义 pipeline 不会掩盖 `core.store=memory` 等不正确配置。新增 `memory + custom pipeline` 必须调和的回归用例，`openbkn_trace_profile_test.sh` 已通过 89 checks。

/review

PR #1132, opened 2026-08-23
fix(bkn-safe): backport resource_manage implies view_detail to 0.1.4

sh00tg0a1:
Backport of #1123 to `release/0.1.4`. Refs #1121.

`release/0.1.4` carries #977, which converged the table's management verbs onto the catalog — so on this branch too, `catalog.resource_manage` is the normal way to hand out table management, and a grant of it alone reaches nothing: every management route loads its target first, and that load is a `view_detail` judgement.

Cherry-pick of the six commits on `fix/1121-resource-manage-implies-view-detail`, clean — no conflicts, no adaptation.

## What it does

- Declares the rule as data: `catalog.resource_manage` `implies` `view_detail`; the seed refuses to boot on an implication that names an undeclared operation, implies itself, or forms a cycle.
- Expands at grant time on all three write faces: admin object-grants, role grants, and the internal `POST /authz/policies`. Role revoke expands in reverse so dropping `view_detail` cannot leave the unusable verb behind.
- Repairs grants written before the rule existed, at seed time, add-only and idempotent.
- Records each repair in the audit trail, labelled by the surface that can actually show the grant (`object-grants` / `roles` / `policies`).

## Compatibility

Additive in code: new column via AutoMigrate, no change to the enforcement engine, socket signatures `ee` consumes unchanged.

**Sign-off needed at merge**, same as on the main PR: the back-fill is the one part that is not inert on upgrade. Accounts already granted `catalog.resource_manage` gain `view_detail` visibility the first time the new build boots. That is the product decision taking effect, as a one-off auditable data change.

## Testing

- `go build ./... && go test ./... -count=1` in `bkn-safe/server` on this branch — all green.
- Verified end to end on the dev VM against real MariaDB data (main branch build):
  - a grant carrying `catalog.resource_manage` alone reproduced the reported 403 naming `view_detail`;
  - after the upgrade the seed logged `rows=1` and repaired exactly that row — `view_detail` count went 17 to 18, nothing else moved;
  - the audit row landed as `object-grants` / `grant` with `_reason: implied by resource_manage`;
  - the same request then passed both permission gates;
  - a fresh grant of `resource_manage` alone through the internal policy route came back carrying `view_detail`.

## Pre-merge checklist

- [x] Unit tests pass
- [x] Verified against a real deployment
- [x] Cherry-pick applied cleanly
- [ ] Human review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1123, opened 2026-08-22
fix(bkn-safe): bind resource_manage to view_detail when a grant is written

sh00tg0a1:
Closes #1121

## Problem

A user holding only `catalog.resource_manage` cannot edit anything in that catalog. `PUT /resources/:id` loads the resource first (`updateResource` → `rs.GetByID`), and that load is a `view_detail` judgement: it refuses with 403 before `Update` ever reaches its `modify` check, which is the check that would have fallen back to `catalog.resource_manage` (#817).

The 403 names `view_detail`, so it reads as a bug in the check rather than as a hole in the grant. #977 converged the table's management verbs onto the catalog, which makes `resource_manage` the normal way to hand out table management — so this is now on the main path, not an edge case. Seeded roles do not hit it because `network_builder` gets `view_detail` and `resource_manage` together; hand-made object grants do.

Product decision on #1121: managing must rest on being able to see. `resource_manage` carries `view_detail` with it.

## Change

**Declared as data.** `model.Operation` gains `ImpliedOperationIDs`, `catalog.json` gives `catalog.resource_manage` `"implies": ["view_detail"]`, and the seed refuses to boot on an implication that names an operation the type does not declare, implies itself, or forms a cycle — the same treatment `parent_operation` already gets.

**Applied at write time, not at enforcement time.** The enforcement engine is untouched. A grant is stored complete, so the policy table says what the accessor holds and no reader has to replay a rule to know it.

- `POST /admin/object-grants` expands the requested set *after* validating it against the catalog, so a typo is still a 400. Replace semantics make it self-healing: a console that clears `view_detail` while leaving `resource_manage` ticked sends a set the expansion puts back.
- `GrantRolePermission` expands the same way.
- `RevokeRolePermission` expands in reverse: revoking `view_detail` takes `resource_manage` with it, because leaving the implying verb behind rebuilds exactly the grant that cannot be used.

The implication runs one way only — granting `view_detail` never drags `resource_manage` in behind it.

**Repaired at seed time.** `backfillImpliedOperations` adds the implied operations to every policy row that already holds an implying one, driven by the declarations in `catalog.json` rather than a hardcoded pair, transitively, add-only and idempotent. This is the same place a shipped vocabulary correction is migrated onto existing object grants (`renameLegacyOperations`).

[...]

PR #1134, opened 2026-08-23
fix(bkn-safe): decide a role revoke against the set, not one operation at a time

sh00tg0a1:
Follow-up to #1123. Refs #1121.

#1123 merged at `c6242a75`, one commit short of the fix below. The review round on the 0.1.4 backport (#1132) raised it as a blocker after that merge, so `main` currently carries the defect while the backport branch does not — this closes the gap and puts the two lines back on identical code.

## Problem

`RevokeRolePermission` judged each operation against the **live** state, so an operation earlier in the caller's list was still held when a later one was considered.

`DELETE /admin/roles/:id/permissions` with `operations: ["view_detail", "resource_manage"]` kept `view_detail` — `resource_manage` had not been removed yet, so `view_detail` took the "still implied, retain" branch — while `["resource_manage", "view_detail"]` dropped both. Same request, two answers, and the surviving grant is the silent one: the route answers 204 either way.

## Change

Decide against the **remainder** — what the role is left with once the whole request is applied — computed once, before the first removal, so no operation's verdict can depend on another's.

| `DELETE` body | before | after |
| --- | --- | --- |
| `[view_detail, resource_manage]` | keeps `view_detail` | drops both |
| `[resource_manage, view_detail]` | drops both | drops both |
| `[view_detail]`, `resource_manage` kept | keeps `view_detail` | keeps `view_detail` (the invariant) |

That needs the set, so the socket gains `RevokeRolePermissions` and the route makes one call instead of looping. The per-operation entry point stays, delegating to the new one and marked `Deprecated`: it is part of the contract `ee` consumes, so it is removed only after the consumers move over.

`adminwrite.Routes` lives in this repo and `ee` registers it, so the route change covers both builds.

## Testing

`go build ./... && go test ./... -count=1` in `bkn-safe/server` — all green.

New cases: both array orders drop both operations; revoking `view_detail` alone while `resource_manage` is kept retains it.

## Pre-merge checklist

- [x] Unit tests pass
- [x] Same content as the 0.1.4 branch (#1132)
- [ ] Human review

### DIFF
--- a/bkn-safe/server/extension/adminwrite/adminwrite.go
+++ b/bkn-safe/server/extension/adminwrite/adminwrite.go
@@ -103,25 +103,14 @@ type Services interface {
 	UpdateRole(ctx context.Context, id string, patch RolePatch) error
 	// DeleteRole deletes a custom role and purges its bindings and grants.
 	DeleteRole(ctx context.Context, id string) error
-	// RevokeRolePermissions revokes a set of operations from a custom role over
-	// one resource pattern, judged against the state the role will be left in.
-	//
-	// The set matters. An operation another operation implies cannot be dropped
-	// while that implying operation is kept (#1121), and deciding that one
-	// operation at a time makes the answer depend on the order the caller listed
-	// them: revoking [view_detail, resource_manage] would keep view_detail,
-	// while [resource_manage, view_detail] would drop both. Judged against the
-	// remainder, both orders drop both.
-	RevokeRolePermissions(ctx context.Context, roleID, resourceType, resourceID string, ops []string) error
-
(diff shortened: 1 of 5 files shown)


sh00tg0a1:
同步 #1132 复评的阻塞项修复(`e2c401d4`):`RevokeRolePermissions` 不能直接加进 `Services` 接口——那会让 ee 侧已有的 `Services` 实现(含 mounter 测试替身)在升级 core 依赖后编译不过。改为独立可选接口 `RoleSetRevoker` + 路由类型断言回落,照 `AnyPermissionRequirer` 的既有先例。

本 PR 与 #1132 的对应内容 patch-id 逐字节相同。

PR #1139, opened 2026-08-23
fix(bkn-trace): backport collector timestamp pipeline repair to 0.1.4

leecky:
## What Changed

- [x] Backported #1137's Collector runtime compatibility fix to `release/0.1.4`.
- [x] Uses the Trace index default pipeline rather than unsupported Collector exporter configuration.
- [x] Reconciles existing 0.1.4 Collector releases that retain the invalid value.
- [x] Keeps the deployed Collector image runtime-validation guard.
- [x] Retains the Collector chart README compatibility note.

---

## Why

- Related Issue: N/A — release backport of #1137.
- Related Feature: BKN Trace timestamp repair.
- Background: an upgraded Collector can CrashLoop when the previous exporter-level pipeline setting is rendered by the current runtime image.

---

## How

- Applies the same two reviewed commits from #1137 without conflict.
- The OpenSearch template applies the repair only to the configured Trace index.
- Breaking changes: none; this corrects an invalid runtime configuration.

---

## Testing

- [x] Unit tests passed locally: `go test ./...` in `bkn-trace/agent-observability`.
- [x] Integration-style validation passed locally: `openbkn_trace_profile_test.sh` (89 checks) and real Collector `validate`.
- [ ] Verified in test environment: pending release PR merge and localhost rollout.

Test notes: `helm lint` and configuration checksum validation also passed.

---

## Risk & Rollback

- Potential risks: OpenSearch must be reachable when agent-observability bootstraps the index template; otherwise startup reports the existing bootstrap error.
- Rollback plan: revert this backport PR. For an already-running environment, preserve the Collector chart without `opensearchExporter.pipeline` and remove the Trace index default-pipeline setting only if rollback is required.

---

## Additional Notes

- Source PR: #1137.
- This PR must merge before the planned localhost community and enterprise E2E rollout.

PR #1140, opened 2026-08-23
fix(bkn-trace): reconcile index-level timestamp repair

leecky:
## What Changed

- [x] Added an installer-managed timestamp-repair revision to reconcile same-version deployments after an implementation change.
- [x] Creates the exact Trace index with its default ingest pipeline on fresh installs instead of installing a mappings-displacing index template.
- [x] Added regressions for frozen release profiles and fresh Trace index initialization.
- [x] Documented required OpenSearch permissions for restricted accounts.

---

## Why

- Related Issue: N/A — hardening follow-up to #1137 and release backport #1139.
- Related Feature: BKN Trace timestamp repair.
- Background: release chart versions are frozen, so profile completeness needed an explicit revision; a settings-only high-priority template could exclude generic SS4O mappings.

---

## How

- A durable AO profile now requires `traceTimestampPipelineRevision=index-default-pipeline-v1`.
- Existing Trace indexes receive an index-setting update; absent Trace indexes are created with that setting.
- Breaking changes: none.

---

## Testing

- [x] Unit tests passed locally: `go test ./...` in `bkn-trace/agent-observability`.
- [x] Integration-style validation passed locally: `openbkn_trace_profile_test.sh` (91 checks) and real Collector `validate`.
- [ ] Verified in test environment: pending merge and localhost rollout.

Test notes: `helm lint` and diff checks passed.

---

## Risk & Rollback

- Potential risks: a restricted OpenSearch account needs ingest-pipeline, index-create, and index-settings permissions for the configured Trace index.
- Rollback plan: revert this PR; the existing Trace index setting can be removed manually only if a rollback is required.

---

## Additional Notes

- Related release backport PR: #1139.
- OpenSearch index-template precedence is avoided rather than overridden.

### DIFF
--- a/bkn-trace/agent-observability/charts/agent-observability/templates/deployment.yaml
+++ b/bkn-trace/agent-observability/charts/agent-observability/templates/deployment.yaml
@@ -22,10 +22,11 @@ spec:
         {{- with .Values.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.podAnnotations }}
       annotations:
+        openbkn.ai/trace-timestamp-pipeline-revision: {{ .Values.opensearch.traceTimestampPipelineRevision | quote }}
+        {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- end }}
     spec:
       serviceAccountName: {{ include "agent-observability.serviceAccountName" . }}
       {{- with .Values.imagePullSecrets }}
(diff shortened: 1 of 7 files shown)


leecky:
Fixed in 4139b8737:
- revision now changes the AO Pod template and triggers rollout;
- concurrent index creation is accepted;
- legacy settings-only template is removed idempotently.

Validated: `go test ./...`, `openbkn_trace_profile_test.sh` (92 checks), `helm lint`, and actual Collector `validate`.

/review

leecky:
Follow-up in 5064f01c0:
- after a concurrent index creation, retry `_settings` exactly once so the default pipeline is bound during this startup;
- legacy-template deletion is now best-effort and cannot block AO startup.

No retry loop or new state was added. Full Go tests, Trace Profile checks, Chart lint, and real Collector validation pass.

/review

PR #1151, opened 2026-08-24
fix(bkn-safe): let a resource owner look up the colleague they are sharing with

sh00tg0a1:
Backport of #1150 to `release/0.1.4`. Cherry-pick of a37240d4, clean; `bkn-safe/server/internal/httpapi` suite green on this base.

## Problem

The owner-facing authorization drawer needs `GET /admin/departments` and `GET /admin/users/:id` to name a grantee. Both sit behind `RequireAdmin`, and an owner holds `authorize` on one object and no admin capability — so both answered 403 and the console rendered an error toast over a working form.

## Change

Those reads move to a second group at the same prefix guarded by `RequireAdminOrResourceOwner`. Paths are unchanged, so the console calls one URL whoever is looking.

Ownership is read from `ListObjectGrants` rather than the enforcer: a `Check` answer folds in type-wide role grants, and the seeded `network_builder` role carries `authorize` on five types — honouring those would open the directory to the whole role rather than to people who actually created something. A lookup error refuses rather than passes.

An owner receives `id`, `account`, `name`, capped at 50 rows; the administrator's shape (email, telephone, roles, departments) is unchanged.

## Why this belongs on the release line

`release/0.1.4` already carries the owner-delegation half (#1135) and bkn-studio `release/0.1.4` carries the drawer (#509). Without this, that feature 403s on every deployment built from this line — which is what happens on 118.196.95.132 today.

### DIFF
--- a/bkn-safe/server/internal/directory/departments_admin_list.go
+++ b/bkn-safe/server/internal/directory/departments_admin_list.go
@@ -149,9 +149,9 @@ func (s *Service) ListAllDepartments(ctx context.Context, search string, offset,
 	if err := q.Count(&total).Error; err != nil {
 		return nil, 0, err
 	}
-	// Same reason as the user list: the allocation is sized by a constant the
-	// scanner can see, not by the request parameter the clamp happens to bound.
-	deps := make([]model.Department, 0, min(limit, maxDepartmentPageSize))
+	// Same reason as the user list: constant capacity, no request parameter in
+	// the allocation at all.
+	deps := make([]model.Department, 0, defaultDepartmentPageSize)
 	if err := q.Order("parent_id, name").Offset(offset).Limit(limit).Find(&deps).Error; err != nil {
 		return nil, 0, err
 	}
--- a/bkn-safe/server/internal/directory/departments_admin_list.go
+++ b/bkn-safe/server/internal/directory/departments_admin_list.go
@@ -10,6 +10,11 @@ import (
 	"github.com/openbkn-ai/bkn-foundry/bkn-safe/server/internal/model"
 )
 
+const (
+	defaultDepartmentPageSize = 100
+	maxDepartmentPageSize     = 1000
+)
+
 // DepartmentListItem is the admin list view of a department (snake_case JSON).
 type DepartmentListItem struct {
 	ID                 string `json:"id"`
@@ -128,10 +133,10 @@ func (s *Service) subtreeMemberCounts(ctx context.Context, deptIDs []string) (ma
 // ListAllDepartments returns a flat, paginated department list with member_count.
 func (s *Service) ListAllDepartments(ctx context.Context, search string, offset, limit int) ([]DepartmentListItem, int64, error) {
 	if limit <= 0 {
-		limit = 100
+		limit = defaultDepartmentPageSize
 	}
-	if limit > 1000 {
-		limit = 1000
+	if limit > maxDepartmentPageSize {
+		limit = maxDepartmentPageSize
 	}
 	if offset < 0 {
 		offset = 0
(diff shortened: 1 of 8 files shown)


sh00tg0a1:
Both blocking findings were real; fixed rather than dismissed.

**`8b632c8f` — the cap bounded one response, not the table.** `offset` is now pinned to 0 for an owner (a picker opens on the first page and narrows by typing; it never pages), `?role_id=` is dropped for them — "which accounts hold this privileged role" is a target list, not a lookup — and `total` reports the rows actually returned instead of the platform head count. The administrator's paging and totals are untouched.

**`a3c243eb` — departments had neither projection nor cap.** Correct, and it was the worse of the two: the owner branch returned the full admin row, including `manager_id`/`manager_name`/`code`/`email`/`remark` and the member counts — exactly the contact data the projection on the user list exists to withhold. Owners now get `id`/`name`/`parent_id`, capped the same way, on both the `parent_id` branch and the flat one.

Tests added for each: offset does not move the owner's window, `total` never exceeds the page, `role_id` does not narrow the result, department rows carry none of the withheld fields and stay under the cap — each paired with an assertion that the administrator's shape and page size did not change.

**On the non-blocking item** (stale grants after an object is deleted): vega does clean up — `catalog_service.go:1331-1335` and `connector_type_service.go:398` call `DeleteResources`, which reaches `RemoveResourcePolicies`. I could not verify the `knowledge_network` path: bkn-backend is not in this repository, so nothing here proves it clears the creator's `authorize` row on delete. Flagging it rather than claiming it is handled — if it does not, the consequence predates this PR (the owner keeps object-grant write access too, which is the larger half), but this PR does widen what a stale row is worth. Worth its own issue.

Full `bkn-safe/server` suite green.

sh00tg0a1:
`a65707c5` is right, and it is the same mistake as the last round one axis over: pinning `offset` bounded paging and left partitioning open. Walk the department tree this endpoint's sibling hands out, ask `?department_id=` for each node in turn, and the roster reassembles itself fifty rows at a time.

`?department_id=` is now dropped for an owner and `?include_subtree=` goes with it. The administrator keeps both. The owner's window is narrowed by typing, not by slicing — `search` stays, which is the same exposure `/me/grantable-users` already accepts by design.

Nothing in the console needs the filter on this path: the grantee picker sends `search` and `limit` only. If a department-browsing UI is built later it should come with its own contract rather than inherit the administrator's filters.

**Non-blocking item** (silent truncation of a branch): taken. The owner's department page now carries `truncated` when the cap cut it, on both the `parent_id` branch and the flat one. A tree that is silently short reads as "this is all of it", and the payload gave the client no way to tell.

Tests added: `department_id` does not narrow an owner's result while it still narrows the administrator's; the `parent_id` branch flags truncation when it bites.

Full `bkn-safe/server` suite green. The release-line branch carries the identical file contents — verified file by file, not just by review.

sh00tg0a1:
`5e8e02b3` confirmed — the flag I added in the previous round could not fire on the flat branch. The page cap is applied by the query, so the rows reaching the projection are already at most one page and the in-memory cut never triggers; a tenant with 500 departments got 50 rows and `truncated: false`, which is a partial org chart presented as the whole one.

It now compares against the unpaged count from `ListAllDepartments`. The flag stays a boolean deliberately: the owner is told that more exists, not how much, and with `offset` pinned there is no next page to ask for anyway. The `parent_id` branch keeps its own comparison, where `deps` really is the full child set and the in-memory cut is the thing that bites.

Test extended: the flat-branch case now asserts `truncated` is set when the cap bites, alongside the existing row-count and projection assertions.

**CodeQL** — the three `go/uncontrolled-allocation-size` alerts are open on `main` and predate this branch; they surface here because this PR moves the callers. `min(limit, ceiling)` did not satisfy the query (the allocation still reads a value that arrived in the query string), so the capacity is now a plain constant and a larger page grows by append. Behaviour is unchanged either way.

Both branches carry identical file contents — verified file by file.

PR #1156, opened 2026-08-25
fix(context-loader): support MCP hosts that flatten lifecycle schemas

leecky:
## What Changed

- Flatten the public `bkn_start_interaction` and `bkn_finish_interaction` input schemas: retain properties, required fields, enums, length checks, and `additionalProperties: false`; remove composition keywords.
- Enforce lifecycle field combinations in server validation: `continue` requires `conversation_id`, `new` omits it, and `completed` requires a non-empty `answer`.
- Return actionable `invalid_params` guidance with a correct call example for each cross-field error.
- Add regression coverage that the published wire schemas contain no composition keywords and that invalid calls never reach Trace Core.

## Why

Some MCP hosts destructively flatten root `oneOf` before their local tool-argument validation. They then reject a valid `new` start call because `question` and `agent_name` are incorrectly treated as additional properties. The request never reaches ContextLoader. This keeps the wire schema portable while preserving the lifecycle contract server-side.

## How

No public lifecycle behavior changes. Cross-field rules previously expressed using JSON Schema `oneOf` are enforced explicitly after generic schema validation.

## Testing

- [x] `make test`
- [x] `go test ./server/driveradapters/mcp -count=1`
- [x] `go vet ./server/driveradapters/mcp`
- [x] `git diff --check`

## Risk & Rollback

- Risk: an integration relying on introspecting `oneOf` will no longer see those conditional rules in the wire schema; the tool descriptions and recoverable validation errors remain the contract.
- Rollback: revert commit `f18faa553`.

## Additional Notes

This PR targets `release/0.1.4` and is intentionally not merged by this change.

### DIFF
--- a/adp/context-loader/agent-retrieval/server/driveradapters/mcp/lifecycle_adapter_test.go
+++ b/adp/context-loader/agent-retrieval/server/driveradapters/mcp/lifecycle_adapter_test.go
@@ -160,6 +160,11 @@ func TestLifecycleToolsRejectInvalidArgumentsBeforeCallingCore(t *testing.T) {
 			wantMessage: "bkn_start_interaction expects top-level agent_name, question, and conversation_mode; use continue with conversation_id or new without it",
 		},
 		{
+			name: "continue with empty conversation id", toolName: "bkn_start_interaction",
+			args:        map[string]any{"question": "查询库存", "agent_name": "供应链分析助手", "conversation_mode": "continue", "conversation_id": ""},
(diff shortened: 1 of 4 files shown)


leecky:
Fixed blocker 17383492 in f129a3b8d.

Generic schema failures now use field-level guidance; lifecycle combination guidance is emitted only after schema validation succeeds. Added regressions for `continue` plus an empty `question`, and `completed` plus an empty `interaction_id`.

/review

leecky:
Addressed the non-blocking guidance gap in 20952d746.

Empty `conversation_id` for `continue` and empty `answer` for `completed` now return their exact non-empty examples; unrelated schema failures retain generic field guidance. Added both regressions.

/review

PR #1171, opened 2026-08-27
fix(context-loader): keep the tool's own shape when attaching a trace receipt

sh00tg0a1:
## Problem

Managed MCP calls returned structured content the tool's own output schema does not describe. `search_instance` failed outright on hosts that validate it:

```
MCP error -32602: structured content does not match the tool output schema: data must have required property 'nodes'
```

Reproduced live against a managed deployment — the payload comes back nested:

```json
{"bkn_receipt":{...},"result":{"nodes":[],"message":"未检索到符合条件的实例数据。"}}
```

## Root cause

`session_guard.go` attached the durable receipt by type-asserting structured content to `map[string]any`. That assertion never held: business handlers build their result with `BuildMCPToolResult`, which passes the response **struct**. So every managed call took the fallback branch and nested the whole payload under `result`.

This broke the published contract for **every** tool, not just one. `search_instance` is the only schema of the 23 that declares `"required": ["nodes"]`, so it is the only one a validating host rejects; the rest silently handed agents a shape their schema says does not exist. The existing tests all used map-shaped fake results, so the struct branch had no coverage.

## Fix

- Marshal the payload into a map and inject `bkn_receipt` in place, keeping the tool's fields at the top level. The decode uses the precision decoder so wide integers stay `json.Number` — this round-trip is the last place the precision the driven adapters preserve could be rounded through float64.
- Payloads that are not JSON objects (text-only or errored results) have no shape to protect and keep the nested envelope rather than losing the receipt.
- A replayed terminal operation had the same defect for a different reason: the tool was not re-executed, so only the receipt exists and no payload in the declared shape can be produced. It now returns a `receipt_terminal` lifecycle error — hosts skip output validation on error results — instead of posing as a successful result carrying fields no schema describes.

No output schema needed changing: no business schema is a closed object, so a top-level `bkn_receipt` is valid for all of them. Lifecycle tools (`bkn_start_interaction` / `bkn_finish_interaction`) bypass the guard and were never affected.

## Tests

- New: struct payload keeps `nodes` at the top level, carries `bkn_receipt` alongside it, and preserves a 64-bit integer through the attachment.
- Updated: both terminal-replay tests now assert the lifecycle error envelope and that no schema-violating structured content is published.
- `go test ./...` green for agent-retrieval.

### DIFF
--- a/adp/context-loader/agent-retrieval/server/driveradapters/mcp/session_guard.go
+++ b/adp/context-loader/agent-retrieval/server/driveradapters/mcp/session_guard.go
@@ -141,10 +141,7 @@ func guardBusinessToolCallWithCompletion(
 			status = receiptStatus(ensured.Receipt)
 		}
 		if status == "completed" || status == "failed" {
-			replay := map[string]any{"operation": ensured.Operation, "receipt": ensured.Receipt}
-			result := mcpsdk.NewToolResultStructured(replay, "operation is terminal; durable receipt reused")
-			result.IsError = status == "failed"
-			return result, nil
+			return receiptTerminalToolError(status, ensured.Operation, ensured.Receipt), nil
 		}
 		if ensured != nil && !ensured.Execute && status == "pending" {
 			return receiptPendingToolError(ensured.Receipt), nil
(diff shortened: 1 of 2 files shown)


sh00tg0a1:
On the serialization note: keeping it as is.

The extra marshal + unmarshal is real, but it is not where this path spends its time. Receipts from the live verification on the test server: `search_instance` took 1.37s (issued 08:46:20.590 → terminal 08:46:21.963), `list_knowledge_networks` 30ms. A round-trip over a 100KB payload is sub-millisecond — roughly 0.1% of the large call, and lost inside the two lifecycle HTTP round-trips to Core on the small one.

Reusing the map would mean lifting it out of `rest.marshalTOON` and handing it to the MCP guard (TOON is the default on the MCP surface; the JSON branch is the rare one). That couples the shared REST serialization path to a guard concern — the same kind of coupling this PR is fixing — to save that sub-millisecond.

If this path is ever worth optimizing, the bigger cost is that every call serializes the same payload twice, once for text content and once for structured content; the fix there is to derive one from the other, which is a separate change and does not belong in this PR.

PR #1181, opened 2026-08-28
fix(sandbox): render BKN_SANDBOX_MCP_URL into the control plane

sh00tg0a1:
## What this fixes

The sandbox control plane already has a `bkn_sandbox_mcp_url` setting, and both schedulers write it into every sandbox pod:

```python
# k8s_scheduler.py
bkn_mcp_url = get_settings().bkn_sandbox_mcp_url.strip()
if bkn_mcp_url:
    env_vars.append(V1EnvVar(name="BKN_SANDBOX_MCP_URL", value=bkn_mcp_url))
```

Nothing ever set it. The chart's ConfigMap never rendered the key, and the Deployment references config by explicit per-key `configMapKeyRef` rather than `envFrom` — so adding it to the ConfigMap alone would not have been enough either.

On a default install the setting therefore stays empty, `bkn.available()` is always false, and every function that calls BKN reports the same thing regardless of how the caller authenticated:

```
RuntimeError: BKN not configured: missing MCP url or caller token.
```

Confirmed on the VM — the sandbox pod's environment:

```
CONTROL_PLANE_URL / DISABLE_BWRAP / S3_BUCKET / SESSION_ID  → present
BKN_SANDBOX_MCP_URL                                          → absent
```

## Defaults

| Chart | Default | Why |
|---|---|---|
| `sandbox` | `http://agent-retrieval:30779/api/agent-retrieval/v1/mcp/` | Ships alongside Context Loader in the same namespace, so it can carry a working value. Matches the address `bkn-agent`'s chart already uses. |
| `sandbox_standalone` | `""` | May be deployed with no BKN at all; the operator fills it in. |

Either can be overridden, and setting it back to `""` ships without the capability surface — functions then have to pass `mcp` in the event themselves, which is what the SDK already falls back to.

The trailing slash is required and the comment in `values.yaml` says why: without it the gateway answers 307 and the sandbox's `urllib` does not follow redirects on POST, so the symptom is a 400 with no body.

## Upgrade safety

The `configMapKeyRef` is `optional: true`. An upgrade may reuse a ConfigMap rendered before this key existed (`configMap.create=false`, or a hand-managed one); a required ref would leave the control plane unable to start. Optional degrades to the current behaviour instead.

Verified:

```
$ helm template t sandbox --set configMap.create=false | grep -A5 'name: BKN_SANDBOX_MCP_URL'
            - name: BKN_SANDBOX_MCP_URL
              valueFrom:
                configMapKeyRef:
                  name: sandbox-config
                  key: BKN_SANDBOX_MCP_URL
                  optional: true
```

No ConfigMap is rendered in that case, and the control plane still starts with the variable unset.

## Rendering checks

| Case | Result |
|---|---|
| `sandbox` default | `BKN_SANDBOX_MCP_URL: "http://agent-retrieval:30779/api/agent-retrieval/v1/mcp/"` |
| `sandbox_standalone` default | `BKN_SANDBOX_MCP_URL: ""` |
| `--set …BKN_SANDBOX_MCP_URL=http://cl.other/mcp/` | `"http://cl.other/mcp/"` |
| `--set …BKN_SANDBOX_MCP_URL=""` | `""` |
| `--set configMap.create=false` | Deployment renders, ref stays optional |
| `helm lint` both charts | `0 chart(s) failed` |

Additive only: 33 insertions, 0 deletions, no existing key or default touched.

## Scope

This closes the first of the three deployment gaps listed in #1177. The other two are separate:

- The sandbox template image still needs `bkn.py` + `_bkn_tools.py` (the VM's image carries only `__init__.py`, so `import sandbox_sdk.bkn` raises `ModuleNotFoundError`).
- `agent-operator-integration` needs a rebuild past #955 for the `bkn_token` field to be honoured at all.

Both charts get the same plumbing so they do not drift; `docker-compose` is left alone, since a local stack has no Context Loader to point at.

Related: #1177, #955

### DIFF
--- a/infra/sandbox/deploy/helm/sandbox/README.md
+++ b/infra/sandbox/deploy/helm/sandbox/README.md
@@ -77,6 +77,7 @@ The deprecated `image.defaultTemplate` value is still accepted as a compatibilit
 | `image.defaultTemplates.multiLanguage.repository` | Multi-language template image repository | `dip/sandbox-template-multi-language` |
 | `image.defaultTemplates.multiLanguage.tag` | Multi-language template image tag; empty uses `/app/VERSION` | `""` |
 | `controlPlane.env.ENVIRONMENT` | Control Plane environment | `staging` |
+| `controlPlane.env.BKN_SANDBOX_MCP_URL` | In-cluster Context Loader MCP endpoint written into every sandbox pod, so `sandbox_sdk.bkn` can call back into BKN. Trailing slash required. Empty ships without the capability surface. | `http://agent-retrieval:30779/api/agent-retrieval/v1/mcp/` |
 | `depServices.rds` | Core-provided database service configuration | enabled by values |
 
 ## Rendering
--- a/infra/sandbox/deploy/helm/sandbox/templates/configmap.yaml
+++ b/infra/sandbox/deploy/helm/sandbox/templates/configmap.yaml
@@ -33,6 +33,7 @@ data:
   CLEANUP_INTERVAL_SECONDS: {{ .Values.controlPlane.env.CLEANUP_INTERVAL_SECONDS | quote }}
   DISABLE_BWRAP: {{ .Values.controlPlane.env.DISABLE_BWRAP | quote }}
   PYTHONUNBUFFERED: {{ .Values.controlPlane.env.PYTHONUNBUFFERED | quote }}
+  BKN_SANDBOX_MCP_URL: {{ .Values.controlPlane.env.BKN_SANDBOX_MCP_URL | default "" | quote }}
   DEFAULT_TEMPLATE_IMAGE_REGISTRY: {{ $templateImageRegistry | quote }}
   DEFAULT_TEMPLATE_IMAGE_REPOSITORY: {{ $pythonBasicTemplate.repository | quote }}
   DEFAULT_MULTI_LANGUAGE_TEMPLATE_IMAGE_REPOSITORY: {{ $multiLanguageTemplate.repository | quote }}
--- a/infra/sandbox/deploy/helm/sandbox/templates/control-plane-deployment.yaml
+++ b/infra/sandbox/deploy/helm/sandbox/templates/control-plane-deployment.yaml
@@ -245,6 +245,12 @@ spec:
                 configMapKeyRef:
                   name: {{ .Values.configMap.name }}
                   key: PYTHONUNBUFFERED
+            - name: BKN_SANDBOX_MCP_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ .Values.configMap.name }}
+                  key: BKN_SANDBOX_MCP_URL
+                  optional: true
             - name: DEFAULT_TEMPLATE_IMAGE_REGISTRY
               valueFrom:
                 configMapKeyRef:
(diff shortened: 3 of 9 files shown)


sh00tg0a1:
第一条属实，已在 `b313b47b` 补上；第二条确认存在，做不到联动，改为把耦合写进注释。

## 1. K8s 原生 YAML 路径 — 已补

`deploy/manifests/` 确实是 README 里并列的第三种部署方式，`05-control-plane-deployment.yaml` 同样按 key 逐条引用 ConfigMap，上一版漏了。

不过它部署的是 `sandbox-system` 命名空间下的独立沙箱，旁边没有 BKN，所以取值跟 `sandbox_standalone` 一致——**键渲染出来但留空**：知道有这个旋钮，但不填就什么都不变。`configMapKeyRef` 同样带 `optional: true`。

解析验证：

```
configmap value: ''
deployment ref: {'name': 'sandbox-config', 'key': 'BKN_SANDBOX_MCP_URL', 'optional': True}
```

## 2. 默认值写死 agent-retrieval 地址 — 确认，但没有联动手段

查过了，Helm 读不到另一个 release 的 values，`deploy/scripts/bkn-foundry.template.yaml` 也只列 chart 名、没有 global/shared values 机制。同一个地址在仓里已经硬编码在多处，最贴近的先例是 `infra/bkn-agent/charts/values.yaml:49` 的 `contextLoaderMcpUrl`，写法完全相同。

所以这不是本次引入的耦合，但你说的失败形态是对的：地址错了不会 fail closed——`bkn.available()` 只检查「URL 和 token 都存在」，拨不动号它不知道，于是 `true` 照给，调用失败在网络层。

改不了机制，就把线索留在原地。`values.yaml` 的注释补了三句：Helm 为什么不能派生、权威端口在 agent-retrieval 的 `project.port`（它自己的 `sandboxMCPURL()` 渲染出同一个串给 run_code 用）、以及这个默认值不 fail closed 这件事本身。改端口的人至少能顺着注释找到另一处。

你说的「PTC 路径不受影响」也核过了：`ptc_server.go:187` 把 `event["mcp"]` 填成服务端自己算的地址，event 优先于环境变量（`bkn.py` 的 `_mcp_url()` 先读 `_RUNTIME` 再回退 env），所以 run_code 那条路确实完全不碰这个设置。只有 `@tool` 函数直调受影响。

## 关于「存量安装默认打开能力面」

这是本 PR 的目的本身：bundled chart 与 Context Loader 同批部署、同命名空间，打开之后调用是通的。失败形态从「未配置」变成网络层报错只发生在地址填错的情况下，而默认值就是 agent-retrieval 自己算出来的那一个。要退回旧行为，`--set controlPlane.env.BKN_SANDBOX_MCP_URL=""` 即可。

sh00tg0a1:
说得对，已在 `64b38278` 改掉。

本目录把沙箱固定在 `sandbox-system`，BKN 一般在别的命名空间，短名确实解析不到；同文件的 `DATABASE_URL` 用的就是 FQDN，示例跟它对齐：

```
http://agent-retrieval.<bkn-namespace>.svc.cluster.local:30779/api/agent-retrieval/v1/mcp/
```

同时把「地址错了不 fail closed」这句也写进了该文件的注释——照抄短名的人正是最需要看到这句的人。

PR #1197, opened 2026-08-29
fix: backport managed BKN functions for supply sample

leecky:
## Background\nOpenBKN 0.1.4 Function Toolbox calls could not let a sandboxed Function read Knowledge Network data under the managed caller scope. Third-party Agents therefore had to fall back to local calculation or failed with resource_not_disclosed.\n\n## Changes\n- forward server-captured authorization and managed Interaction IDs only for Function Toolbox calls;\n- expose published function Toolbox discovery and execution through Context Loader MCP;\n- add a parameter-only 0.1.4 Helm patch kit for agent-retrieval and agent-operator-integration, including rollback, readiness verification, and release digest record.\n\n## Safety\n- no DDL, data migration, sample import, or Knowledge Network mutation;\n- installation upgrades only the two named deployments with --reuse-values;\n- rollback requires the explicitly recorded original image tag.\n\n## Verification\n- go test -count=1 ./server/driveradapters/mcp ./server/drivenadapters (agent-retrieval)\n- go test -count=1 ./server/logics/toolbox ./server/driveradapters/toolbox (agent-operator-integration)\n- bash deploy/patches/0.1.4-function-bkn-context/test_patch_scripts.sh\n- Helm template image assertions for both charts\n\n## Release note\nAfter review, create immutable tag v0.1.4-supply-sample-p1 on the approved patch commit. Do not merge in a way that overwrites existing 0.1.4-release images.

PR #1198, opened 2026-08-29
fix: backport managed BKN functions for supply sample

leecky:
Supersedes #1197 (invalid branch name only).\n\n## Background\nOpenBKN 0.1.4 Function Toolbox calls could not let a sandboxed Function read Knowledge Network data under the managed caller scope. Third-party Agents therefore failed with resource_not_disclosed.\n\n## Changes\n- forward server-captured authorization and managed Interaction IDs only for Function Toolbox calls;\n- expose published Function Toolbox discovery and execution through Context Loader MCP;\n- include Sandbox Control Plane injection required by the 0.1.4 upstream baseline;\n- add a parameter-only Helm patch kit, rollback, readiness verification, and release digest record.\n\n## Safety\nNo DDL, data migration, sample import, or Knowledge Network mutation. The patch changes only agent-retrieval, agent-operator-integration, and sandbox-control-plane images.\n\n## Verification\n- Context Loader and Execution Factory targeted unit tests\n- patch-kit script test\n- Helm render assertions\n\n## Release\nAfter approval, publish immutable tag v0.1.4-supply-sample-p1 from this exact commit and populate the release record with its multi-architecture image digests.


sh00tg0a1:
Reviewed against `main` rather than only against the diff. The two blockers the automated review raised both reproduce; three further points below that a diff-scoped read cannot see.

## 1. This reverses an invariant that `main` states explicitly

`main` forbids exactly what this change does on the proxy path, and says why in the code:

`adp/execution-factory/operator-integration/server/driveradapters/common/proxy.go:172`
> The proxy path must not use this: it runs code registered by a third party, and its route authenticates by trusted header rather than by introspection, so the Authorization value there is an unverified passthrough. Injecting it would let a function author read and exfiltrate the invoking user's live credential.

`proxy.go:219`
> The credential is deliberately withheld — the code being run belongs to whoever registered the version, not to the caller.

Both landed after `release/0.1.4` branched, so this PR could not have seen them.

I merged this branch into a `main`-based branch locally to see what happens. `proxy.go` conflicts, and the conflict block puts the two positions side by side inside one function — one side saying the credential is deliberately withheld, the other assigning it. The signatures differ too (`main` takes `c *gin.Context`, this branch takes the three values). There is no resolution that keeps both.

The risk is not this PR. It is the forward-merge, where a `-X theirs` quietly drops the constraint and nobody notices. Worth deciding explicitly, in its own issue on `main`, whether that constraint survives — and saying so here either way.

## 2. A published function can now invoke published functions, including itself

`execute_published_tool` is registered on the main MCP server (`b.add(toolKeyExecutePublishedTool, ...)` in `mcp/app.go`), which is what the sandbox connects to:

```yaml
BKN_SANDBOX_MCP_URL: "http://agent-retrieval:30779/api/agent-retrieval/v1/mcp/"
```

`ptcUsableTools` (`mcp/ptc_toolkit.go:326`) filters only two sets, and this PR adds to neither:

```go
var ptcSkipTools = map[string]bool{"bkn_start_interaction": true, "bkn_finish_interaction": true}
var ptcSelfTools  = map[string]bool{toolKeyRunCode: true, toolKeyRunShell: true}
```

So it reaches the toolkit digest and the sandbox SDK exposes it. With the credential and `bkn_context` now forwarded, a function can call `execute_published_tool` with its own `tool_id`, and each level starts another sandbox.

`run_code` already guards the equivalent case — that is what `ptcSelfTools` is for. This opens an equivalent path without the equivalent guard.

What eventually stops it is poor: the operation cap trips only after N sandbox sessions are already running, and timeouts do not compose — the SDK default is 120s inside a typically shorter outer timeout. A single mistake, not malice, is enough to saturate the sandbox pool, and the blast radius is the cluster rather than one conversation.

Two ways out. Smallest: add the three tools to `ptcSkipTools`, so the sandbox does not see them. That closes it in one line, at the cost of function-calls-function — worth confirming whether that capability is in scope for this backport at all. Otherwise: bound the depth. `BusinessContext` already carries `ParentOperationID` and the lifecycle already validates the parent belongs to the same interaction, so the tree exists; walking it up gives the nesting depth with no new field and no contract change. Either way the inner timeout should be clamped to the outer remaining time.

## 3. `requestCredential` duplicates `GetToken`

`drivenadapters.GetToken(c)` (`drivenadapters/hydra.go:185-193`) already implements this precedence, and `main`'s `fillExecutionCredentialFromRequest` calls it. The comment is right that this mirrors existing behaviour — the `?token=` branch is pre-existing, not new risk — but two copies will drift. Reuse the existing one.

---

The catalog discovery and schema trimming read well, and the "partial interaction is treated as absent" guard is the right call. Points 1 and 2 are what I would want settled before this merges; point 3 is cleanup.

PR #1207, opened 2026-08-30
ci(review): register bkn-trace and data-migrator as code

sh00tg0a1:
## Problem

The review workflow's `paths` list decides which directories count as code. A directory missing from it is a whole service line that never gets reviewed — and it fails silently: nothing errors, the workflow simply does not run.

`bkn-trace` was missing. #1201 changed only that tree, so both `opened` and `synchronize` were filtered out; the author noticed because no review appeared, not because anything reported it. That PR changed session ownership and error disclosure semantics — the kind of change a second reader is worth the most on.

The tree has 128 non-test Go files and its own CI pipeline. It had simply never been listed here. `data-migrator` is in the same position with 67 Python files.

## Changes

Add `bkn-trace/**` and `data-migrator/**`.

The criterion is whether a directory holds code that runs, not how important it looks:

- `docs/` and `ref/` carry a script or two and stay out.
- `deploy/` keeps only its `scripts/` level; the rest is charts and manifests, which the file's own comment already excludes as configuration.

## Note

This file lives under `.github/`, which is not in the list either, so this PR does not review itself.

### DIFF
--- a/.github/workflows/automation-claude-review.yml
+++ b/.github/workflows/automation-claude-review.yml
@@ -34,6 +34,16 @@ name: Claude Code Review
 #
 # pull_request 触发只对代码路径生效；纯配置类改动（release manifest 刷新、镜像同步、
 # workflow 自身调整、文档）不进评审，既省额度也避免噪声。
+#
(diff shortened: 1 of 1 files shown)


sh00tg0a1:
@claude review pls

PR #1214, opened 2026-08-30
fix(trace): remove provenance projection tenant scope

leecky:
# Pull Request

## What Changed

- [x] Remove `tenant_id` from the historical-provenance build event.
- [x] Add immutable Core-owned MariaDB v021 to remove the corresponding projection-table column and obsolete index.
- [x] Keep the projection identity constrained to `interaction_id` and sealed `facts_hash`.

---

## Why

- Related Issue: [openbkn-ee#31](https://github.com/openbkn-ai/openbkn-ee/issues/31)
- Related Feature: [Historical provenance projection design](https://github.com/openbkn-ai/bkn-docs/pull/86)
- Background: business-domain has been removed; tenant has no approved architecture for historical-projection scope. Retaining it in a new historical read model would create an invented authorization boundary.

---

## How

- Key implementation points: omit tenant from the Core outbox value object and append v021 rather than changing released migrations.
- Breaking changes (API, config, database, etc.): MariaDB v021 drops only the unused historical-projection `tenant_id` column and its scope index. It does not alter Core lifecycle tables that still use tenant.

---

## Testing

- [ ] Integration tests passed locally (not applicable: no live MariaDB required for the unit suite)
- [ ] Verified in test environment (requires controlled migration deployment)

Test notes: `GOCACHE=/tmp/openbkn-go-build-cache GOMODCACHE=/tmp/openbkn-go-mod-cache go test ./... -count=1`

---

## Risk & Rollback

- Potential risks: deployments with v019/v020 apply one additive version that drops only projection-scope data already prohibited by the final model.
- Rollback plan: roll back the application image before applying v021. After v021 is applied, do not downgrade a live database; forward repair is a new migration restoring a nullable compatibility column only if an approved architecture later requires it.

---

## Additional Notes

- This is the first implementation PR. The signed restricted BKN read contract and EE builder follow in separate PRs after this baseline is reviewed.

### DIFF
--- a/bkn-trace/agent-observability/migrations/mariadb/v021/init.sql
+++ b/bkn-trace/agent-observability/migrations/mariadb/v021/init.sql
@@ -1,0 +1,5 @@
+-- Historical provenance projections are task-scoped by interaction_id and
+-- facts_hash. They must not retain tenant as an invented authorization scope.
+DROP INDEX IF EXISTS idx_provenance_projection_scope_tenant ON bkn_trace_ee_historical_provenance_projections;
+ALTER TABLE bkn_trace_ee_historical_provenance_projections
+    DROP COLUMN IF EXISTS tenant_id;
--- a/bkn-trace/agent-observability/migrations/mariadb/v021/schema.go
+++ b/bkn-trace/agent-observability/migrations/mariadb/v021/schema.go
@@ -1,6 +1,7 @@
-// Copyright openbkn.ai
-//
-// Licensed under the OpenBKN License. See LICENSE-OPENBKN.txt in the project root.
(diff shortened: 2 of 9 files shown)


leecky:
Release dependency found during review: [openbkn-ee#55](https://github.com/openbkn-ai/openbkn-ee/pull/55) removes the EE writer's legacy `tenant_id` / `business_domain_id` SQL.

Do not release this migration independently. Coordinate the release: stop the old EE projection writer, apply Core v020/v021, deploy EE #55, then resume consumption. This preserves the task-scoped historical projection contract without reintroducing an invented authorization scope.

PR #1226, opened 2026-09-01
feat(trace): add signed BKN projection reader

leecky:
## Why
Implements the BKN prerequisite for Issue #31 historical projections: Trace can export only the explicitly recorded current knowledge network with a sealed, signed grant.

## Scope
- Adds one internal projection-read endpoint at /api/bkn-backend/in/v1/trace/projection/knowledge-networks/{kn_id}; no ontology-manager alias, list, batch, scan, or reverse lookup.
- Verifies Ed25519 ProjectionReadGrant issuer, audience, key ID, expiry, signature, and exact kn_id scope.
- Fails closed with 404 when disabled, missing, invalid, expired, or out of scope.
- Adds explicit Helm opt-in that injects rotating public keys only; BKN never receives a signing private key.
- Documents the internal contract.

## Verification
- go test ./... -count=1
- go vet ./...
- helm lint and template with default and enabled projection-read configuration
- make api-docs-lint

Related: openbkn-ai/openbkn-ee#31

### DIFF
--- a/adp/bkn/bkn-backend/helm/bkn-backend/templates/deployment.yaml
+++ b/adp/bkn/bkn-backend/helm/bkn-backend/templates/deployment.yaml
@@ -87,6 +87,17 @@ spec:
             value: {{ .Values.bknSafe.url | quote }}
           - name: BKN_SAFE_BASE_URL
             value: {{ .Values.bknSafe.url | quote }}
+          {{- if .Values.bknTrace.projectionRead.enabled }}
+          - name: BKN_TRACE_PROJECTION_GRANT_ISSUER
+            value: {{ required "bknTrace.projectionRead.issuer is required when projection reads are enabled" .Values.bknTrace.projectionRead.issuer | quote }}
+          - name: BKN_TRACE_PROJECTION_GRANT_AUDIENCE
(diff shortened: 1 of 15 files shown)


leecky:
Addressed the review findings in c63ee497a:
- Projection grant verification now authorizes a dedicated ExportKNForProjection path. It does not call user resource filtering or account-name enrichment, so it does not need or synthesize a caller identity.
- Any post-grant source failure is now a bodyless 404, matching the documented capability boundary.
- The alias invariant test now reads the actual RegisterPublic route table.
- Projection reads require the explicit BKN_TRACE_PROJECTION_GRANT_ENABLED=true switch; unrelated partial grant configuration no longer affects BKN startup.

Verified with full go test ./..., go vet ./..., Helm lint/render, and OpenAPI lint.

/review

leecky:
Addressed the remaining export dependency: ExportKNForProjection now assembles the granted network through raw BKN access interfaces only. It no longer calls child List services, so child authorization checks and account-name enrichment cannot reintroduce caller or user-directory dependencies. The export retains network and model definitions, IDs, and names. Full go test ./... and go vet ./... pass.

/review

PR #1273, opened 2026-09-02
fix(vega): split index key fields

### LINKED ISSUE #1240: [Bug][Vega] Incremental build silently misses rows when UUID is used as build key
## Description

Vega batch incremental index build uses `build_key_fields` as a high-water-mark cursor and only reads rows whose key is lexicographically greater than the committed checkpoint. The current validation permits a string UUID primary key as the build key, although newly inserted UUID values are not monotonically increasing. New rows whose UUID sorts before the previous checkpoint are therefore permanently skipped.

Observed resource configuration:

```json
{
  "index_config": { "build_key_fields": ["id"] }
}
```

The `id` source column is `char(36)` and contains UUIDs.

## Steps to reproduce

1. Prepare a source table with 35 rows using random UUID values in `id`; run a full build to completion. The index contains 35 documents.
2. Without changing the source, run an incremental build. It processes 0 rows.
3. Insert 2,000 rows with random UUID values, for 2,035 source rows total.
4. Start a new full build and stop it after its first 1,000-row batch.
5. Without changing the source, start an incremental build and immediately request stop. It processes 155 rows.
6. Run another incremental build. It completes with 0 processed rows.
7. Inspect the published resource index. It contains only 190 documents.

## Expected behavior

Incremental build should index all 2,000 rows inserted after the last completed build, or the product should reject a build key that cannot satisfy the required monotonic high-water-mark contract.

Stopping an incomplete full build must not make subsequent incremental behavior misleading; the incremental baseline and any unpublished full-build index should be observable and safely managed.

## Actual behavior

- The stopped full task writes 1,000 documents to a task-specific index but does not publish that index or checkpoint to the resource.
- The next incremental task correctly starts from the last completed 35-row resource checkpoint.
- Only 155 of the 2,000 inserted UUIDs compare greater than that checkpoint, so only those rows are selected and upserted.
- The checkpoint then advances to the maximum UUID. The following incremental build sees 0 rows.
- The published index contains `35 + 155 = 190` documents; 1,845 newly inserted rows are permanently behind the checkpoint.

The effective query is equivalent to:

```sql
SELECT *
FROM <source_table>
WHERE id > :last_committed_uuid
ORDER BY id
LIMIT 1000;
```

UUID uniqueness is sufficient for deterministic document IDs, but not for a monotonic incremental cursor.

## Environment

- Branch: `main`
- Source connector: MySQL table
- Index backend: OpenSearch
- Build mode: `batch`
- Execute type: `incremental`

- [ ] The supported semantic contract for incremental build keys is explicitly defined: uniqueness, immutability, ordering, monotonicity and late-arriving-row behavior.
- [ ] The system prevents silent data loss when a UUID/random/non-monotonic field is selected as the incremental build key, either by rejecting it or by using an algorithm that captures all new rows.
- [ ] An automated regression test starts with a completed baseline, inserts rows whose keys sort both before and after the checkpoint, and verifies that no promised rows are omitted.
- [ ] Composite-key tests cover equal prefixes, batch boundaries and late-arriving values.
- [ ] Empty follow-up incremental builds do not mask rows skipped behind the checkpoint.
- [ ] Full-build stop behavior remains compatible: an incomplete full index is not published, and unpublished physical indexes have an explicit cleanup/retention policy.
- [ ] API/Studio documentation tells users which fields are safe as incremental build keys.
- [ ] Existing full, incremental stop/resume and checkpoint atomicity tests remain green.

## Boundaries / Risk

- Changing cursor semantics may require checkpoint migration or invalidation for existing resources.
- A timestamp-plus-ID cursor still needs a policy for equal timestamps, clock precision, updates and late-arriving rows.
- Overlap-window rescans require deterministic upsert and may increase source/OpenSearch load.
- CDC/binlog-based resolution introduces connector and deployment changes.
- Existing stopped/failed full tasks may have unpublished OpenSearch indexes; cleanup must not delete a currently published resource index.

## Additional context

Relevant implementation paths:

- `adp/vega/vega-backend/server/worker/build_worker_batch.go`: lexicographic cursor filter, 1,000-row batch loop, incremental checkpoint commit and full-build publication.
- `adp/vega/vega-backend/server/logics/build_task/build_task_service.go`: incremental baseline and build-key validation.
- `adp/vega/vega-backend/server/worker/build_worker_batch_test.go`: current cursor/checkpoint tests.

Suggested diagnostic query, using the checkpoint from the completed 35-row baseline:

```sql
SELECT
  COUNT(*) AS total_count,
  SUM(id > :checkpoint) AS visible_to_incremental,
  SUM(id <= :checkpoint) AS behind_checkpoint
FROM <source_table>;
```

For the observed data distribution this should explain approximately 155 visible rows and 1,845 inserted rows behind the checkpoint.

kenlee1988:
## What Changed

- [x] Replace legacy `build_key_fields` with `primary_key_fields` and `incremental_fields`.
- [x] Use primary keys exclusively for document IDs and incremental fields exclusively for batch cursors/checkpoints.
- [x] Reject streaming task creation and start; update OpenAPI and localized errors.

---

## Why

- Related Issue: Closes #1240
- Related Feature: N/A
- Background: UUID v4 is suitable as a document primary key but not as a monotonically advancing batch cursor. The legacy shared field conflated these responsibilities.

---

## How

- Key implementation points: validate both field groups against schema and their separate supported types; snapshot both groups on task creation; preserve available local index name/status while clearing the checkpoint when either group changes.
- Breaking changes (API, config, database, etc.): `build_key_fields` is removed without compatibility. Existing resources must be reconfigured with both new fields before running a batch build. No database schema migration is required.

---

## Testing

- [x] Unit tests passed locally (`make ci`).
- [ ] Integration tests passed locally (N/A: hermetic unit suite; no integration environment used).
- [ ] Verified in test environment (not performed).

Test notes:

- `make ci` passed: golangci-lint, unit tests, coverage gate.
- `make api-docs-lint` passed; it reports two pre-existing warnings in `docs/api/bkn/action-schedules.yaml`.

---

## Risk & Rollback

- Potential risks: intentional breaking configuration contract; resources configured only with the removed field cannot start new batch builds until reconfigured.
- Rollback plan: revert commit `d59c7e908`; no data or schema migration is involved.

---

## Additional Notes

- Streaming is temporarily unavailable through Create/Start APIs.
- `incremental_fields` use best-effort offline keyset progression; users are responsible for selecting appropriate ordered fields.

### DIFF
--- a/adp/vega/vega-backend/server/drivenadapters/build_task/build_task_access_test.go
+++ b/adp/vega/vega-backend/server/drivenadapters/build_task/build_task_access_test.go
@@ -797,7 +797,10 @@ func sampleBuildTask() *interfaces.BuildTask {
 		FinishTime:       2000,
 		LastProgressTime: 1800,
 		IndexConfig: &interfaces.BuildTaskIndexConfig{
-			IndexConfigContract: interfaces.IndexConfigContract{BuildKeyFields: []string{"id"}},
+			IndexConfigContract: interfaces.IndexConfigContract{
+				PrimaryKeyFields:  []string{"id"},
+				IncrementalFields: []string{"id"},
+			},
 			Features: map[string]interfaces.BuildTaskFieldIndexFeature{
 				"title": {
 					Vector:   &interfaces.SmallModel{ModelID: "embedding", EmbeddingDim: 1024},
--- a/adp/vega/vega-backend/server/drivenadapters/resource/resource_access_test.go
+++ b/adp/vega/vega-backend/server/drivenadapters/resource/resource_access_test.go
@@ -44,7 +44,7 @@ func TestResourceAccessCreate(t *testing.T) {
 				"public.orders",
 				`{"properties":{"row_count":42}}`,
 				`[{"name":"id","display_name":"","type":"integer","description":"","original_name":"","original_type":"","original_description":"","features":null,"attributes":null}]`,
-				`{"build_key_fields":["updated_at","id"],"default_fulltext_analyzer":"ik_max_word","default_embedding_model":"embedding"}`,
+				`{"primary_key_fields":["id"],"incremental_fields":["updated_at","id"],"default_fulltext_analyzer":"ik_max_word","default_embedding_model":"embedding"}`,
 				"",
 				"[]",
 				interfaces.ResourceLocalIndexStatusAvailable,
@@ -110,7 +110,8 @@ func TestResourceAccessGetByID(t *testing.T) {
 		require.NotNil(t, got)
 		assert.Equal(t, "resource-1", got.ID)
 		require.NotNil(t, got.IndexConfig)
-		assert.Equal(t, []string{"updated_at", "id"}, got.IndexConfig.BuildKeyFields)
+		assert.Equal(t, []string{"id"}, got.IndexConfig.PrimaryKeyFields)
+		assert.Equal(t, []string{"updated_at", "id"}, got.IndexConfig.IncrementalFields)
 		assert.Equal(t, interfaces.ResourceLocalIndexStatusAvailable, got.LocalIndexStatus)
 		assert.Equal(t, "vega-build-resource-1-task-1", got.LocalIndexName)
 		assert.Equal(t, `{"mode":"batch","cursor":[10,"a"]}`, got.SyncMark)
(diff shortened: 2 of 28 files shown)


kenlee1988:
已修复复评中的阻塞问题：无 checkpoint 的资源重配置主键/增量键时不再执行无变化的状态更新，从而避免 MySQL 返回 0 行受影响并回滚；同时将键字段比较改为按元素比较，避免 `nil` 与空数组造成误触发。已新增回归测试并通过 `make ci`。\n\n`/review`

PR #1275, opened 2026-09-02
feat(sandbox): bake in the Python OSDK and make it the recommended BKN face

sh00tg0a1:
Acts on the handoff from `openbkn-ai/bkn-sdk#86`: `bkn-sdk/docs/handoff/2026-09-02-bake-python-osdk-into-sandbox.md`.

**The two BKN faces coexist.** `sandbox_sdk.bkn` keeps working exactly as before — existing functions need no rewrite — while `bkn-osdk` becomes what the scaffolds and generator prompts teach for new ones.

## Why coexistence, not replacement

Replacing was never viable, and not merely as a matter of caution — the signatures differ in ways no shim can bridge:

```python
# built-in face: no kn_id, because it POSTs to the MCP tool where the
# resource is named by the {{.id}} placeholder inside the SQL
def run_sql(sql: str, query_timeout: int = None, response_format: str = 'json') -> dict

# OSDK: the REST route needs the network id
def run_sql(kn_id: str, sql: str, *, ...) -> Any
```

A stored function calling `bkn.run_sql(sql=...)` has no network id to forward. 0.1.4 is deployed on the VM, test server 23 and the PoC, so those stored functions exist.

## 1. Bake `bkn-osdk` into the template image

Sandbox Python that reads a knowledge network had to `pip install` once per session — 25-30 seconds, and only where the sandbox can reach github.com at runtime. `/opt/sandbox-common` exists for exactly this: its header already states the goal as avoiding per-session installs and working in no-egress deployments.

| Check | Result |
|---|---|
| `https://github.com/openbkn-ai/bkn-sdk/archive/00e128e.zip` | 200, redirects to codeload |
| Package | `bkn-osdk` 0.1.0, from `python/pyproject.toml` |
| `requires-python` | `>=3.10`; the image runs 3.11 |
| Runtime deps | `httpx>=0.27` only — already pinned here at `0.27.2` |
| Repo visibility | public, so no build-time credential |

Pinned by sha, not branch: pip's wheel cache keys on the URL, so rebuilding against a branch reports success while reinstalling the cached older build and leaving the previous commit in `direct_url.json`.

Only `python-basic` gets it, and that needed no change — `build.sh` already defaults `PREINSTALL_COMMON_TEMPLATES` to that template alone.

## 2. `BKN_BASE_URL` joins `BKN_SANDBOX_MCP_URL`

Added through the whole chain: configmap, both charts, both schedulers, bwrap's `--setenv` passthrough.

**Both are live settings**, because each face reads its own:

| Variable | Face | Shape |
|---|---|---|
| `BKN_SANDBOX_MCP_URL` | `sandbox_sdk.bkn` | full MCP path, trailing slash required — without it the gateway answers 307 and that face's `urllib` does not follow redirects on POST |
| `BKN_BASE_URL` | `bkn-osdk` | host and port only; the SDK appends `/api/agent-retrieval/v1/kn/` and `/api/agent-retrieval/v1/mcp` itself |

A deployment that sets only the old key still gets working OSDK access: `resolve_bkn_base_url()` derives the base by stripping the MCP path. No values change is required to upgrade.

## 3. Per-execution reset for both faces

`dispatch()` resets both rather than one. The built-in face takes the event as before; for the OSDK, a no-argument `bkn_osdk.configure()` replaces the process default with an empty one, so user code that called `configure()` cannot leak its address or token into the next caller of the same pooled process.

Zero-config otherwise needs no wiring: `bkn-osdk` already reads `BKN_BASE_URL`, `BKN_TOKEN`, `BKN_CONVERSATION_ID` and `BKN_INTERACTION_ID`, and re-reads `os.environ` on every call rather than caching. The import is guarded — an older image has no `bkn_osdk`, and a function using only the built-in face should not fail for its absence.

## 4. Scaffolds and prompts

Both generator prompts and both locale scaffolds now teach `from bkn_osdk import kn`, and say plainly that the older form still works and that an existing function keeps whichever form it already uses unless the request asks to migrate. Signatures were checked against the SDK: `kn.run_sql(kn_id, sql)` takes the network id first, where `bkn.run_sql(sql=...)` did not.

`sandbox_sdk/bkn.py` gains a status note at the top of its docstring: not recommended for new functions, still supported.

## Verification

- `infra/sandbox/runtime/executor`: **148 passed**
- `sandbox_control_plane` scheduler + config: **114 passed, 2 failed** — both pre-existing `test_build_s3_mount_entrypoint*` failures asserting on a `mount --bind` string this code no longer emits, untouched by this diff
- `helm template` renders both charts
- Hard-coded Chinese check passes, 0 new; the baseline file is byte-identical to `main`

**Not verified end to end.** Nothing here has been deployed. Specifically unproven: that `import bkn_osdk` resolves inside the built image, that `BKN_BASE_URL` survives all three hops (scheduler → container → bwrap `--setenv`), that a function calling `kn.get_kn_detail()` returns data, and that the derive-from-legacy-key fallback works against a real control plane. The SDK-side verification in the handoff used the runtime-installed SDK with the built-in face still present, so it does not carry over. Worth running on test server 23 before merge.

Also fixes a pre-existing `NameError` in `test_bwrap.py`, which referenced `settings` without the import its neighbour does. The sandbox workflow runs chart-only tests, so nothing caught it.

## Air-gapped builders

The image build now needs github.com reachable, where before it only needed the pip mirror. Runtime egress is unaffected — removing the runtime requirement is the point.


sh00tg0a1:
已按 review 修改并推送 `05bf7c69`。

**阻塞项 `7de70775`（multi-language 无 OSDK）—— 已修。**

确认属实:`sandbox_sdk` 是无条件 COPY 的(Dockerfile:52),而 `common-requirements.txt` 只在 `PREINSTALL_COMMON=true` 时安装,CI 只给 python-basic 开。生成提示词又向作者声明「已预装进沙箱镜像」,不区分模板。

修法不是给 multi-language 也开 `PREINSTALL_COMMON`——那会把 numpy/pandas/scipy/sklearn/statsmodels 一起拖进去,正是「multi-language stays lean」要避免的。改为把 SDK 拆到独立的 `bkn-requirements.txt` **无条件安装**,数据栈保持 opt-in。BKN 能力面应当与它并列的那个面一样普遍。

multi-language 的代价是 SDK 加 httpx 及其传递依赖,实测 5.1 MB。

**待确认项:BKN_BASE_URL 默认值让兜底形同虚设 —— 已修。**

这条提得准。主 chart 原本默认 `http://agent-retrieval:30779`,而 `resolve_bkn_base_url()` 优先取它,README 承诺的「只设旧键也能工作」那条路根本不会跑。BKN 在别的命名空间、在 `BKN_SANDBOX_MCP_URL` 里填了 FQDN 的存量部署,升级后会被静默指向一个在那里解析不了的短名。

现在默认空,推导路径真正生效;新装要填的地址写在 values 注释与 chart README 里。

**待确认项:`dispatch()` 只挡 ImportError —— 修了,但方向与建议相反。**

放宽成吞掉所有异常是不对的:重置失败意味着上一个调用方的地址与令牌可能还留在进程里,而这个池化进程马上要跑另一个调用方的代码——继续执行等于用错身份。这里应当失败,不该降级。

改成捕获后抛一个说清原因的 `RuntimeError`,而不是把 SDK 原始异常直接抛出去。纯计算函数仍然不受影响:模块缺失走 `ImportError` 分支跳过,只有 `configure()` 真的抛了才中止。

---

**第一轮两条待确认项的答复:**

- **bwrap `--unshare-net` 下用户代码无网络** —— 属实(`bwrap.py:135-136`),但这对两个面**同等**:`sandbox_sdk.bkn` 在 bwrap 下同样打不通。不是本 PR 引入的回归。k8s 部署里 configmap 的 `DISABLE_BWRAP` 默认 `"true"`,这也是内建面今天能工作的原因。环境变量那半不成立:`--clearenv` 之后 `_build_execution_env` 会显式 `--setenv` 回传 `BKN_TOKEN` / `BKN_BASE_URL` / `BKN_SANDBOX_MCP_URL`,已实测。
- **`USE_MIRROR` 不覆盖直连 URL** —— 属实,PEP 508 直连 URL 不走 index。已在 `bkn-requirements.txt` 顶部注明:镜像源构建仍需构建机能连 github.com。彻底解决要把包发到内部源或改用 vendored wheel,不在本 PR 范围。

---

**验证补充**(上一轮 PR 正文说「端到端未验」,现已补):

用 CI 发的真镜像 `shac0e3448`,在 VM 上打活的 BKN,9 个能力实测通过——`get_kn_detail` / `search_schema` / `get_object_types` / `query_object_instance`(取到真实例)/ `search_instance` / `list_resources` / `list_skills` / `list_action_executions` / `run_sql`(占位符聚合返回 `{'n': 3}`)。

同一进程里**两个面都通**:osdk 自动开轮次(`conv_221b…/int_c181…`),内建面经真的 `dispatch()` 拿到同样结果。零配置成立——容器里只给 `BKN_BASE_URL` + `BKN_TOKEN`。

仍未验:控制面注入那条完整链路(scheduler → Pod → bwrap)组装起来跑,三个环节分别验过。

PR #1285, opened 2026-09-03
feat(bkn): expose child resource operations

### LINKED ISSUE #1284: feat(bkn): return child resource operations in knowledge-network APIs
## Goal

Return the current accessor allowed operations for every knowledge-network child resource in list and detail API responses, so Studio can render record-level actions correctly.

## Module

bkn

## Background / Context

Parent Epic: #1241.

The child-resource PEP implemented by #1247 currently filters list visibility using the view_detail operation, but discards the returned operation set. Knowledge-network root responses populate the operations field; child responses do not. Concept groups declare the field but never populate it, while object types, relation types, action types, metrics, and risk types do not declare it.

Studio #986 must use record-level operations to show or hide edit, delete, mapping configuration, permission configuration, query, task-management, and execute entries.

- [ ] List and detail responses for concept groups, object types, relation types, action types, metrics, and risk types include the operations field for the current accessor.
- [ ] A child list first requires view_detail for visibility, then returns only the allowed operations for each visible canonical child resource in the kn_id/child_id form.
- [ ] Candidate operations match the #1241 contract: non-action children return view_detail, query_data, modify, delete, and authorize; action types additionally return task_manage and execute.
- [ ] The implementation does not use root COMMON_OPERATIONS or return instance-level create.
- [ ] The default value and rollout behavior of KN_CHILD_RESOURCE_PEP_ENABLED are unchanged; enabling it remains a deployment configuration decision after end-to-end validation.
- [ ] Focused unit tests cover exact operation projection, canonical IDs, list visibility, and detail responses.
- [ ] Relevant BKN backend CI checks pass.

## Boundaries / Risk

- Do not modify KN_CHILD_RESOURCE_PEP_ENABLED, Helm values, deployment manifests, Safe Policy data, or ResourceParent data in this task.
- Do not deploy or run migrations.
- This response-field addition is consumed by Studio #986; it must preserve the existing list envelopes and pagination behavior.

## Additional context

- Parent Epic: #1241
- Existing child PEP/list filtering: #1247
- Studio consumer: openbkn-ai/bkn-studio#986

rongwei-liu:
## What Changed

- [x] Return instance-level operation lists for concept groups, object types, relation types, action types, metrics, and risk types.
- [x] Preserve legacy authorization behavior while the child-resource PEP flag is disabled.
- [x] Document the operations response field in the BKN OpenAPI specifications.

---

## Why

- Related Issue: Closes #1284
- Related Feature: Knowledge-network child-resource authorization
- Background: Studio needs an authoritative per-resource operation set to decide which actions are visible to each accessor.

---

## How

- Added canonical child-resource operation candidates and permission projection helpers.
- List and single-resource APIs attach operations only when the child-resource PEP flag is enabled.
- Breaking changes: None by default. The feature remains inactive until the existing deployment flag is explicitly enabled.

---

## Testing

- [ ] Integration tests passed locally (not applicable: the change is covered by mocked unit tests)
- [ ] Verified in test environment (not deployed)

Test notes:

- go vet ./...
- I18N_MODE_UT=true go test ./... -count=1
- make api-docs-lint

---

## Risk & Rollback

- Potential risks: Enabling the child-resource PEP requires corresponding authorization records in the permission service.
- Rollback plan: Keep or restore the deployment flag to false; the previous parent knowledge-network authorization behavior is retained.

---

## Additional Notes

- The deployment configuration is intentionally unchanged in this PR.
- Studio integration should be enabled only after this API contract is deployed and verified.

### DIFF
--- a/adp/bkn/bkn-backend/server/interfaces/action_type.go
+++ b/adp/bkn/bkn-backend/server/interfaces/action_type.go
@@ -119,7 +119,8 @@ type ActionType struct {
 	Updater    AccountInfo `json:"updater" mapstructure:"updater"`
 	UpdateTime int64       `json:"update_time" mapstructure:"update_time"`
 
-	ModuleType string `json:"module_type" mapstructure:"module_type"`
+	ModuleType string   `json:"module_type" mapstructure:"module_type"`
+	Operations []string `json:"operations,omitempty"`
 
 	IfNameModify bool `json:"-"`
 	// Vector.
--- a/adp/bkn/bkn-backend/server/interfaces/metric.go
+++ b/adp/bkn/bkn-backend/server/interfaces/metric.go
@@ -273,6 +273,7 @@ type MetricDefinition struct {
 	Updater    AccountInfo `json:"updater,omitempty" mapstructure:"updater"`
 	UpdateTime int64       `json:"update_time,omitempty" mapstructure:"update_time"`
 	ModuleType string      `json:"module_type,omitempty" mapstructure:"module_type"`
+	Operations []string    `json:"operations,omitempty"`
 
 	Vector []float32 `json:"_vector,omitempty"`
 	Score  *float64  `json:"_score,omitempty"`
--- a/adp/bkn/bkn-backend/server/interfaces/object_type.go
+++ b/adp/bkn/bkn-backend/server/interfaces/object_type.go
@@ -104,7 +104,8 @@ type ObjectType struct {
 	Updater    AccountInfo `json:"updater" mapstructure:"updater"`
 	UpdateTime int64       `json:"update_time" mapstructure:"update_time"`
 
-	ModuleType string `json:"module_type" mapstructure:"module_type"`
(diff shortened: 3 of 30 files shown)


rongwei-liu:
Addressed the review feedback in fac7db16d.

- The legacy projection path is now reachable in every child-resource list API when KN_CHILD_RESOURCE_PEP_ENABLED is false. The APIs preserve the existing parent knowledge-network view check, then project the returned parent operations onto each listed child resource. This prevents Studio from treating an administrator as view-only before the child PEP flag is enabled.
- When the child PEP is enabled, single-resource detail APIs use GetKNChildOperations directly rather than performing a preceding CheckPermission call, so the Safe lookup is not duplicated.
- Restored the total-query error logging and span error status in the affected list services.
- The multi-ID detail helpers intentionally remain bulk/internal retrieval APIs and do not project per-item operations; Studio uses the single-resource detail endpoints for its authorization UI.

Validation completed:

- go vet ./...
- I18N_MODE_UT=true go test ./... -count=1

/review

PR #1288, opened 2026-09-03
fix(trace): retain metrics in signed projection exports

leecky:
## What

Retains `metrics` from BKN's existing single-network projection export in `KnowledgeNetworkDetail`.

## Why

Historical facts containing `metric_id` otherwise become permanently `not_evaluable` after the DTO drops a field BKN already exports.

## Boundary

No endpoint, identity propagation, Vega call, or new authorization path is added. EE reads metrics from the cached signed network snapshot.

## Verification

`go test ./server/interfaces -run TestKnowledgeNetworkDetailDecodesExportedMetrics -count=1`


leecky:
Closing this DTO change after verification: `KnowledgeNetworkDetail` is shared by get_kn_detail, so retaining the full exported metrics list there would expand its summary response. The fix moved to EE's projection-only reader, which privately decodes metrics from the same signed single-network response. No Foundry endpoint or public DTO change is needed.

Issue #1334, opened 2026-09-06
bug(bkn-trace): preserve managed interaction previews in conversation summaries


leecky:
[...]

PR #1335, opened 2026-09-06
fix(bkn-trace): preserve conversation terminal previews

leecky:
## What Changed

- [x] Load terminal question/result artifacts by canonical Interaction IDs for the selected conversation page.
- [x] Batch-load lifecycle interactions and use them for conversation turn counts.
- [x] Preserve the legacy Trace-based read path when no lifecycle Interaction exists.

---

## Why

- Related Issue: #1334
- Background: The conversation list identifies rows by `conversation_id`, but terminal artifacts can be scoped only to `interaction_id` and omit `trace_id`. The fast path replaced the conversation selector with Trace IDs, so question/result previews could be empty even when the interaction detail remained complete.

---

## How

- Add a normal scoped `InteractionIDs` projection selector; it is distinct from the pre-authorized Core handoff.
- Resolve selected conversations to their lifecycle interactions in one batch store query.
- Query interaction-scoped artifacts without dropping the normal access Scope.
- Breaking changes: none; no schema migration, API shape, or configuration change.

---

## Testing

- [x] Integration-style projection adapter tests passed locally
- [ ] Verified in test environment

Test notes:

- `make test`
- `make lint`
- `make license-check`
- `make build`

---

## Risk & Rollback

- Potential risks: The list fast path adds one bounded batch interaction lookup and a terms selector for the selected page.
- Rollback plan: Revert this single commit; legacy Trace-based behavior is restored.

---

## Additional Notes

- Regression coverage includes trace-less terminal artifacts, preserved scope filtering, legacy fallback behavior, and lifecycle-based turn counts.

### DIFF
--- a/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
+++ b/bkn-trace/agent-observability/src/domain/service/evidencesvc/summary.go
@@ -1546,9 +1546,20 @@ func (s *Service) listConversationIdentityPage(ctx context.Context, options evid
 		return evidencevo.ConversationSummaryPage{Entries: []evidencevo.ConversationSummary{}, Total: identityPage.Total, Page: normalizeSummaryPage(options.Page), PageSize: normalizeSummaryLimit(options.Limit)}, true, nil
 	}
 	ids := summaryIdentityIDs(identityPage.Entries)
+	interactionIDs, interactionsByConversation, err := s.listCanonicalInteractionIDs(ctx, ids)
+	if err != nil {
+		return evidencevo.ConversationSummaryPage{}, true, err
+	}
+	terminalArtifactTypes := []evidencevo.ArtifactType(nil)
+	if len(interactionIDs) > 0 {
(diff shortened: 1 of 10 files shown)


leecky:
Fixed the blocking compatibility issue:

- Preserve the existing Trace-scoped artifact read, including artifacts without `interaction_id` and all evidence types used for completeness.
- Add a separate, scoped terminal preview read for canonical Interaction IDs limited to `question` and `result`, then deduplicate by artifact ID.
- Added an adapter regression test asserting both queries and both Scope filters.

Local validation: `make test`, `make lint`, `make license-check`, and `make build`.

/review

leecky:
Aligned the in-memory projection with the production compatibility contract without adding a second in-memory query:

- When the conversation list supplies Interaction preview selectors, the in-memory scope pass retains legacy artifacts rather than filtering them out.
- Added a regression test covering an artifact without `interaction_id` alongside a terminal preview artifact.

Local validation: `make test`, `make lint`, `make license-check`, and `make build`.

/review

PR #1389, opened 2026-09-08
feat(bkn-backend): compare two knowledge networks and count the data behind them

sh00tg0a1:
# Pull Request

## What Changed

Compare two knowledge networks, definition by definition and field by field, and count the data behind an object type on each side.

- [x] `bkn-specification`: field-level comparison (`DiffNetworkModels`), and the checksum now covers the frontmatter
- [x] `bkn-backend`: `POST /bkns/diff` and `POST /bkns/diff/object-data-stats`
- [x] `docs/api/bkn/bkn.yaml`: both endpoints and their schemas

**This PR shows the whole `release/0.1.4-yf` line against `release/0.1.4`, which is 15 commits.** Nine of them (`b289194db`..`60a281a06`) are the openCypher subset that was already frozen onto this line before this work started. The six under review here are `8edeec260`..`555a7f6a0`.

---

## Why

- Related Issue: none. `#1229` (本体版本与变更管理) and `#1232` are the adjacent epics; this is comparison between two networks, not version management, and was requested directly.
- Background: there was no way to see what differs between two knowledge networks. The repo already carried a differ (`bkn-specification/bkn/differ.go`) with **zero callers and zero tests**; it compared file checksums, so it could say an object type changed but not what changed inside it.

---

## How

- Key implementation points:
  - **The checksum covered the body only.** Everything in the frontmatter — name, tags, the capability block — sat outside it, so renaming a definition produced a byte-identical hash and a diff reported "unchanged" for a file that had visibly changed. Verified before and after on a real example network. CHECKSUM files already shipped inside `.bkn` packages declare no format, are read as format 1 and verify exactly as before, so no existing package starts warning on import.
  - **Field walking is reflective**, not six hand-written comparators. Hand-written ones go stale silently: a field added to the model is simply never compared and nothing fails.
  - **Paths are keyed by name**, so `data_properties[alt_part].description` is stable under reordering; a reordered table produces no changes at all.
  - **Lineage is reported up front.** Networks built independently share no definition ids — ids are generated per network — and their comparison degrades into a wall of creates and deletes. Name pairing is offered as the way out, opt-in, carrying both ids on any entry it produced.
  - **Counting is a separate endpoint** because it queries the customer's own database. Opening a comparison does not pay for it; only the object type someone opens does. Two counts only: per-column statistics cost another pass over the table per column.
  - Primary keys are mapped through the data properties to their physical columns; a key with no mapped column fails rather than counting the wrong thing. A composite key is concatenated rather than passed as several arguments to `COUNT(DISTINCT ...)`, which is MySQL's own form and this statement is transpiled to the connector's dialect.
- Breaking changes: none. No migration, no change to any existing endpoint's response. `include_unchanged` defaults to false, so callers that do not ask do not pay for the identical definitions.

**Landing this on `main` needs more than a merge.** The data statistics call `VegaBackendAccess.RawQuery`, which exists only on this line (added by `bf25105d0` for Cypher); `main` also lacks `BknNetwork.BusinessDomain`, the `VEGA_PAGING_MODE_SINGLE` constant and `replyHandlerError`. `bf25105d0` is self-contained (4 files, +245 lines, adapter + interface + mock + test) and can be cherry-picked without the Cypher subset.

---

## Testing

Test notes:

- `go build ./...` clean; `bkn-specification`, `logics/bkn`, `logics/object_data_stats`, `driveradapters` all green. The differ had no tests at all before this; it now has them, including the frontmatter-only edit that used to slip through and a stability check against Go's map iteration order.
- redocly lint clean (the 2 warnings are pre-existing, in the file header). `tools/check_i18n_catalogs.py` passes.
- Deployed to the VM (`10.211.55.4`, main-line backend + these commits cherry-picked) and exercised against real networks:
  - two unrelated networks: `created:125 updated:4 deleted:24 unchanged:0`, lineage `4/153`
  - a network against itself: `unchanged:28`, network-level entry `skip`
  - counts: `世界杯赛事` 30 rows, `国家队` 88 rows, delta 58 — matching the same statement run directly against vega
- **One defect the unit tests could not have caught** was found there and fixed in `c020e14a3`: the vega adapter decodes with `UseNumber`, so counts arrive as `json.Number`; `asInt64` had no case for it and fell through to zero, reporting an empty table for a resource with thirty rows, with no error anywhere. The regression test now uses that shape.

---

## Risk & Rollback

- Potential risks:
  - The checksum format change is the only thing that touches existing data. It is versioned: old CHECKSUM files verify under format 1 unchanged, new ones declare `# format: 2`. The four example networks were regenerated and `validate-dir` reports `errors: null` for all of them.
  - `POST /bkns/diff` loads both networks in full. Authorization is not repeated in the handler — each side goes through `GetKNByID`, which checks the caller against that network, so a comparison cannot be used to read a network sideways.
  - Counting runs a statement against the customer's database, gated on `query_data` for the network, with vega authorizing the resource itself against the same caller.
- Rollback plan: both endpoints are additive. Reverting the six commits removes them and restores the previous checksum coverage; packages written under format 2 would then be read as format 1, which under-reports frontmatter changes but does not fail.

---

## Additional Notes

- The comparison page lives in `bkn-studio` on the branch of the same name; that PR is separate.
- Known gap, not addressed here: `lineage.related` is `common_ids > 0`. Two networks sharing 4 of 153 ids are reported as related, which is not useful. It wants a ratio threshold, and the threshold is a product call.


sh00tg0a1:
Closing: wrong base. This targeted release/0.1.4, which would merge the 0.1.4-yf customer line — including the openCypher subset frozen onto it — into the 0.1.4 release line. The work is already on release/0.1.4-yf and is not intended for release/0.1.4.

PR #1391, opened 2026-09-08
feat(execution-factory, context-loader): rank Skills, Function tools and MCP tools in one space

sh00tg0a1:
Closes #1370 · Epic #1266 · consumes #1296 and #1379

## Why

A knowledge network can mount three kinds of capability, and each was retrieved
by its own rule: Skills through a Vega dataset over BM25, Function tools through
a SQL `LIKE`, MCP tools through a literal substring match inside Context Loader.
Context Loader concatenated the three lists and called it a ranking. It was not
one — BM25 scores are unbounded, `LIKE` has no score at all — so the order only
said which list came first. An agent asking "what can convert currency" got
three lists to merge, and a network mounting one excellent MCP tool and twenty
mediocre toolbox tools showed the twenty first.

## What this does

One dataset, `bkn_execution_factory_capability_dataset`, holding all three under
the three-part identity the bindings already use, and one retrieval endpoint
that ranks them together.

**Retrieval** issues two channels and fuses them by rank. The vector channel puts
the whitelist inside the knn field object, which vega renders as its filter: that
placement makes it a pre-filter, so a network that mounted three capabilities
gets the best of its three rather than the intersection of its three with a
global top-k. The channels are separate requests, never one boolean query — knn
scores live in 0..1 while BM25 has no ceiling, and adding the clauses pushes
every vector hit out of the candidate pool.

**Writes** reach the index immediately and are reconciled underneath. Skills are
mirrored at the Skill index sync, the single point every Skill write already
passes through. Function tools are synced at each tool write, after the
transaction commits, by re-reading the rows rather than trusting the caller's
copy — so a write that later rolled back cannot leave a phantom. MCP tools are
synced from the MCP service, which owns the client that can list a remote server.
Each kind also has a periodic pass, because immediate writes cannot cover a
recreated dataset, a future write path nobody hooked, or a remote server whose
tool list moves without telling anyone.

**The Skill-only dataset is retired.** Its schema, embedding model, rebuild and
restore all existed to answer one question that the shared index now answers.
The vega resource is left in place rather than deleted on startup: deleting it
would make a rollback unrecoverable.

## Details worth stating

**The whitelist filters on a composite `capability_key`.** A thousand mounts as
three-clause AND groups exceeds `indices.query.bool.max_clause_count`; one terms
filter over composite keys is a single clause at any size.

**The document id is a hash of that key.** The id travels in a URL path segment
and an MCP tool name comes from a remote server that is not ours to constrain; a
[...]

PR #1396, opened 2026-09-08
fix(execution-factory): keep the capability index able to heal after a lost index

sh00tg0a1:
Follows #1370 (merged in #1391). Fixes a defect that landed with it.

## What is broken on main

`Init` gained a lock in the last review round, and with it one line too many:

```go
s.initMu.Lock()
defer s.initMu.Unlock()
if s.isInitialized() {
    return nil
}
```

`initialized` is set inside `Init` and cleared nowhere, so that shortcut makes
`Init` a one-shot. The three reconcilers call it on every pass, and that pass is
what rebuilds a dataset whose managed index was later lost or marked
unavailable — the exact failure this feature exists to survive.

With the shortcut, such a dataset waits for a process restart. Every write in
between answers 400 `dataset resource has no available local index`. That is the
same shape as #1381, where recovery meant a human deleting a vega resource by
hand because nothing questioned the row again.

The review on #1391 caught this after the PR had already merged, which is why it
comes as a follow-up rather than a fixup.

## The change

The lock alone is what the race needed — two callers in the rebuild branch at
once, the second delete removing the dataset the first had just created.
Serialising them is enough; nothing about the race required skipping the work.

The comment now says why there is no shortcut, so the line does not come back.

## Tests

The concurrency test asserted "created exactly once", which pinned the one-shot
behaviour rather than the ordering. It now pins that two creates never overlap.
A second test pins that a later pass still notices an index that went away — the
property this fix restores.

## Test plan

- [x] `go test ./server/logics/capability/... -race`
- [x] `go test ./server/logics/...` in execution-factory

### DIFF
--- a/adp/execution-factory/operator-integration/server/logics/capability/index_sync.go
+++ b/adp/execution-factory/operator-integration/server/logics/capability/index_sync.go
@@ -116,8 +116,17 @@ func (s *capabilityIndexSync) Init(ctx context.Context) (err error) {
 	s.initMu.Lock()
 	defer s.initMu.Unlock()
 
-	initialized := false
-	defer func() { s.setInitialized(initialized) }()
+	// Only ever promote. A failed re-check means this pass could not confirm the dataset, not that
+	// the dataset stopped existing — and Init now runs on every reconciler pass, so demoting here
+	// would let one timed-out call to vega mark a working service unready for up to the reconcile
+	// interval. Writes in that window are skipped, and a skipped delete is never retried by
+	// anything: the MCP pass only walks servers that still exist, so a server deleted during it
+	// would leave its tools in the index permanently.
+	defer func() {
+		if err == nil {
+			s.setInitialized(true)
+		}
+	}()
 
 	catalogID, err := s.ensureCatalog(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


sh00tg0a1:
阻塞项 `c2afcff7` 已修,而且比报告的更该修 —— 我原来那版把「数据集是否确认存在过」和「现在能不能写」压在同一个标志上,`defer` 又无条件写入本轮结果,所以一次 vega 抖动就能把已就绪的服务打回未就绪,而删除在该状态下静默成功。删除没有任何补偿路径(对账器从源头算期望集合,已删的东西再也不会出现在里面),所以那是永久残留,不是延迟。

两处改动,对应那个标志被混用的两件事:

**Init 只升不降。** 失败只说明这一轮没能确认数据集,不说明数据集消失了。

**删除失败就报错**,不再返回 nil。写入仍然跳过且安静 —— 这个不对称正是要点:跳过的写入下一轮会被重写,跳过的删除没有人会重做。

补了两个测试:一次瞬时失败后 `isInitialized()` 仍为 true;未就绪时 `DeleteCapability` / `DeleteOwner` 报错而 `UpsertCapability` 仍返回 nil。

**待确认项**(重建窗口):我的判断是可接受。重建期间标志保持 true,所以那段时间的写入会**明确报错给调用方**而不是被丢掉,三个对账器各自在下一轮把自己那一类补回索引。响亮且收敛,好过安静。真要消除这段空窗,得让重建变成「建新索引 → 原子切换 → 删旧」,那是 vega 侧的能力,不在这一层。

/review

PR #1397, opened 2026-09-08
feat(context-loader): one search_capabilities entry over every mounted capability kind

sh00tg0a1:
Closes #1388 · builds on #1370 (merged in #1391)

## Why

Skills, Function tools and MCP tools have shared an index and a ranking since
#1370, but the MCP surface still asked in two places: `find_skills` for one kind,
`search_tools` for the other two. An agent asking what can convert currency
through `search_tools` never saw the Skill that does it, and merging two answers
with no comparable score was left to the agent.

## What this does

`search_capabilities` asks once. On the test server, 「库存预警」 against a
network mounting three Skills and twenty-one API tools returns them interleaved
by the fused rank, with a Skill in first place — an answer neither existing tool
could give.

Every entry carries `capability_type`, which is what makes one list usable: a
`function` or `mcp_tool` is called through `execute_tool` with arguments matching
its input schema, a `skill` is read through `get_skill_content`. Tools carry that
schema and their use rule; Skills carry neither, because a Skill is read rather
than called — the same minimal shape `find_skills` has always returned.

Narrowing is the two levels the product itself uses: `types` for the kind,
`metadata_types` for the tool box kind. The four buckets every list in the UI
shows map onto one request.

## search_tools now delegates

The claim was that `search_tools` is `search_capabilities` with the kinds pinned.
It started as a resemblance — two ref builders, two message switches, two
truncation handlings — and that lasted exactly one review cycle before it cost
something: the truncation probe from #1391's review landed on one of them and not
the other, so the new entry was written with the bug the review had just found in
the old one.

It delegates now, so the claim is a fact and one fix covers both. What stays is
the shape of its answer: `tool_id` / `toolbox_id` / `input_schema`, which callers,
the sandbox SDK and the API contract are written against.

## The two narrow tools stay

They are marked deprecated in their descriptions and otherwise unchanged. They
have real callers — the sandbox SDK exports `search_tools`, and
`docs/api/context-loader/tool.yaml` documents `/kn/search_tools` as a contract
whose `execute_tool` section tells readers to take ids from it. Removing them is
a later change with migration work of its own, and it should remove both:
dropping only `search_tools` would leave one unified entry beside one narrow one,
which is the same inconsistency in a new shape.

## Sandbox SDK

Regenerated from the MCP tool catalogue (`make -C infra/sandbox bkn-tools`), so
agents writing code in a sandbox can call `search_capabilities` directly. CI's
consistency check would fail without it.

## Verified on the test server

| case | result |
|---|---|
[...]

PR #1399, opened 2026-09-08
test(context-loader): pin the empty-result message to the caller's own filters

sh00tg0a1:
Follows #1388 (merged in #1397). Adds a regression test that a previous commit
claimed to add and did not.

## What happened

The round that fixed the empty-result message said it had replaced the "no match
at all" case with one asserting that the message does not name a parameter the
caller cannot pass. It had not. The edit matched on text differing by one
character, silently changed nothing, and the case it meant to replace — asserting
only `Message != ""` — passes under every branch. Nothing went red, and I
reported the work as done.

The review on #1397 caught the discrepancy and marked it as the one fix left
without a regression test. It was right.

## The test

`search_tools` pins the capability kinds when it delegates to
`search_capabilities`. If that counted as a caller's filter, an empty answer
would tell the caller to drop a `types` parameter it never set and has no input
for. The fix distinguishes intrinsic narrowing from requested narrowing; this
pins it.

Verified by removing the fix rather than by trusting a green run: with the
`kindsAreIntrinsic` branch reverted the case fails with the exact message a
caller would have received —

```
不该让调用方去改它传不了的参数: "该知识网络挂载了能力，但没有 types / metadata_types
指定类型的。去掉这两个参数可看到全部已挂载能力。"
```

— and passes again with the fix restored. The second case pins the other
direction: a caller who does pass `metadata_types` is told about it.

## Test plan

- [x] `go test ./server/logics/kntools/...`
- [x] Fix reverted locally to confirm the assertion fails, then restored

### DIFF
--- a/adp/context-loader/agent-retrieval/server/logics/kntools/capabilities.go
+++ b/adp/context-loader/agent-retrieval/server/logics/kntools/capabilities.go
@@ -139,9 +139,14 @@ func (s *knToolsService) SearchCapabilities(ctx context.Context,
 	switch {
 	case len(entries) == 0 && total > 0:
 		resp.Message = infraErr.LocalizedDetail(ctx, "ToolsMatchedButNotVisible")
-	case len(entries) == 0 && !req.kindsAreIntrinsic && len(req.Types) > 0,
-		len(entries) == 0 && len(req.MetadataTypes) > 0:
+	case len(entries) == 0 && !req.kindsAreIntrinsic && len(req.Types) > 0:
+		// Both filters are the caller's, so both can be named.
 		resp.Message = infraErr.LocalizedDetail(ctx, "NoCapabilitiesOfRequestedKind")
+	case len(entries) == 0 && len(req.MetadataTypes) > 0:
+		// Only the tool box kind was the caller's. Naming types here would repeat the original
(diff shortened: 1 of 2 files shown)


sh00tg0a1:
两条待确认项都改了,第二条是真缺陷不只是覆盖问题。

**提示仍在让调用方去掉 types** —— 确认。我上一轮只修了「哪一支命中」,没动那一支的**文案**,所以缺陷在下一层原样存活:`search_tools` 传 `metadata_types` 时读到的是「去掉这两个参数 —— types / metadata_types」,而 `types` 正是它没有入参的那个。

改成按**调用方实际设过的**过滤器选文案:`types` 来自请求时才两个都点名;`types` 是入口固有的,就只点 `metadata_types`。`NoToolsOfRequestedKind` 的文案正好如此 —— 它在委托重构时变成了孤儿,现在归位。

**重复子测试** —— 也确认,和既有那条逐字相同。改成断言真正缺的那条性质:只传 `metadata_types` 时提示不得连带点名 `types`。

按上一轮的教训,这条断言我**撤掉修复验证过会红**:

```
不该连带让调用方去掉它设不了的 types: "该知识网络挂载了能力，但没有 types / metadata_types 指定类型的。去掉这两个参数可看到全部已挂载能力。"
```

恢复后全量绿。

/review

PR #1402, opened 2026-09-09
feat(bkn-backend): compile a read-only openCypher subset into one SQL statement

### LINKED ISSUE #1210: feat(ontology-query): Cypher 只读子集查询端点
> 方案已调整：编译目标从 ontology-query 自有遍历改为 **SQL 直接下推 Vega Raw Query**。完整技术设计见 [bkn-docs](https://github.com/openbkn-ai/bkn-docs/blob/docs/ontology-query-cypher-subset/docs/foundry/bkn/design/issue-1210-ontology-query-Cypher%E5%8F%AA%E8%AF%BB%E5%AD%90%E9%9B%86-%E6%8A%80%E6%9C%AF%E8%AE%BE%E8%AE%A1.md)（分支 `docs/ontology-query-cypher-subset`，待合并）。

## 目标

在 ontology-query 上提供只读 Cypher 查询端点，把 Cypher 子集编译成 **SQL 并下推 Vega Raw Query 执行**，让 join 发生在数据库里而非 Go 内存。不引入图存储。

命名前提：这不是 openCypher 实现，也不能这样命名。openCypher 的 Apache-2.0 归属声明要求，未经 openCypher Implementers Group 共识流程批准的衍生实现不得称为 "Cypher" 或 "openCypher"。端点名 `/cypher` 为占位。

## Module

ontology-query（新增调用方）、vega-backend（不改契约）

## 为什么走 Vega 直连

ontology-query 自有遍历（面 A）在 Go 侧逐跳查询 + 内存拼路径，下推面只有单表 `ViewQuery`，**无 join 无聚合**，且受 `PathQuotaManager` 全局 10000 条配额约束、超限静默截断。

Vega Raw Query（面 B，`POST /api/vega-backend/{v1,in/v1}/resources/query`）接受单条顶层 SELECT，**原生支持 JOIN / GROUP BY / ORDER BY / LIMIT / OFFSET / 派生表**，SQL 中用 `{{.resource_id}}` 占位引用资源，Vega 从 FROM 与 JOIN 中提取替换（`raw_query_service.go:759,841`），并用 sqlglot 转译到目标 catalog 方言。

走面 B 直接消掉面 A 的三个问题：跨节点谓词不再需要残余过滤、`NOT` 不需要德摩根律改写、路径配额静默截断不存在。

映射元数据建模层已齐：对象类 → `DataSource.ID`（占位符值）、属性 → `DataProperty.MappedField.Name`（物理列）、关系 → `MappingRules`（join 键对）。

## 已确认的限制（不做降级兜底）

- **只支持单 Catalog**。一条 SQL 涉及的资源必须同属一个 Catalog，否则 Vega 501 `MultiCatalogNotSupported`（`raw_query_service.go:771`）。跨源模式**编译期报错**，不透传 501。
- **不支持 OpenSearch**。SQL 面的方言映射 `targetDialectForCatalog`（`raw_query_service.go:706`）只覆盖 mysql / mariadb / postgresql / sqlserver。OpenSearch 虽在 `SupportedConnectorTypesForQuery`（`interfaces/connector_type.go:43`）内，但走 `dsl` 面无 join 能力，含此类对象类的模式一律报错。

**因此不设面 A 降级路径。** 编译不出 SQL 的一律报错——两条执行面语义与性能差异过大（数据库 join vs Go 内存拼接 + 10000 条配额静默截断），同一条 Cypher 在两条面上的行为用户无法预期。单一执行路径换可解释性。

## 缺口（按约束强度）

1. **`WITH` 被只读策略禁用**（`querypolicy/sqlglot_adapter.go` 的 `FORBIDDEN_NODE_NAMES`）⇒ 无递归 CTE。变长路径只能按长度枚举成多条定长 SQL，最短路无法实现，上界必须有限且需硬上限。
2. **`UNION`/`Intersect`/`Except` 同样被禁** ⇒ 省略标签展开、无向边、多关系类型三种展开无法合并进一条 SQL，退化成「发 N 条查询 + 客户端归并」。planner 必须编译期估算展开数并设阈值。
3. **授权粒度下移**。直连 Vega 后生效的是资源级授权，与 ontology-query 的对象类级授权粒度不同，需确认是否被绕过；若会，Cypher 端点须自行校验。
4. **sqlglot 是 python3 子进程**（`querypolicy/sqlglot_adapter.go:101`、`logics/query/sqlglot/sqlglot.go:46`）。一条 Cypher 展开成 N 条 SQL 即 N 组子进程调用，延迟与并发压力随展开数线性放大。
5. **branch 维度在 Vega 不存在**。BKN 对象类有 branch，Vega 资源没有，落法待定。
6. **两处适配器缺失**：`OntologyManagerAccess.ListObjectTypes`（label → ot_id 解析需要，bkn-backend `GET .../object-types` 已存在）、ontology-query 侧的 Vega Raw Query 适配器（现只有 `GET /resources/{id}/data`）。
7. **匿名函数一律被拒**。sqlglot 不认识的函数按 `exp.Anonymous` 拒绝，生成的 SQL 不能用非标准函数。

## 方案

新增 `server/logics/cypher/`：`parser.go`（openCypher 官方 `Cypher.g4` 经 antlr4-go 生成，版本钉死）、`binder.go`（label/rel type/属性 → id 与物理列，schema 快照缓存）、`planner.go`（模式分段、catalog 归属检查、展开数估算）、`sqlgen.go`（生成单条 SELECT，资源占位）、`executor.go`（提交 Raw Query、多查询归并、关系唯一性去重）。

路由外部内部双份：`POST /api/ontology-query/{v1,in/v1}/knowledge-networks/:kn_id/cypher`。

planner 在生成 SQL 前必须校验三条前置约束：全部对象类同 Catalog、该 Catalog 的 `ConnectorType` ∈ `{mysql, mariadb, postgresql, sqlserver}`、无逻辑属性引用。任一不满足直接报错。

## Phase 1 范围

支持：`MATCH` 多跳定长模式（标签 + 内联属性 + 单一关系类型 + 有向/无向）、`[:R*m..n]`（有限上界）、`WHERE`（含跨节点比较与 `NOT`）、`RETURN` 投影与 `AS`、`DISTINCT`、`ORDER BY`、`SKIP`、`LIMIT`、`$param`。

编译阶段拒绝：写子句、`UNION`、`[r {k:v}]` / `type(r)`、`[:A|B]`、`OPTIONAL MATCH`、`shortestPath`、`[*0..n]` 与无上界 `[*]`、跨 Catalog 模式、含 OpenSearch 对象类的多跳模式、逻辑属性引用、sqlglot 不认识的函数。

## 关系唯一性

Cypher 9 规定单模式内同一条边不重复匹配。边非实体无身份，在 SQL 里表达为相邻同类型边的 join 键不等式；无向边需端点规范化排序后比较。

## Acceptance Criteria

- [ ] 单跳端到端：`MATCH (o:Order)-[:PLACED_BY]->(c:Customer) WHERE o.amount > $min RETURN o.id, c.name ORDER BY o.amount DESC LIMIT 20` 生成一条合法 SELECT 并返回正确结果。
- [ ] 多跳 join：三节点模式生成单条两 join 的 SELECT，join 条件来自 `MappingRules`。
- [ ] 关系唯一性：`(a)-[:R]-()-[:R]-(b)` 不返回 `a = b` 的行。
- [ ] 列名映射：SQL 用 `MappedField.Name` 而非逻辑属性名。
- [ ] 跨 Catalog 模式在编译期报错，不把 Vega 501 透传给调用方。
- [ ] OpenSearch 支撑的对象类模式在编译期报错。
- [ ] 变长路径 `[:R*1..3]` 生成 3 条 SQL 并正确归并去重。
- [ ] 生成的 SQL 全部通过 Vega `ValidateSQL`，无 `With`/`Union`/匿名函数。
- [ ] 拒绝面每项各一条用例，断言报错且信息说明原因。
- [ ] 双路由（`/v1` 与 `/in/v1`）行为一致。
- [ ] CI green。

## 待确认

- 对象类级授权是否被 Vega 直连绕过；校验放 planner 还是 executor。
- 变长路径上界硬上限、展开查询数拒绝阈值取值。
- 多查询归并场景是否放弃游标分页只支持 `single`。
- branch 在 Vega 资源维度的语义。

## Boundaries / Risk

- 只读，不触碰写路径；不改建模 API 与现有查询契约；不改 Vega 契约，只作为 Raw Query 新调用方。
- 风险集中在展开数放大（省略标签 + 无向 + 变长叠加，乘以 sqlglot 子进程开销）与单 Catalog 边界对可用性的削减。

## Related

- #532 逐跳过滤条件——**不再是本 issue 的依赖**，面 A 不参与执行
- **完整技术设计**：[openbkn-ai/bkn-docs#96](https://github.com/openbkn-ai/bkn-docs/pull/96) → `docs/foundry/bkn/design/issue-1210-ontology-query-Cypher只读子集-技术设计.md`。含 14 节实测附录：只读策略形态判定、sqlglot 子进程实测开销、三方言族转义差分、活体 MariaDB/PostgreSQL/SQL Server 往返、端到端 JOIN 与聚合、真实 `mapping_rules` 与遍历引擎对答案、`filtered_cross_join` 与字段重命名的探针验证。
- #1297 Logic View 参数插值只转义单引号——**Vega 侧缺陷**，本设计自行转义，不受影响，非前置依赖
- #1298 SQL Server 目标缺 N 前缀——**Vega 侧缺陷**，本设计直出 tsql 并自行渲染 `N'...'`，不受影响，非前置依赖

sh00tg0a1:
# Pull Request

## What Changed

Compiles a read-only subset of openCypher 9 into a single SQL statement and runs it through vega-backend's Raw Query interface.

- [x] `logics/cypher` — parser (ANTLR, openCypher M19 grammar), analyzer, binder, planner, SQL generator
- [x] `logics/cypher/query_service.go` + `driveradapters/cypher_handler.go` — `POST /knowledge-networks/{kn_id}/cypher-queries` on both the external and internal faces
- [x] `logics/cypher/visibility.go` — per-object-type and per-relation-type authorization, applied while the schema is built
- [x] `interfaces` / `drivenadapters/vega_backend` — `VegaBackendAccess.RawQuery`, which fails closed when the caller carries no identity
- [x] `docs/api/bkn/cypher-queries.yaml` — API contract
- [x] `locale/cypher.{en-US,zh-CN}.toml`, `errors/cypher.go` — error codes and messages

---

## Why

- Background: Cypher is how people outside this project expect to ask a graph question. Everything the platform models already lives in relational resources behind vega-backend, so the way to answer such a question here is to compile it, not to build a graph engine. Design: `docs/foundry/bkn/design/issue-1210-BKN-Cypher只读子集-技术设计.md` in bkn-docs (PR openbkn-ai/bkn-docs#96).
- This supersedes #1325, which targeted `release/0.1.4` and was closed. The compiler is unchanged from that branch; the authorization model is not — see below.

---

## How

Key implementation points:

- **One Cypher, one SQL.** The compiler emits a single SELECT with `{{.resource_id}}` placeholders, which vega-backend replaces with quoted physical table names. No table name is ever composed here.
- **Read-only by construction, not by permission.** A write clause is refused in the analyzer, before anything is bound or generated.
- **Refusals name the construct.** Anything that parses as openCypher but is outside the subset comes back as `OPTIONAL MATCH is not supported`, with a line and column, rather than as a syntax error or as quietly different results. OR and NOT are refused for now because they change which rows a join produces, and a wrong answer there returns as data rather than as an error.
- **Authorization matches this branch's model, and is applied by filtering.** main authorizes a knowledge network's children individually, so `query_data` is checked on each object type and relation type — while the schema is being built, not after the query is compiled. A concept the caller may not query is therefore absent rather than refused, so the endpoint cannot be used to enumerate a model; a relation type is dropped when either endpoint is hidden, so a join cannot return rows of a hidden object type through the relation's own grant; and a caller who may read nothing gets 403. The network-wide grant is not checked separately because `query_data` on a child inherits from `query_data` on its parent network — checking it as well would have made a grant on a single object type worthless.
- **vega-backend still checks each resource** for `view_detail` under the caller's own identity, so a model that binds an object type to a resource the caller cannot read is stopped there.
- **Escaping lives in one place.** Literals are the only caller-controlled text that reaches the statement — identifiers come from the model, table aliases are generated — so escaping is one function per dialect, carrying rules that were round-tripped against a live MariaDB, including the trailing backslash that swallows a closing quote when it is missed.
- **Paging is asked for explicitly.** vega-backend wraps the statement and applies a page of its own that defaults to 20 rows, so the request repeats the statement's own LIMIT. A LIMIT above the ceiling is refused rather than clamped, and a query without one is capped at 1000.
- **The generated SQL is not returned.** It names physical tables and columns; being allowed to read the data is not being entitled to the schema behind it.

Accepted for now: one `MATCH` over a path of at most one relationship, each node with exactly one label and each relationship with exactly one type and a direction; `WHERE` of properties compared against literals, joined by `AND`; `RETURN` of property references with optional aliases, `DISTINCT`, `ORDER BY`, `SKIP`, `LIMIT`. Labels and relationship types accept either the id or the name.

Breaking changes: none. New endpoint, new package; the only change to existing code is one added method on `VegaBackendAccess` and its adapter.

---

## Testing

- [ ] Integration tests passed locally — not applicable, no integration suite for this path
- [ ] Verified in test environment — **not on this branch's authorization model; see below**

Test notes:

`go test ./...` in `adp/bkn/bkn-backend/server`: 45 packages pass.

The compiler was verified end to end on the test server (14.103.77.23) and on the development VM, against real knowledge networks — but **both run the 0.1.4 line, where knowledge-network children are not separate resources**. The compile, generation and execution path exercised there is byte-identical to this branch; the authorization rework in b357fb5 is covered by unit tests only, because a 0.1.4 bkn-safe has no `object_type` resources to filter on. Verifying it live needs a main-line environment.

What was verified live (0.1.4 line):

| Check | Result |
|---|---|
| Single-table projection, JOIN, WHERE, ORDER BY, DISTINCT, SKIP/LIMIT | Correct rows, 0.25–0.5 s |
| One-key JOIN generated from a relation type's `mapping_rules` | `item 20732 → order 18357`, matching what the existing traversal engine returns |
| Two-key JOIN (compound `ON`) | Same answer as the one-key join over the same data |
| Renamed columns (every property name differs from its physical column) | Correct; the physical name and the display name are both refused |
| Label and relationship type by name, including Chinese names | Correct |
| `filtered_cross_join` relation in a pattern | 400, named as such |
| Injection payloads (`a'; DROP TABLE referees --`, trailing backslash, four-byte emoji) | Treated as ordinary string literals; the table is intact |
| Query without LIMIT | 1000 rows; `LIMIT 2500` returns all 1248 rows, so vega-backend's 20-row default page is not silently truncating |
| No identity | 403 before any compilation |
| Rejection surface (write clauses, `count(*)`, `OR`, unknown label/property, relationship written backwards, LIMIT above ceiling) | 400, each naming the construct with its position |

A probe knowledge network (`cypher_probe_fcj`, three object types with every property renamed, two direct relations, one filtered cross join) was left on 14.103.77.23; it binds the ecommerce demo resources read-only and writes no business data. Delete with `DELETE /api/bkn-backend/in/v1/knowledge-networks/cypher_probe_fcj`.

---

## Risk & Rollback

Potential risks:

- **New read path to customer data.** Mitigated by per-object-type authorization here, by vega-backend's per-resource check under the caller's identity, by the read-only refusal happening before compilation, and by vega-backend's own sqlglot read-only policy. The statement is never assembled from caller text except through literal escaping.
- **The authorization rework has no live coverage yet.** It is the one part of this branch that behaves differently from what was tested on 0.1.4. A reviewer with a main-line environment should exercise a caller whose grant on one object type was revoked.
- **openCypher semantics that SQL does not share.** The subset is deliberately narrow for this reason; anything whose translation is not obvious is refused rather than approximated.
- **Unverified ground.** SQL Server targets still lose non-ASCII literals through vega-backend's own rendering (#1298, not this branch); the tsql dialect is therefore not generated. Postgres generation exists but is not submitted yet — MySQL is the input dialect, matching the platform's own SQL tool.

Rollback plan: the endpoint is additive. Revert this branch, or drop the two route registrations in `routers.go` to take it out of service while keeping the package.

---

## Additional Notes

- The design document proposed the path `/cypher`; the implementation uses `/cypher-queries`, matching `/metrics` and `/object-types`. The design document should follow, or the code should change — reviewer's call.
- The ANTLR runtime (`github.com/antlr4-go/antlr/v4`) was already a dependency of vega-backend at the same version; `logics/cypher/parsing` is generated and carries the grammar's Apache-2.0 header, with `parser/generate.sh` recording how to regenerate it.

Closes #1210


sh00tg0a1:
## Test report

Numbers below are from `go test` on this branch at f41fe0c, run as CI runs it (`I18N_MODE_UT=true go test -count=1 -gcflags="all=-l" ./...` in `adp/bkn/bkn-backend/server`). **45 packages pass, 0 failures.**

### Coverage of the code this PR adds

| Package / file | Statement coverage |
|---|---|
| `logics/cypher` (whole package) | **92.3%** |
| ├ `parse.go` | 100.0% |
| ├ `sqlgen.go` | 99.4% |
| ├ `visibility.go` | 97.0% |
| ├ `binder.go` | 94.2% |
| ├ `analyze.go` | 93.0% |
| ├ `plan.go` | 87.0% |
| ├ `schema_source.go` | 86.9% |
| ├ `query_service.go` | 77.8% |
| └ `ast.go` | 75.0% |
| `drivenadapters/vega_backend` → `RawQuery` | **100.0%** |
| `drivenadapters/vega_backend` → `buildUserHeaders` | **100.0%** |
| `driveradapters/cypher_handler.go` → `RunCypherQuery` | **100.0%** |
| `driveradapters/cypher_handler.go` → `replyHandlerError` | **100.0%** |

`logics/cypher/parsing` is excluded: it is ANTLR output, regenerated by `parser/generate.sh`, and covering generated code measures the generator rather than this change.

### What the tests actually assert

183 test cases across the three packages (162 in `logics/cypher`, 12 for the handler, 9 for the vega call). Grouped by what each set is there to prove:

**The subset is what we say it is** — 19 parse cases (in-subset and out-of-subset syntax both parse), 47 rejection cases, each asserting the message names the construct: `OPTIONAL MATCH`, `UNION`, `WITH`, `UNWIND`, procedure calls, write clauses, undirected and variable-length relationships, multi-hop, `RETURN *`, aggregates, arithmetic, `OR`/`XOR`/`NOT`, `IN`, `IS NULL`, `STARTS WITH`, chained comparisons, list and map literals, pattern predicates, `EXISTS` subqueries. A rejection also carries a line and column, which is asserted separately.

**Compilation produces the SQL we intend** — exact-string assertions on the generated statement for single-table projection, one-key and two-key joins, a relationship read backwards, `DISTINCT`, `ORDER BY`, `SKIP`/`LIMIT`, the injected default limit, and the PostgreSQL dialect. Not "contains", full equality: a change in what we emit has to be deliberate.

**Escaping** — the payloads that were round-tripped against a live MariaDB are asserted at the unit level too: `O'Brien`, `a'; DROP TABLE x --`, a trailing backslash, an escaped-quote injection, a backtick, a quote inside an identifier. Plus every Cypher escape (`\b \f \n \r \t \" \'`), both unicode forms, and the five malformed forms the decoder must refuse.

**Authorization** — object types are filtered by `query_data` while the schema is built; an unreadable object type resolves as an unknown label rather than a refusal; a network where nothing is readable answers 403; an empty network answers "unknown label" instead; the operation asked of bkn-safe is asserted to be `query_data` and the resource id to be the `knID/childID` form.

**Nothing leaks** — the generated SQL is absent from the response; the dependency's message is absent from the error on all three failure modes of the vega call (transport, error status, unparseable body); the identity refusal fires before the network is touched.

**The endpoint behaves** — both faces reachable, branch reaches the service, the caller is in the context before the service runs, non-JSON body and content type refused, missing network is 404, a service refusal keeps its status, a plain error becomes 500 rather than a panic.

### What is not covered, and why

- **Constructors and wiring** (`NewCypherQueryService`, 0%): they assemble singletons from `common.AppSetting`; a test would assert the assignment statements it duplicates.
- **`PlanError.Unwrap`** (0%): present for `errors.Is`/`As` callers that do not exist yet.
- **Defensive branches** whose inputs the earlier stages make impossible — a null literal reaching generation, `SKIP` without a `LIMIT`. These are covered by calling the stage directly, not through a query, because a query cannot produce them; that is why the file numbers are 99.4% rather than 100%.
- **Malformed escapes** are likewise tested against `decodeStringLiteral` directly: the lexer refuses them first, so a query-level test would have asserted the lexer while appearing to cover the decoder.
- **The authorization model, live.** This is the one real gap. The compile and execute path was verified end to end on 14.103.77.23 and on the development VM, but both run the 0.1.4 line, where knowledge-network children are not separate resources in bkn-safe — there is nothing there to filter on. The filtering in `visibility.go` therefore has unit coverage only. Verifying it needs a main-line environment and a caller whose grant on one object type has been revoked.

sh00tg0a1:
## Live language-coverage run

**107 cases, 107 pass**, executed against the deployed service on the test server (14.103.77.23), not in a unit test: every case is an HTTP request to `POST /api/bkn-backend/in/v1/knowledge-networks/{kn}/cypher-queries` whose answer came back through the compiler, vega-backend and a real PostgreSQL.

### Environment

The test server had moved to the main line on 2026-09-08 — it now runs `0.1.5-main.20260908231848.sha83f2b88`, six commits behind this branch's base, with no bkn-backend change in between. This branch was therefore built and deployed there directly, which also closes the authorization gap reported earlier: the run below exercises the per-object-type check on the model that check was written for.

Queries run against the probe network `cypher_probe_fcj`: three object types whose every property name differs from its physical column, a one-key direct relation, a two-key direct relation, and a filtered cross join.

### Accepted surface — 38 cases, all returning rows

| Group | Cases | What is covered |
|---|---|---|
| projection | 4 | aliased and unaliased columns, several properties, `DISTINCT` |
| clauses | 6 | `ORDER BY` ascending, descending and multi-key; `SKIP`/`LIMIT`; `DISTINCT` with a sorted returned column |
| operators | 7 | `=`, `<>`, `<`, `>`, `<=`, `>=`, and a three-term `AND` |
| literals | 13 | single- and double-quoted strings, integer, negative, float, hex `0x`, octal `0o`, `\t`, four- and eight-digit unicode escapes (including an emoji), the escaped-quote injection payload, a trailing backslash |
| patterns | 9 | outgoing and incoming relationships, anonymous node, parenthesized pattern, two-key join, label and relationship type by Chinese name in backticks, lower-case keywords, multi-line query |

Median round trip 2.25 s, worst 3.2 s, including the Context Loader hop and vega's translation subprocesses.

### Refused surface — 69 cases, each 400 naming the construct

| Group | Cases | Examples |
|---|---|---|
| reject-clause | 8 | `OPTIONAL MATCH`, `UNION`, `WITH`, `UNWIND`, `CALL`, two `MATCH`es, comma-separated patterns, path variable |
| reject-write | 5 | `CREATE`, `SET`, `DETACH DELETE`, `REMOVE`, `MERGE` — all refused as read-only before anything is bound |
| reject-pattern | 11 | no label, two labels, inline property maps on node and relationship, undirected, both arrows, untyped relationship, relationship variable, variable length, alternative types, multi-hop |
| reject-expr | 11 | `RETURN *`, bare variable, `count(*)`, function call, arithmetic, parameter, list index, nested property, bare literal, `CASE`, list comprehension |
| reject-where | 15 | `OR`, `XOR`, `NOT`, `IN`, `IS NULL`, `STARTS WITH`, `CONTAINS`, chained comparison, non-comparison predicate, list literal, `null` comparison, pattern predicate, `EXISTS` subquery, parenthesized expression, property-to-property comparison |
| reject-paging | 5 | arithmetic in `LIMIT`, non-integer `LIMIT`, negative `SKIP`, `LIMIT` above the ceiling, `DISTINCT` sorted by an unreturned value |
| reject-literal | 2 | unicode escape past the last code point, unpaired surrogate |
| reject-model | 11 | unknown label, unknown property, undefined variable, reused variable, duplicate output name, relationship written backwards, filtered cross join, unknown relationship type, display name used as a property, physical column used as a property, syntax error |

Each case asserts the status *and* that the message names the construct, so a refusal that started arriving for the wrong reason would fail the run.

### Authorization, live on the main line

| Caller | Result |
|---|---|
| Network-wide `query_data` grant | 200 — the child check passes by inheritance, which is the behaviour this branch relies on rather than checking the network separately |
| Account holding no grant | 403 |
| No identity at all | 403, before anything is compiled |

Still unverified: a caller whose grant on **one** object type has been revoked, which should make that object type read as an unknown label while the rest of the network stays queryable. That needs a grant edit in bkn-safe.

### One correction to our own expectations

The octal literal case was written as `025126` and failed. openCypher M19 spells octal `0o25126`, and with the prefix it returns the right row (`0o25126` = 10838). The grammar was right and the test was wrong; the case now asserts the language as specified.

sh00tg0a1:
## Grammar coverage

The earlier run showed 107 cases passing, which says the cases pass — not how much of the language they touch. So the corpus was measured against the grammar itself: parse every query, walk the tree, record which of the 103 rules in `Cypher.g4` were entered.

**103 of 103 openCypher M19 grammar rules reached — 100%.** 116 live cases, all passing.

The first measurement was 92.2% (95/103). The eight rules nothing reached, and the case added for each:

| Rule | Construct it covers | Case added | Answer |
|---|---|---|---|
| `oC_StandaloneCall` | `CALL proc()` as the whole query | `CALL db.labels()` | 400 procedure calls |
| `oC_ImplicitProcedureInvocation` | `CALL proc` with no parentheses | `CALL db.labels` | 400 procedure calls |
| `oC_ProcedureResultField` | `YIELD field AS alias` | `CALL db.labels() YIELD label AS l RETURN l` | 400 procedure calls |
| `oC_MergeAction` | `ON CREATE` / `ON MATCH` | `MERGE (o:probe_order) ON CREATE SET o.o_state = 'x' RETURN o.o_key` | 400 read-only |
| `oC_Quantifier` | `all` / `any` / `none` / `single` | `WHERE any(x IN [1] WHERE x = 1)` | 400 quantified expressions |
| `oC_PatternComprehension` | `[(a)-[:R]->(b) \| b.x]` | `RETURN [(o)-[:probe_direct]->(i:probe_item) \| i.i_key]` | 400 pattern comprehensions |
| `oC_ReservedWord` | a reserved word used as a schema name | `MATCH (o:ALL) RETURN o.o_key` | 400 unknown label |
| `oC_BooleanLiteral` | `true` / `false` | `RETURN o.o_key LIMIT true`, `LIMIT false` | 400 non-integer LIMIT |

Two queries in the corpus reach no rule at all, on purpose: they are the malformed ones, and they exercise the syntax-error path instead.

### The corpus is now part of the repository

`logics/cypher/testdata/grammar_corpus.cypher` holds all 116 queries, and two tests state what they buy:

- **every grammar rule is reached** — a rule nothing reaches is a construct nobody has ever handed the compiler, which is where an unhandled shape hides. If a future grammar bump adds a rule, this fails until the corpus covers it.
- **every query is answered or refused by name** — each one comes back as a compiled query, or as a refusal carrying both the construct and its position. Never a panic, and never an error of a type the service cannot turn into a status.

The file is written against the probe network the live runs use, so it can be replayed against a deployment rather than only run in-process — which is how it was checked before being committed.

### One correction

The octal case was written `025126` and failed. openCypher M19 spells octal `0o25126`; with the prefix it returns the right row (`0o25126` = 10838). The grammar was right and the test was wrong.

PR #1404, opened 2026-09-09
ci(build): build each architecture on a native runner instead of QEMU

sh00tg0a1:
## Why

`reusable-build.yml` ran `platforms: linux/amd64,linux/arm64` in one buildx invocation on an amd64 runner, so the arm64 leg was emulated under QEMU. None of the Dockerfiles cross-compile (no `--platform=$BUILDPLATFORM` builder stage), which put `go build` with CGO and every `pip install` under emulation. Latest `release-adp-bkn-backend` on main: the `Build and push image` step took 16 minutes (15:19:49 → 15:35:49), almost all of it in the arm64 leg. Python-based images (model-factory-base, bkn-agent, sandbox) pay the same tax.

## What

Same shape bkn-studio already ships in `release-studio.yml`:

- `setup` job maps the existing `platforms` input to a matrix (`linux/amd64` → `ubuntu-latest`, `linux/arm64` → `ubuntu-24.04-arm`, GitHub-hosted, free for public repos). Unknown platforms fail loud instead of silently falling back to emulation.
- `build` matrix: one leg per architecture, no `setup-qemu-action`. Each leg pushes `:<version>-<arch>` to GHCR and (when credentials exist) SWR, and records the image digest as an artifact.
- `manifest` job unions the legs under the plain `:<version>` tag with `docker manifest create` + `push`, addressing children by the recorded digests (`repo@sha256:…`) rather than re-resolving the arch tags — versions are not unique per run (`X.Y.Z-release`, rebuilt `v*` tags) and no caller sets `concurrency`, so tag lookup could pair legs from two interleaved runs. `docker manifest` emits the legacy Docker manifest list media type, which Huawei SWR requires; `provenance: false` + `oci-mediatypes=false` on the legs keep the children in Docker schema2 as before.
- Verify-only runs (`publish: false`) build both legs and skip the manifest job.

Caller interface is unchanged; all 17 `release-*` workflows pick this up without edits.

## Registry shape

- `:<version>` stays a manifest list — pulls, charts, digest pinning, `skopeo copy --all` mirroring all behave as before.
- New per version: `:<version>-amd64` and `:<version>-arm64`. They are the list's children and must stay as long as the list does. They also let air-gapped users `docker pull` one architecture without `--platform` or skopeo — documented in `deploy/README*.md` with the pull/tag/save/import sequence.
- Known residual: if one leg fails after the other already pushed its arch tag, `:<version>` keeps pointing at the previous list and that arch tag is ahead of it until the version builds cleanly again. Noted in the workflow header.
- `automation-ghcr-cleanup.yml` is not touched. It has failed on every scheduled run since at least 2026-06-21 (`Unable to resolve action snok/container-retention-policy@v3` — only `v3.1.0` exists), so no retention runs today. Once repaired to v3.1.x, its multi-arch protection is digest-based and covers tagged children too (`filter_out_multi_arch_children.rs` filters both `tagged` and `untagged` candidates), so the arch tags of protected versions need no extra rule. Repairing the workflow itself (action ref, and its `image-tags` patterns are globs, not the regexes they are written as) is a separate issue.

## Verification

Every branch push triggers all `release-*` workflows with `publish: true` (they watch `reusable-*.yml`), so the branch builds exercise the new path end to end against GHCR and SWR.

- First wave (tag-addressed list): 17/17 release workflows green. `docker manifest inspect` on both GHCR and SWR: `application/vnd.docker.distribution.manifest.list.v2+json` with two `manifest.v2+json` children (linux/amd64, linux/arm64), identical list digest on both registries. GHCR registry API serves `:<version>` as the list and `:<version>-arm64` as a plain manifest.
- Verify-only path: `workflow_dispatch` with `publish=false` on release-deploy-redis (run 34306687678) and release-infra-oss-gateway (run 34306791501): both legs build, `manifest` skipped, downstream `chart / package` runs, run conclusion success.
- Timings vs main: bkn-backend 16 min → amd64 100s / arm64 83s + 23s manifest; model-factory-base ~10 min → ~2.6 min; bkn-agent ~7 min → ~1.4 min. arm64 runner queue wait across 17 concurrent jobs: median 5s, max 6s.
- Second wave (digest-addressed list) results are in the checks on the latest commit.

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -218,7 +218,11 @@ jobs:
 
       # The manifest job pins the list to these digests (see header). One file
       # per leg, named by arch; the artifact name carries the image so several
-      # reusable-build calls in one workflow run don't collide.
+      # reusable-build calls in one workflow run don't collide. `__` is the
+      # delimiter because image names use `-` and can be prefixes of each other
+      # (vega-backend / vega-backend-python-base in one run): the manifest job
+      # matches `digest__<image>__*`, and `digest__vega-backend__*` cannot
+      # match `digest__vega-backend-python-base__amd64`.
       - name: Record image digest
         if: ${{ inputs.publish }}
         shell: bash
@@ -239,7 +243,7 @@ jobs:
         if: ${{ inputs.publish }}
         uses: actions/upload-artifact@v4
         with:
-          name: digest-${{ inputs.image-name }}-${{ matrix.arch }}
+          name: digest__${{ inputs.image-name }}__${{ matrix.arch }}
           path: ${{ runner.temp }}/digests/*
           if-no-files-found: error
           retention-days: 1
@@ -254,12 +258,15 @@ jobs:
       contents: read
       packages: write
     steps:
+      # No merge-multiple: each artifact unpacks into its own directory named
+      # after the artifact, and the union step reads the exact
+      # `digest__<image>__<arch>/<arch>` path. Even if the pattern ever
+      # over-matched, another image's digest could not be read by mistake.
       - name: Download image digests
         uses: actions/download-artifact@v4
         with:
-          pattern: digest-${{ inputs.image-name }}-*
+          pattern: digest__${{ inputs.image-name }}__*
           path: ${{ runner.temp }}/digests
-          merge-multiple: true
 
       - name: Login to GHCR
         uses: docker/login-action@v3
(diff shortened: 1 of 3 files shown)


sh00tg0a1:
Fixed in the latest commit (`digest__<image>__<arch>` artifact names, no `merge-multiple`, exact-path read with sha256 validation). `vega-backend` / `vega-backend-python-base` is the only prefix pair across all `release-*` callers; a `workflow_dispatch` of release-adp-vega-backend with `force-build-python-base=true` (run 34318130894) exercises both images in one run — results below once it finishes.

On the SWR note: the digest-addressed union already ran against SWR on the previous commit — run 34307631980 (`release-deploy-redis`), `build / manifest` job: `Created manifest list swr.cn-east-3.myhuaweicloud.com/openbkn-ai/redis:…` then `Pushed multi-arch …`, with `docker manifest inspect` on the SWR side listing the same two child digests the legs recorded (`sha256:954add…` amd64, `sha256:931550…` arm64). All 17 release workflows on that commit went through the same path.

/review

PR #1408, opened 2026-09-09
feat(bkn-backend): conditions, parameters and aggregation for the Cypher subset

sh00tg0a1:
# Pull Request

> **Stacked on #1402.** That PR carries the compiler and the endpoint; this one carries the three constructs a reader reaches for first. Merge #1402 before this.

## What Changed

- [x] `WHERE` is read as a tree: `OR`, `NOT`, `IN`, `IS NULL` / `IS NOT NULL`, and parentheses to group them
- [x] Parameters: `$name` in the query, values in a `parameters` object on the request
- [x] Aggregation: `count`, `sum`, `avg`, `min`, `max`, their `DISTINCT` forms and `count(*)`, with grouping derived the way Cypher derives it
- [x] `ORDER BY` accepts the name of a returned column and an aggregate written out again
- [x] `docs/api/bkn/cypher-queries.yaml` and the grammar corpus updated

---

## Why

- Background: with #1402 the endpoint answers "filtered, sorted single-hop reads". The first three things anyone asks of a query language after that are a count, an `OR`, and somewhere to put their values. Until they exist, a caller who needs any of them has to leave for the SQL tool, which is the thing this interface was meant to spare them.

---

## How

**Conditions.** The old shape held a flat conjunction, which is all `AND` needs. A tree replaces it, and the four constructs land on it. Parentheses had to come too: without grouping, `AND` and `OR` could only ever be written in their default precedence, which is a different query from the one someone meant. They group conditions in a `WHERE` and stay refused where there is nothing to group.

`XOR` stays refused: MySQL has the operator, PostgreSQL does not, and the rewrites disagree about a null operand. An empty `IN` list is written out as a constant, because Cypher says it matches nothing and SQL has no empty `IN` to say that with.

**Parameters** are bound to literals in the planner and escaped by the rules a written literal gets, so a parameter can change which rows come back and never which resource or column is read. Numbers decode as `json.Number`: an identifier that does not survive a float would otherwise be compared against the wrong value.

**Aggregation.** Cypher has no `GROUP BY` — aggregating anything groups by everything else returned — so the planner derives the grouping from the projection. That is also what stops a statement from grouping by something the caller never sees. Aggregating every column groups by nothing, because the answer is one row. The function name in the generated SQL comes from a fixed table, never from the query.

`ORDER BY` grew the two forms a grouped result needs. Sorting by a name is checked against what the query returns rather than left to the database; sorting a grouped or `DISTINCT` result by a value that survived neither is refused with the key named — the rule `DISTINCT` already had, now stated once for both.

Breaking changes: none. `parameters` is a new optional field; every query that compiled before compiles the same way.

---

## Testing

- [x] Unit tests passed locally — 45 packages, `logics/cypher` at 90.4% statement coverage
- [ ] Integration tests passed locally — not applicable

**147 live cases, 147 pass**, run against the deployment on 14.103.77.23 (main line) — every one an HTTP request answered through the compiler, vega-backend and a real PostgreSQL.

| Group | Cases | | Group | Cases |
|---|---|---|---|---|
| predicate | 11 | | reject-where | 13 |
| parameter | 6 | | reject-parameter | 4 |
| aggregate | 9 | | reject-aggregate | 5 |
| projection, clauses, operators, literals, patterns | 39 | | the other refusals | 60 |

The aggregate answers were checked for arithmetic, not only for status:

| Check | Result |
|---|---|
| `count(*)` over the probe network | 15000 |
| `count(*)` grouped by status | 7 groups summing to exactly 15000 |
| `count(DISTINCT o.o_state)` | 7 — the same 7 |
| `WHERE o.o_state = $s` with `s = "refunding"` | 286, matching that group's row |
| `WHERE k = 10774 OR k = 10963` vs `k IN [10774, 10963]` | both 2 |
| `WHERE k IN []` | 0 |
| Injection payload passed as a parameter | treated as a string; the table is intact |

The grammar corpus in `logics/cypher/testdata/` was updated as constructs changed sides, so the test that every grammar rule is reached still holds.

---

## Risk & Rollback

- **`OR` and `NOT` change which rows come back.** They were refused in #1402 for exactly that reason: a wrong answer arrives as data rather than as an error. What makes them safe now is that the generated condition is asserted as a whole string against the query that produced it, including where the parentheses fall, so a change in precedence fails a test rather than a report.
- **Aggregation moves work into the database.** A `GROUP BY` over a large resource is a heavier statement than a filtered read. The default limit still caps the rows returned, but not the work done to produce them.
- **Parameters are a new input path.** They are values only, bound and escaped like written literals; the tests include the payloads that were round-tripped against a live MariaDB, passed as parameters this time.

Rollback: revert this branch. #1402 stands on its own and the endpoint keeps working without these constructs.

---

## Additional Notes

Still refused, and deliberately: `WITH`, variable-length paths, `UNION`, multi-hop patterns, `OPTIONAL MATCH`, undirected relationships, string predicates (`STARTS WITH` and friends), and general functions. The first three need subqueries or recursion, which vega-backend's SQL tool does not promise; the rest are next in line rather than out of scope. Multi-hop is the largest of them and wants a decision on join order before it is written.


sh00tg0a1:
## Coverage report

Numbers from this branch at 1f26a04, run the way CI runs it (`I18N_MODE_UT=true go test -count=1 -gcflags="all=-l" ./...` in `adp/bkn/bkn-backend/server`). **45 packages pass, 0 failures.**

### Statements

| Package / file | Coverage |
|---|---|
| `logics/cypher` (whole package) | **91.1%** |
| ├ `parse.go` | 100.0% |
| ├ `sqlgen.go` | 99.0% |
| ├ `visibility.go` | 97.0% |
| ├ `parameters.go` | 95.5% |
| ├ `binder.go` | 94.2% |
| ├ `analyze.go` | 92.6% |
| ├ `plan.go` | 91.7% |
| ├ `schema_source.go` | 86.9% |
| ├ `query_service.go` | 77.8% |
| └ `ast.go` | 62.5% |
| `drivenadapters/vega_backend` → `RawQuery`, `buildUserHeaders` | 100.0% |
| `driveradapters/cypher_handler.go` → `RunCypherQuery`, `replyHandlerError` | 100.0% |

File percentages above are the mean over that file's functions, with one category left out: the ten one-line marker methods (`predicatePosition`, `planPredicate`) that exist only to close the two interfaces. They have no body to exercise, and calling them from a test would raise the number without testing anything. They are included in the 91.1% package figure, which is why it sits below the file means.

`logics/cypher/parsing` is excluded throughout: it is ANTLR output, regenerated by `parser/generate.sh`, and covering it would measure the generator.

### What the tests assert

365 test cases in `logics/cypher` across 56 top-level tests, plus 147 cases run against the deployment. Grouped by what each set is for:

**Conditions** — the generated `WHERE` is asserted as a whole string, including where the parentheses fall: `OR`, `AND` binding tighter than `OR`, `NOT`, `NOT` over a group, a double negation cancelling, `IS NULL` / `IS NOT NULL`, `IN` over strings and numbers, `IN []` and its negation, and a mixed tree. Precedence is the thing being pinned: a query whose meaning shifted would fail as a diff, not as a wrong report.

**Parameters** — bound values reach the statement through the same escaping as written literals, asserted with the payloads that were round-tripped against a live MariaDB (`a'; DROP TABLE x --`, a trailing backslash). Numbers: an integer, an integer that arrived as a float, a float, a `json.Number` at `int64` maximum, one past `int64` falling back to a float, and a malformed one refused. Refusals: not supplied, null, an unsupported type, positional (`$0`), negated, and in place of a label.

**Aggregation** — whole-statement assertions for `count(*)`, `count(x)`, `count(DISTINCT x)`, several groups with several aggregates, everything aggregated (no `GROUP BY`), an aggregate over a join, sorting by the output name, sorting by the aggregate written again, an unaliased aggregate keeping the name the author wrote, and filtering before grouping. Refusals: sorting a grouped result by an unreturned value, sorting by a name the query does not return, an aggregate over a literal, over an unknown property, a non-aggregate function, and an aggregate over two arguments.

**Grammar reach** — the corpus in `logics/cypher/testdata/grammar_corpus.cypher` is now 143 queries and still reaches all 103 rules of openCypher M19. Every one of them comes back as a compiled query or as a refusal naming the construct and its position — never a panic, never an error type the service cannot turn into a status.

**Live** — 147 cases against 14.103.77.23, all passing, with the aggregate answers checked for arithmetic rather than status: `count(*)` = 15000, the same 15000 split across 7 groups that sum back to it, `count(DISTINCT)` agreeing at 7, a parameterised filter agreeing with its group's row at 286, `OR` and `IN` agreeing at 2, and `IN []` at 0.

### What is not covered

- **Constructors and wiring** (`NewCypherQueryService`, 0%) — they assemble singletons from `common.AppSetting`.
- **`PlanError.Unwrap`** (0%) — present for `errors.Is`/`As` callers that do not exist yet.
- **`ast.go` at 62.5%** — two `describe` helpers that only appear inside error messages for operand shapes the analyzer refuses earlier.
- **Defensive branches** reached by calling a stage directly rather than through a query, because a query cannot produce them: a null literal arriving at generation, `SKIP` without a `LIMIT`, a malformed escape the lexer refuses first.
- **The per-object-type authorization, live.** Unchanged from #1402: the two ends are verified on the main line — a network-wide grant passes, an account with no grant gets 403 — but a caller whose grant on *one* object type was revoked still needs a grant edit in bkn-safe to test.

sh00tg0a1:
`1a350da3` is fixed, in 113c66d7 — the same commit this round's own summary describes. It carries over because the verification stage produced no conclusion (the `verify` job is red with every step green, so nothing adjudicated the item), not because the finding still stands.

The finding was right, and it was fixed at the root rather than at the branch:

- The grammar allows a pair of backticks with nothing between them, so an empty name can be written wherever a name can. `namedIdentifier` now refuses one for a variable, label, property, column name, relationship type and parameter, which closes the `ORDER BY ``` panic and the `AS ``` empty identifier together.
- `bareVariable` no longer reports an empty name as a reference, so a sort key cannot come out with neither an alias nor a property.
- `planSortKey` switches on `key.Property != nil` and names its last case, so a form added later fails as an error instead of dereferencing whichever branch it fell into.

Regression tests: `TestAnalyzeRejectsEmptyNames` covers all seven positions, and the corpus in `logics/cypher/testdata/grammar_corpus.cypher` carries `ORDER BY ``` and `AS ``` so the panic path stays exercised.

The two non-blocking items from the same round are also fixed, in the same commit: the ordering guard now tests `aggregating()` rather than a non-empty `GroupBy`, which covers `RETURN count(*) AS n ORDER BY o.amount` (every column aggregated, so no GROUP BY is derived), and sorting by an aggregate the projection does not have is refused rather than silently collapsing the result.

/review

PR #1447, opened 2026-09-10
feat(bkn-backend): let one MATCH describe a shape, not only a chain (0.1.4-yf)

sh00tg0a1:
# Pull Request

## What Changed

One MATCH may now describe a shape rather than only a chain.

- [x] Comma-separated paths in one MATCH: `MATCH (a)-[:R]->(b), (a)-[:S]->(c)`
- [x] Several MATCH clauses, which Cypher gives the same meaning
- [x] A variable written twice is the same node, so two paths meet at it
- [x] Paths that share no variable become a `CROSS JOIN`
- [x] `docs/api/bkn/cypher-queries.yaml` and the grammar corpus

---

## Why

Asking about one thing from two sides is the most ordinary thing a reader does, and it was refused:

```
MATCH (pa:player_appearances)-[:rel_player_appearances_team]->(t:teams)
MATCH (pa)-[:rel_player_appearances_player]->(p:players)
WHERE t.team_code = 'CHN'
RETURN p.family_name AS f
```

That came back as `multiple reading clauses is not supported`. So did the comma form. Both describe one node with two relationships hanging off it — a shape a chain cannot hold, which is exactly why it was refused rather than compiled wrong.

---

## How

The pattern was a list where edge *i* joined node *i* to node *i+1*. It is now a graph: an edge carries the two nodes it connects, and the analyzer keeps a variable-to-node map for the whole MATCH — so a variable written twice means the node it meant the first time. **That is the part that makes two paths one shape**: the second mention joins to the first instead of reading the table again. A first mention still has to name its object type; a label written again on a known variable must agree with the one it was declared with.

Generation follows. Tables come out in an order where each one joins to a table already in the statement, and anything left over is written as `CROSS JOIN` — which is what Cypher means by unconnected patterns, and writing it out rather than leaving it implicit is what lets a reader see the cost. A join whose two tables are both in the statement already still contributes its condition, which a naive ordering would have silently dropped.

The size cap moves with the shape: it counted relationships in one path, and now counts them across the pattern.

Breaking changes: none — three shapes that were refused now compile.

---

## Testing

- [x] Unit tests passed locally — 41 packages, no failures

Live on the development VM, which runs this line, against `worldcup_vega_catalog_bkn`:

| Query | Result |
|---|---|
| The two-MATCH query above | 200 — 89 distinct players over 408 appearances |
| `MATCH (g:goals)-[:rel_goals_player]->(p:players), (g)-[:rel_goals_team]->(t:teams)` | 200 — 53 goals |
| `MATCH (t:teams), (m:matches)` (no shared variable) | 200 — cartesian product, `CROSS JOIN` in the statement |
| The same shape written as one chain | 200 — identical answer |

New unit tests assert that all three spellings of the same shape — comma-separated, two MATCH clauses, one chain — **compile to the same SQL string**, and cover the branching shape, the cartesian product, a second mention carrying its own conditions without reading the table twice, and three refusals (a first mention with no label, a second mention contradicting the first, `OPTIONAL MATCH` among them).

### One thing found while testing, unrelated to this change

The `players` object type in that network was bound to a Vega resource that no longer existed, so anything touching it returned 500. The resource had been deleted from the catalog; re-running discovery brought it back under a new id and the object type was rebound. No code involved — but it is worth noting that a dead binding surfaces as a 500 rather than as a named 400, because the dependency's error is deliberately not passed through. Reporting it at compile time is a separate change.

---

## Risk & Rollback

- **A cartesian product is easy to write by accident.** `MATCH (a:X), (b:Y)` with no shared variable multiplies the two tables. The row limit caps what comes back, not what it costs to produce; the `CROSS JOIN` is written out so it is visible in the statement rather than implied.
- **Join ordering is now derived rather than positional.** The tests assert whole statements, so a change in the order or a dropped condition fails as a diff.

Rollback: revert this branch. The line keeps everything #1439 brought and the endpoint works without these shapes.


sh00tg0a1:
Both blockers reproduced exactly as described, and both are fixed in 2c006fb5.

`bb9615ed` — a relationship whose two ends are already in the statement attaches no table but still carries a condition, and that condition was appended to the FROM list with `AND`. That only works when the list happens to end in an ON clause. Reproduced:

- `MATCH (a:Order)-[:FOLLOWS]->(a) RETURN a.id` → `FROM {{.res_order}} t0 AND t0.f_id = t0.f_prev`
- a cycle beside an unconnected part → `CROSS JOIN {{.res_customer}} t3 AND t0.f_id = t2.f_prev`, which the review predicted and which is the worse of the two

A condition that attaches nothing belongs in WHERE, where it can stand on its own. The generator now collects those joins while ordering the FROM list and writes them as conditions:

- self-relationship → `FROM {{.res_order}} t0 WHERE t0.f_id = t0.f_prev`
- cycle plus cross join → `... CROSS JOIN {{.res_customer}} t3 WHERE t0.f_id = t2.f_prev AND (…)`

`d7bc543b` — two hops of one relation type between the same pair of nodes are the same relationship by construction, so `differentEdges` had nothing to compare and returned an empty negation: `WHERE NOT ()`. What the rule is asking for there is something nothing satisfies, so the plan now says that directly (`PlanNever`, generated as `1 = 0`) instead of leaving an empty operator behind. `MATCH (a:Order)-[:FOLLOWS]->(b:Order), (a)-[:FOLLOWS]->(b)` now ends in `AND 1 = 0`.

The review's framing was right and worth keeping: the analyzer accepted shapes the generator and the distinctness rule could not handle. Regression tests cover all three statements — self-relationship, cycle closing on tables already read, and a deferred condition beside a cross join — plus the empty-negation case, and the corpus gained the two shapes so they stay exercised.

/review

PR #1451, opened 2026-09-10
feat(bkn-backend): let one MATCH describe a shape, not only a chain

sh00tg0a1:
# Pull Request

## What Changed

Ports the two batches this line is missing, both already merged on `release/0.1.4-yf` (#1447):

- [x] Comma-separated paths in one MATCH, and several MATCH clauses, which Cypher gives the same meaning
- [x] A variable written twice is the same node, so two paths meet at it; paths sharing no variable become a `CROSS JOIN`
- [x] The two defects review found on the yf side, fixed here in the same shape
- [x] `docs/api/bkn/cypher-queries.yaml` and the grammar corpus

---

## Why

- Asking about one thing from two sides was refused on this line:

```
MATCH (pa:player_appearances)-[:rel_player_appearances_team]->(t:teams)
MATCH (pa)-[:rel_player_appearances_player]->(p:players)
WHERE t.team_code = 'CHN'
RETURN p.family_name AS f
```

`multiple reading clauses is not supported`. So was the comma form. Both describe one node with two relationships hanging off it — a shape a chain cannot hold.

The customer line already has this. Leaving it out here means the next port from main to that line would silently take it away again.

---

## How

The pattern was a list where edge *i* joined node *i* to node *i+1*. It is now a graph: an edge carries the two nodes it connects, and the analyzer keeps a variable-to-node map for the whole MATCH, so a variable written twice means the node it meant the first time. That is what makes two paths one shape — the second mention joins to the first instead of reading the table again. A first mention still names its object type; a label written again on a known variable must agree.

Generation emits tables in an order where each joins to one already in the statement, and writes `CROSS JOIN` for what the query never connected.

**Two defects found in review on the yf side are fixed here as part of the same change**, not left for a follow-up:

- A relationship whose ends are both in the statement already — a cycle closing, or a node related to itself — attaches no table but still carries a condition. Appending it to the FROM list only works when that list ends in an ON clause; it does not for a self-relationship, where no table is emitted, nor after a cross join. Such a condition now goes to WHERE, where it can stand on its own.
- Two relationships of one type between the same pair of nodes are the same relationship by construction, and Cypher requires them to differ. The distinctness rule had nothing to compare and emitted `NOT ()`. It now says what it means: a condition nothing satisfies.

Breaking changes: none — shapes that were refused now compile.

---

## Testing

- [x] Unit tests passed locally — 45 packages, no failures
- [ ] Integration tests — not applicable
- [x] Verified in test environment — on the development VM, which runs the yf line carrying identical compiler code

The compiler here is byte-identical to what merged as #1447; this line differs only in how the endpoint authorizes, which these changes do not touch. On that deployment, against a real database:

| Query | Result |
|---|---|
| The two-MATCH query above | 200 — 89 distinct players over 408 appearances |
| `MATCH (g:goals)-[:rel_goals_player]->(p:players), (g)-[:rel_goals_team]->(t:teams)` | 200 — 53 goals |
| `MATCH (t:teams), (m:matches)` — no shared variable | 200 — cartesian product, `CROSS JOIN` in the statement |
| The same shape written as one chain | 200 — identical answer |

Unit tests assert that all three spellings of one shape compile to the same SQL string, and cover the branching shape, the cartesian product, a second mention carrying conditions without reading the table twice, the self-relationship, a cycle closing on tables already read, a deferred condition beside a cross join, the impossible-by-construction pair, and four refusals.

---

## Risk & Rollback

- **A cartesian product is easy to write by accident.** `MATCH (a:X), (b:Y)` with no shared variable multiplies the tables. The row limit caps what comes back, not what it costs to produce; the `CROSS JOIN` is written out so it is visible in the statement.
- **Join ordering is derived rather than positional.** Tests assert whole statements, so a reordering or a dropped condition fails as a diff rather than as a wrong answer.

Rollback: revert this branch; the endpoint keeps working without these shapes.


sh00tg0a1:
Fixed in 7d1cfa7.

`382c4c37` is real and I reproduced it before fixing: `MATCH (n0:Order), (:Order) x40 RETURN count(*)` compiled to 41 tables in one `FROM` with 40 `CROSS JOIN`s. The bound now covers nodes as well as relationships (`CYPHER_MAX_PATTERN_NODES = CYPHER_MAX_PATH_LENGTH + 1 = 9`), refused at the analyzer with the same "a pattern this large" message. The regression test asserts both directions: too many nodes is refused, and a path of the longest allowed length — which names exactly nine nodes — still compiles. The API contract now states both limits.

On the 待确认项 (relationship uniqueness applied across MATCH clauses rather than within one): agreed, that is the wrong scope and it silently drops rows. It needs the hop to carry which MATCH it came from, so it is a separate change rather than a patch on this one — I will send it as a follow-up on both this line and 0.1.4-yf.

The `lint` failure is not from this PR: `docs/api/_shared/errors.yaml:139` uses `nullable` without `type`, which has been red on main since 2026-09-09 and fails on any PR that touches `docs/api`.

PR #1473, opened 2026-09-11
fix(vega,bkn): consolidate resource paging and Go lint

kenlee1988:
Consolidates resource paging to Paging.Offset and Paging.Limit, migrates BKN callers and cursor state, and standardizes make lint across production Go modules. Old duplicate paging fields are intentionally unsupported. Verified make lint for Vega, BKN backend, ontology-query, agent-retrieval, operator-integration, bkn-safe, agent-observability, comm-go, and OSS gateway; go test ./... for Vega; make test for bkn-safe and operator-integration. Rollback: revert this PR.


kenlee1988:
CodeQL reported the pre-existing AnyShare InsecureSkipVerify setting as a new alert because the golangci suppression changed that same line. The suppression now sits on the preceding line, preserving the existing TLS field unchanged; Vega make lint passes (0 issues). /review

kenlee1988:
Addressed the DB.Close review: writer.Close now still runs when reader.Close fails, while preserving the reader error as the return value. Verified comm-go make lint (0 issues) and go test ./db/sqlx. /review

PR #1490, opened 2026-09-12
chore: bring the 0.1.4-yf work onto main

sh00tg0a1:
## What

Everything that landed on `release/0.1.4-yf` and exists nowhere on main. The next release is cut from main, so without this a customer upgrading from the yf line would lose `run_cypher`, network comparison, and two fixes.

Already on main and **not** repeated here: the node bound (#1451), relationship uniqueness scope (#1460), `run_code` `bkn_context` (#1423), `_display` typing (#1421), the vega adapter log fix.

| Picked | What |
|---|---|
| 8edeec2, bb2d05b, fd0e23c | bkn-specification: checksum covers frontmatter; field-level network diff; one-sided definitions reported field by field |
| 1c8f075, 07ee7ab, 1e0e82c, c020e14, 555a7f6 | bkn-backend: compare two knowledge network branches, over HTTP, with object-type data counts on both sides |
| 624262d | bkn-backend: read an object type whose JSON column is empty |
| #1417 | context-loader: attach only the receipt fields a caller reads to MCP results |
| #1455 | context-loader: `run_cypher` MCP tool and `/kn/run_cypher` on both REST faces |

Each pick carries `-x`; where it needed more than a mechanical merge the commit message says what changed.

## Adapted for main

- **Data stats authorization** (own commit, 610f56e): yf checks `query_data` on the network; main grants it per object type with inheritance from the network. Checked on the network, the stats refused a caller holding `query_data` on exactly the object type being counted. Now `KNChildPermissionResource(RESOURCE_TYPE_OBJECT_TYPE, …)`, and the test pins the resource — it fails if the network-level check returns.
- **Network comparison authorization — reviewed, unchanged**: both sides load through `GetKNByID(..., Mode_Export)`, which on main requires direct visibility of the network and refuses child-only visibility. That is the export endpoint's own rule, so a comparison reveals nothing an export does not.
- **Capability bindings** (added on main after the yf cut) are assembled in `ExportToTar`, not in the shared `buildNetwork`: the model differ does not compare them, and reading them fails when the execution factory is unreachable, so assembling them for a comparison would only add a failure unrelated to the two models.
- **`BusinessDomain`** dropped from the comparison header — main retired the field in #1185.
- **CHECKSUMs** regenerated from main's example networks with the new checksum rule (`tools/regenerate_checksums.go`), not taken from yf.
- **`run_cypher` on main's CL**: annotated read-only (main requires every tool to declare its hints); route parity test rebuilt on main's handler shape (`find_skills` / semantic search are gone); README entries in English to match main's translated index; authorization wording changed to main's per-object-type model; tool texts now state the nine-node bound from #1451; sandbox stub regenerated.

## Testing

- bkn-backend: `go build ./...`, `go test -count=1 -gcflags="all=-l" ./...` (the CI invocation) — green
- context-loader: `go build ./...` and `-tags ee_dev`, `go test ./server/...`, the `ee_dev` matrix — green
- `make -C infra/sandbox bkn-tools-check` — consistent
- redocly lint on `docs/api/context-loader/*` and `docs/api/bkn/*` — no errors

The `lint` job will still be red until #1461 lands: `docs/api/bkn-safe/property-grants.yaml` has a duplicated mapping key on main, which stops redocly before it validates anything else.

Not yet exercised against a running main-line environment; the images this branch builds are what I will verify with before asking for merge.


sh00tg0a1:
Both 待确认项 were real; fixed rather than deferred.

**NULL primary keys in the data stats** — 206fce4. Only rows whose key is complete now take part in the distinct count (`COUNT(DISTINCT CASE WHEN <all key columns IS NOT NULL> THEN <key> END)`), the complete rows are counted alongside, and the remainder is reported as a new `missing_keys`. `duplicate_keys` is complete rows minus distinct complete keys, so a row without a key no longer reads as a duplicate and `(a, NULL)` / `(NULL, a)` no longer fold into one key. The new field is additive; `duplicate_keys` only changes where a key was missing. A new case (10 rows, 8 keyed, 7 distinct → 2 missing, 1 duplicate) fails against the old arithmetic, which reports 3 duplicates. Contract updated.

**Blank JSON columns on the relation and action type readers** — 8f94ff5. The helper moved to `common.UnmarshalStoredJSON`, and every stored-column decode in the object, relation and action type readers goes through it (mapping rules; condition, affect, action source, parameters, schedule). Malformed content is still an error. The impact contracts column already had its own guard and is untouched. A relation type reader test with a blank `f_mapping_rules` pins it and fails if the bare decode returns.

Also since the last round: the two comparison spans were renamed in English (0cd4ecc) for the hard-coded Chinese check.

/review

PR #1452, opened 2026-09-10
fix(vega): clarify policy-omitted semantic samples

kenlee1988:
## What Changed

Brief description of changes:

- [x] Add audited sample_context states and policy-omitted field metadata to resource semantic-understanding tasks.
- [x] Reconcile misleading Agent sample warnings into stable sample_omitted_by_policy warning details.
- [x] Add Prompt v3 migration, OpenAPI descriptions, and focused regression coverage.

---

## Why

- Related Issue: Closes #1418
- Related Feature: Vega semantic understanding
- Background: Binary and Other values are intentionally excluded from Agent sample payloads, but were previously liable to be described as unavailable samples.

---

## How

- Key implementation points: Distinguish not_requested, available, no_rows, unavailable, payload_limited, and all_fields_omitted_by_policy; retain genuine source failures for ordinary fields while preserving field-level policy evidence; add structured warning details for Studio localization.
- Breaking changes (API, config, database, etc.): Backward-compatible task JSON extension. MariaDB prompt data migration adds v3 and advances only installations still on built-in v2; operator overrides remain unchanged.

---

## Testing

- [x] Unit tests passed locally
- [x] Integration tests passed locally: not applicable; changed paths use mocked unit tests.
- [ ] Verified in test environment

Test notes:

- make lint
- make test
- make ci
- git diff HEAD --check

---

## Risk & Rollback

- Potential risks: New task input snapshots include sample_context and task results may include warning_details; older consumers ignore unknown JSON fields, while Studio PR #607 renders the structured detail.
- Rollback plan: Revert this commit. If prompt behavior must be reverted independently, set resource-semantic-understanding-prompt current version back to v2; v3 is append-only.

---

## Additional Notes

- Screenshots, logs, documentation links, etc.: Paired UI PR: https://github.com/openbkn-ai/bkn-studio/pull/607. Deploy Studio before or alongside Vega to render localized structured warnings.

PR #1499, opened 2026-09-12
docs(context-loader): give the execute_tool request examples the kn_id the schema requires

sh00tg0a1:
Both request examples on `POST /kn/execute_tool` omitted `kn_id`, which `ExecuteToolRequest` marks required and the handler refuses without — a reader copying the example got a 400. The OpenAPI lint's two findings on this file were exactly these; the file now lints clean with the pinned `@redocly/cli` 1.34.17.

Context: the docs `lint` job's red on 9/9–9/10 was a YAML duplicate key in `bkn-safe/property-grants.yaml` (introduced by #1420, fixed by #1452), not these warnings — `.redocly.yaml` keeps every rule but `no-unresolved-refs` at `warn`, and warnings never fail the job. This PR only removes the two real defects in the file I own.

### DIFF
--- a/docs/api/context-loader/tool.yaml
+++ b/docs/api/context-loader/tool.yaml
@@ -204,6 +204,7 @@ paths:
             examples:
               调用:
                 value:
+                  kn_id: kn_finance
                   toolbox_id: box_finance
                   tool_id: tool_fx_convert
                   arguments:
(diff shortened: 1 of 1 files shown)


sh00tg0a1:
/review

Docs-only PR: `docs/**` is outside the review workflow's `paths` whitelist, so no review ran on open; triggering it here.

PR #1512, opened 2026-09-13
fix(ontology-query): walk relations backward on type paths and derive edge endpoints

sh00tg0a1:
# Pull Request

## What Changed

Path queries (`POST /knowledge-networks/{kn_id}/subgraph?query_type=relation_path`, i.e. the `query_instance_subgraph` tool) in ontology-query:

- [x] **Regression fix**: with query authorization on, every edge that walks a relation against its definition was refused (`relation path does not match the published model`). Introduced by #1254, which required each edge to run from the relation's source to its target. Backward hops work again.
- [x] One direction resolver (`TypeEdge.TraversalDirection` + `logics.ResolveTypeEdgeDirection`) is shared by authorization and execution, so neither can accept a path the other rejects. Execution no longer walks an edge whose endpoints are not the relation's two ends as "backward"; it answers 400.
- [x] `source_object_type_id` / `target_object_type_id` are optional. They are the nodes a hop walks **from and to** (not the relation's own source/target), so they are derived from `object_types[i]` / `object_types[i+1]`. Supplied values must still agree, so existing requests behave exactly as before.
- [x] Optional per-edge `direction` (`forward` | `backward`). It only changes anything on a relation whose source and target are the same object type — such a relation could not be walked backward at all before. A `direction` that contradicts the endpoints is a 400.
- [x] `object_types` shorter than `relation_types` + 1 is reported instead of panicking on an out-of-range index (callers saw a bare `EOF`).
- [x] Every path error names the 1-based edge, the relation type, and the expected vs. supplied endpoints (the previous messages used a 0-based index and never said which direction was expected).
- [x] `docs/api/ontology-query/ontology-query.yaml`: `TypeEdge` requires only `relation_type_id`, documents source/target as traversal endpoints and adds `direction`; the `RelationTypePath` / `SubGraphQueryBaseOnTypePath` required lists no longer name fields of the source-based query.

---

## Why

- Related Issue: #1062 (part 1 of 2; the Context Loader side — error passthrough, MCP schema, Context Loader docs — follows in a separate PR that should merge after this one, because relaxing the MCP schema relies on the server deriving the endpoints)
- Background: measured on the test server before the change, using an existing knowledge network:
  - A reverse hop filled in exactly as the current contract asks was rejected: `OntologyQuery.KnowledgeNetwork.InvalidParameter: relation path does not match the published model`.
  - The same hop filled in schema order was rejected with `第 0 条边引用起点对象类 X，但 object_types[0] 为 Y。`: 0-based, and it does not say what to write instead.
  - 2 edges with 2 nodes: `Post ".../subgraph?query_type=relation_path": EOF` (the handler panicked on `index out of range [2] with length 2`).

---

## How

- `interfaces.TypeEdge.TraversalDirection(rt)`: forward when the edge runs `rt.source → rt.target`, backward when it runs `rt.target → rt.source`, not ok otherwise, when an explicit `Direction` contradicts the endpoints, or when `Direction` is not one of empty / `forward` / `backward`. With `Direction` unset, a self relation resolves to forward — what path queries have always done.
- `logics.ResolveTypeEdgeDirection`: the localized 400 (`OntologyQuery.KnowledgeNetwork.InvalidParameter.TypePath`) built on top of it; called from `AuthorizeSubgraphByTypePath` and `buildObjectSubgraphByTypePaths`.
- `validateSubgraphQueryByPathRequest`: node-count check, node-id check, `direction` value check, endpoint derivation/agreement. Nodes beyond `len(relation_types)+1` were always ignored and stay accepted.
- Locale: new detail keys `PathNodeCountMismatch`, `PathNodeIDRequired`, `EdgeDirectionInvalid`, `EdgeEndpointsMismatch`, `EdgeRelationMismatch`, `EdgeDirectionConflict`; the now unused `SourceObjectTypeIDRequired`, `TargetObjectTypeIDRequired`, `EdgeSourceMismatch`, `EdgeTargetMismatch`, `DisconnectedPath` are removed (both catalogs).
- Breaking changes: none for valid requests. Behaviour that does change: a `direction` other than `forward`/`backward` on an edge (previously silently overwritten) is now a 400; an edge whose endpoints are not its relation's ends is now a 400 when authorization is off (it already was, with a vaguer message, when authorization is on).

---

## Testing

- [x] Unit tests passed locally (`go test ./...` and `go vet ./...` in `adp/bkn/ontology-query/server`)
- [x] Verified in test environment (14.103.77.23, overlay image on top of the deployed `0.1.5-main.20260912210327.sha9e2e390`, binary + locale files; rolled back afterwards)

New tests, each confirmed red with the fix reverted:

- `Test_validateSubgraphQueryByPathRequestEdgeEndpoints` — derivation, legacy form, explicit `direction`, extra nodes, schema-order reverse fill (zh + en details), single wrong endpoint, too few nodes (panicked before), empty node id, invalid `direction`.
- `TestResolveTypeEdgeDirection` — forward, backward, self relation default/backward, direction conflict, disconnected edge.
- `TestAuthorizeSubgraphByTypePathAcceptsBackwardEdge` (the #1254 regression) and `...RejectsEdgeThatCannotWalkRelation`.
- Service: an edge that is not the relation is rejected before any object is read.

Test-server results (`supply_942_verify`):

| Case | Before | After |
| --- | --- | --- |
| Reverse one hop (`mat2inv` walked inventory → material), contract-conformant endpoints | 400 `relation path does not match the published model` | 200, 2 paths |
| Same, endpoints omitted | — | 200, identical result |
| Forward then backward (product → inventory → material, `direction: backward` on hop 2) | — | 200, 2 two-hop paths |
| Forward one hop (`so2prod`) | 200 | 200, unchanged |
| Reverse hop filled in schema order | 400, 0-based detail | 400 `第 2 条边 …mat2inv 与路径节点不一致：路径要求 inventory → material，当前填写 material → inventory。…` |
| 2 edges, 2 nodes | `EOF` | 400 `路径有 2 条边，object_types 需要 3 个节点，当前只有 2 个。` |
| `direction: forward` against reverse endpoints | ignored | 400 `…指定 direction=forward，要求路径为 material → inventory，当前为 inventory → material…` |
| Relation that does not connect the nodes | walked as "backward" (auth off) | 400 `…连不上路径节点：路径要求 inventory → supplier，而该关系类定义为 material → inventory…` |

Not run end-to-end: `direction: backward` on a self relation. The only knowledge networks on the test server with self relations currently answer 503 `knowledge network proxy mapping is unavailable` for every query. The explicit-`direction` branch was exercised end-to-end (row 3) and the self-relation semantics are covered by `TestResolveTypeEdgeDirection`.

---

## Risk & Rollback

- Potential risks: callers that sent a junk `direction` value on a path edge now get a 400 instead of having it ignored. No in-repo caller, bkn-studio (`subgraphPathFor`) or bkn-sdk sends `direction` on path edges; `explore_subgraph` results carry `forward`/`backward` edges whose endpoints already agree, so feeding them back works.
- Rollback plan: revert this commit; no schema, config or data changes.

---

## Additional Notes

Found while verifying, unrelated to this change and not fixed here:

- A backward hop onto `supply_ontology_hand_salesorder` returns 500 `GetViewDataByIDFailed`: vega answers `text field product_code has no keyword feature, cannot be used for comparison` (the resource's local index maps the join field as text) and reports it as an internal error rather than a 4xx.
- `explore_subgraph` backward from `supply_ontology_hand_product` fails the whole request with 503 `CheckPermissionFailed: object type supply_ontology_hand_mon_task has no published data source`, instead of skipping the one candidate path through an unbound object type.


sh00tg0a1:
Addressed in 888d8d511 (second commit on this branch):

- **Unknown `direction` in the resolver** (inline suggestion): `TypeEdge.TraversalDirection` now returns not-ok for any value other than empty / `forward` / `backward`, and `ResolveTypeEdgeDirection` reports it with the same `EdgeDirectionInvalid` detail as request validation, so the shared resolver no longer depends on validation running first. Covered by a new `TestResolveTypeEdgeDirection` case.
- **`ontology-query.yaml` out of sync**: moved the ontology-query OpenAPI change into this PR (it was queued in the Context Loader follow-up). `TypeEdge` now requires only `relation_type_id`, documents source/target as traversal endpoints and adds `direction`; the `RelationTypePath` / `SubGraphQueryBaseOnTypePath` required lists no longer name fields of the source-based query.
- **`bidirectional` on a path edge now rejected**: checked the callers. bkn-studio (`subgraphPathFor`) and bkn-sdk never send `direction` on path edges, Context Loader's MCP schema for these edges had `additionalProperties: false` without a `direction` property, and `explore_subgraph` results only ever carry `forward` / `backward` per edge (bkn-backend's neighbour query emits exactly those two literals), so feeding them back stays valid. `bidirectional` is only meaningful for the source-based exploration, which this does not touch.

PR #1521, opened 2026-09-13
feat(authz): make model access platform-wide

rongwei-liu:
## What Changed

- [x] Give every enabled account baseline display and execute access to all large and small models.
- [x] Give the built-in network_builder role model lifecycle permissions.
- [x] Remove legacy per-model grants during startup reconciliation and prevent their recreation.

---

## Why

- Related Issue: openbkn-ai/bkn-studio#625
- Related Feature: Platform-wide model access
- Background: Model usage is a shared platform capability, while model configuration belongs only to knowledge-network builders and super administrators.

---

## How

- Key implementation points: Seed public display/execute policies, seed model lifecycle operations for network_builder, reject model object/custom-role grants, and stop model creation from writing creator-specific ACLs.
- Breaking changes (API, config, database, etc.): Existing model-specific allow/deny policies are removed on bkn-safe startup; model object-grant endpoints return 404 and custom-role model grants return 403.

---

## Testing

- [ ] Integration tests passed locally — Not applicable; the changed bkn-safe tests are isolated HTTP/unit tests.
- [ ] Verified in test environment — Not run.

Test notes:

- go test ./internal/seed ./internal/httpapi -count=1
- python3 -m py_compile app/utils/permission_manager.py app/test/test_permission_manager_authz.py

---

## Risk & Rollback

- Potential risks: Upgrades intentionally remove legacy per-model exceptions; operators must confirm the platform-wide model-access policy before deployment.
- Rollback plan: Revert this commit before the next bkn-safe restart, or restore policy rows from a database backup and redeploy the previous image.

---

## Additional Notes

- No deployment or data migration was performed from this branch.


rongwei-liu:
Addressed the blocking review finding in cee267ca5.

- Updated mf-model-manager so bkn-safe-authoritative model creation no longer writes a creator-specific large_model policy.
- Added a regression test asserting that this path completes without making a policy write request.
- Python syntax compilation passes. The local unit-test runner cannot import the service because the workstation lacks the runtime dependency opentelemetry; this is an environment setup limitation before test execution.

/review

PR #1522, opened 2026-09-13
fix(ontology-query): stop read-modify-write on action execution logs

sh00tg0a1:
# Pull Request

## What Changed

PR1 of the three-PR plan for #790 ([plan comment](https://github.com/openbkn-ai/bkn-foundry/issues/790#issuecomment-5650683091)). It has no storage model or mapping change.

- [x] `OpenSearchAccess.UpdateData`: partial update primitive (`_update` with `doc` or `script`, `retry_on_conflict`, refresh)
- [x] Action execution records are no longer read-modify-written: typed conditional writes replace `UpdateExecution`
- [x] Cancel never rewrites results, and a cancel is never undone by a later write
- [x] Response-only `results_total/offset/limit` are no longer persisted and no longer leak through the list endpoint
- [x] `keyword` on `GET .../action-logs` filters by execution id
- [x] OpenAPI (`docs/api/ontology-query/ontology-query.yaml`) updated

---

## Why

- Related Issue: #790 (items 4, 5, 6, plus a lost-update race found while working on it). Items 1, 2, 3 and 7 follow in PR2 (split result storage, awaiting owner confirmation on the issue) and PR3 (retention).
- Background: every progress update, status change and cancel read the whole execution document, mutated it in memory and re-indexed it. Reproduced on the test server with the current `main` image (`0.1.5-main…sha9e2e390`):
  - Cancelling a running execution that stores 1500 results rewrote the document with **1000** results: 500 were lost, because the read is clamped to 1000. The same write persisted `results_total: 1500, results_limit: 1000`, and the response said `cancelled_count: 0`.
  - A progress update landing between a cancel's read and its write put `running` back, silently undoing the cancel. Unit reproduction below.
  - `GET .../action-logs?keyword=…` came back unfiltered. Studio's task panel sends it as "Search execution ID", but `ActionLogQuery` had no such field.

---

## How

- **`MarkExecutionRunning`**: painless script, `pending → running` only. A cancel issued before the executor started is kept.
- **`UpdateExecutionProgress`**: `{"doc": {success_count, failed_count, results}}` merge. It does no read first and never writes the status.
- **`FinishExecution`**: painless script. If the stored status is already `cancelled` it stays `cancelled`; counters, results and timing still record what ran. This aligns `per_instance` with the `once`-mode rule from #769. Previously a cancel that arrived after the last batch check was overwritten with `completed` even though the cancel API had answered "cancelled".
- **`CancelExecution`**:
  - Reads the summary with `_source.excludes: ["results"]`.
  - A painless script then cancels only `pending`/`running` executions and never touches `results`.
  - If the script reports `noop` (the execution finished in between), the error names the status that won.
- **Cancellation check inside the executor**: `GetExecutionStatus` reads the `status` field only; before, it read the whole document every 100 instances. Large runs now check at batch boundaries and whenever a second has passed since the last check, so a cancel is noticed within about a second instead of possibly never in the tail of a run.
- **List**: `_source` also excludes `results_total`, `results_offset` and `results_limit`, which older documents persisted with stale values (seen on the test server: `results_total 8800` on an execution storing 8808 results).
- **`keyword`**: a case-insensitive literal substring of the execution id. It is a `wildcard` query with `*`, `?` and `\` escaped, and it applies to the page and to `total_count`. The term is trimmed and capped at 128 characters (400 beyond that).
- The painless scripts were exercised directly against the test server's OpenSearch 2.19.4 on a temporary index before wiring them in. Checked:
  - Every allowed and refused status transition, including the resulting `noop`s.
  - A `doc` merge after a cancel leaves the status alone.
  - A 404 on a missing document.
  - Number round-trip: 19-digit longs and `0.1` survive a partial update unchanged.

**Behavior changes**, documented in the OpenAPI spec:
- An accepted cancel is final (see `FinishExecution`).
- `cancelled_count` in the cancel response is now the number of calls not yet run (`total − success − failed`, as of the last progress update). It used to be always `0`, because pending results are never stored. `completed_count` comes from the same progress counters.
- The list returns 400 for a `keyword` over 128 characters.

**Breaking changes:** none for API consumers. Response shapes are unchanged, and no consumer reads the fields removed from the list: agent-retrieval, Studio, bkn-sdk CLI and the bkn-backend scheduler were checked. `ActionLogsService.UpdateExecution` is replaced by typed methods; it is internal to ontology-query.

---

## Testing

**Test notes**

Reproduction first: against the pre-fix code, two tests failed.
```
--- FAIL: Test_Repro790_CancelRewritesTruncatedResults      Expected: 1500  Actual: 1000
--- FAIL: Test_Repro790_UpdateCarriesStaleStatusAndMetadata  (status carried by a progress write)
```
They are replaced by tests against the new API in `logics/action_logs/action_logs_writes_test.go`. These use a strict mock, so any whole-document `InsertData` fails the test.

New tests:
- `action_logs_writes_test.go`: cancel with more than 1000 results; cancel rejected on a terminal status; cancel racing the terminal write; the document vanishing between read and cancel; progress never carries status or pagination fields; mark-running script; finish keeps `cancelled` and stores an empty array rather than null; status-only read; keyword wildcard, escaping and total count; blank keyword.
- `opensearch_access_test.go`: request path and query, `noop` result, 404 mapped to `ErrDocumentNotFound`, error paths.
- `action_log_keyword_test.go` (handler): binding from the query string, trimming, rune-counted limit.
- `action_scheduler_service_test.go`: the existing executor tests are ported to the typed methods and now assert that the terminal write happens exactly once. A new test covers a `per_instance` execution cancelled midway: it stops invoking, records the rest as `cancelled`, and ends `cancelled`. `Test_shouldCheckExecutionCancellation` covers the one-second check bound.

Commands:
- `go build ./... && go vet ./... && go test ./...` in `adp/bkn/ontology-query/server`: all packages pass.
- `redocly lint` on the changed spec: valid, and the warning count equals `main`'s (2237).

**Test server (14.103.77.23).** The binary was built from the deployed commit `9e2e390` plus these four commits, swapped into `ontology-query` (md5 checked in the pod), and the original image was restored afterwards. A scratch index `ontology_action_executions_verify790_tmp`, created with the service's mapping, was used and then deleted.

| Check | Old image (`sha9e2e390`) | This PR |
|---|---|---|
| Cancel a running execution storing 1500 of 2000 results (3 failed) | stored results **1000**; `results_total: 1500`, `results_limit: 1000` persisted; `cancelled_count: 0, completed_count: 1000` | stored results **1500**; nothing persisted; `cancelled_count: 497, completed_count: 1500` |
| Cancel it again | n/a | 400 `cannot be cancelled, current status: cancelled` |
| `keyword=AFTER_A` (case differs) | unfiltered, 2 of 2 | 1 match, `total_count: 1` |
| List entry of a legacy document with persisted pagination fields | `results_limit` and `results_total` leaked | not returned |
| `keyword=*` / `keyword=_` | n/a | 0 matches (literal) / all ids containing `_` |
| `keyword` of 129 characters | n/a | 400 `keyword must not exceed 128 characters` |
| Real execution of simulated action `supply_942_verify/sim_replenishment` (tool reports `business_write: false`) | n/a | pending → running → completed through the scripts; stored doc has 1 result and no pagination fields; `keyword` on the id prefix finds it; no write errors in logs |

Not exercised end-to-end: a cancel racing a live multi-batch `per_instance` run. It would need a long-running tool with real side effects. The script semantics that decide it were verified directly on OpenSearch (see How).

---

## Risk & Rollback

- Potential risks:
  - `_update` re-serializes `_source` inside OpenSearch rather than storing the client's bytes verbatim. Longs and decimals were verified to round-trip, but key order in stored documents may differ, which is harmless.
  - Painless inline scripts must be allowed. That is the OpenSearch default (`script.allowed_types` empty = all), and it is the case on the test server.
  - During a rolling upgrade, an old pod still does whole-document writes, so the old race can still happen until every replica runs this version.
- Rollback plan: revert or roll back the image. The document shape is unchanged, so old and new versions read each other's records.

---

## Additional Notes

- Commits are split by concern: the OpenSearch primitive, the conditional writes, the list exclusion, then keyword and docs.
- The execution record created by the real-flow check (`01a098de-1732-7834-9775-8492d6d6d833` in `supply_942_verify`) is a normal log entry of a simulated action and was left in place.


sh00tg0a1:
Replies to the two non-blocking notes:

1. **Cancel detection interval on large runs.** Agreed, fixed in 8691bc48e. Before, runs with more than 100 instances checked only at batch boundaries, so a cancel landing in the last 99 invocations went unnoticed, and `cancelled_count` overstated what was stopped. The check now reads the `status` field only, so it also runs whenever a second has passed since the previous check. This bounds detection to about a second (or the next batch boundary) and costs at most one small search per second per running execution. The new `Test_shouldCheckExecutionCancellation` covers it. The OpenAPI description of `cancelled_count` now states the bound, and that calls already issued before the cancel was noticed are still recorded.

2. **Leading-wildcard `keyword` on `id`.** It is an automaton intersection with the term dictionary of one knowledge network's index. Current scale on the test server: the largest index holds 108 executions. Even at hundreds of thousands of ids, a scan of that kind costs milliseconds per segment. The count query repeats it only when `need_total=true`. Growth is bounded by the retention job that PR3 of #790 adds. Studio's task-panel input currently refetches on every keystroke without a debounce; that is a front-end follow-up, not part of this PR.

PR #1518, opened 2026-09-13
fix(operator-integration): keep a published MCP Server serving while it is edited

sh00tg0a1:
# Pull Request

## What Changed

- [x] The MCP proxy (`GET /mcp/proxy/{mcp_id}/tools`, `POST /mcp/proxy/{mcp_id}/tool/call`) serves an **editing** MCP Server from its release instead of its draft config.
- [x] Entering `editing` or `unpublish` through the status API no longer moves the config to a new version; the version moves on the first content edit made on a released version.
- [x] Republishing the version that is already released no longer retires its own instance.
- [x] Editing a version that has no deployment creates the deployment instead of rebuilding the instance in memory only.

---

## Why

- Related Issue: Closes #1478
- Background: `editing` is a published server with a draft beside it (`BizStatusEditing // Published and editing`), and every content edit of a published server moves it there automatically (`editStatusTrans`). Operators (`ExecuteOperator` reads `t_operator_release`) and Skills (`getPublishedSkill`, the skill index) already serve the release while editing. MCP was the only resource with a release table and versioned instances that served the draft instead.

Reproduced on the test server running main (`f317b27`), three scenarios on fresh `tool_imported` servers:

| Scenario | Step | list | call |
|---|---|---|---|
| A. status-only editing (the issue) | `editing` | 404 `MCPInstanceNotFound` | 400 `MCPServerNotPublished` |
| | republished, nothing edited | 404 `MCPInstanceNotFound` | 404 `MCPInstanceNotFound` |
| B. edit a published server | edited from [A] to [A,B] | 200, draft tools [A,B] | 400 `MCPServerNotPublished` |
| C. offline and back | republished from `offline` | 404 `MCPInstanceNotFound` | 404 `MCPInstanceNotFound` |

B means any edit of a published server, down to its description, took it away from every caller until it was republished. C is not in the issue but has the same root: publishing retires the previous release's instance, and when the previous release is the version being republished, it retires the instance about to be served.

---

## How

- `servingListToolsRequest` resolves what the proxy serves. An editing server gets its release's version, mode, URL and headers, so a custom server keeps its released endpoint too. Any other status gets its config; for a published server that is exactly what was released, so the published hot path gets no extra query.
- The call gate admits `editing` alongside `published`. `offline` and `unpublish` are still refused with 400 `MCPServerNotPublished`, as #1491 made them.
[...]

PR #1527, opened 2026-09-13
feat(operator-integration): list an editing MCP Server's draft tools on request

### LINKED ISSUE #1523: feat(mcp): list an editing MCP Server's draft tools for its authoring page
## Background

#1518 (the fix for #1478) serves an editing MCP Server from its release: `GET /mcp/proxy/{mcp_id}/tools` and `POST /mcp/proxy/{mcp_id}/tool/call` answer with the released version while the draft is being edited. That is right for every runtime caller — agents, action-type bindings, capability mounting — but Studio's MCP detail page builds its tool rail, the input form and the Debug button from that same listing (`McpDetailScene.tsx`; `McpDetailDrawer.tsx` outside market mode), while Debug (`POST /mcp/{mcp_id}/tool/{tool_name}/debug`) runs the draft.

On an editing server the author would therefore see the released tools: a tool added in the draft is not listed and cannot be debugged from the page, a removed one is still listed, and the debug form is built from the released schema.

## Proposal

- **Backend**: an optional query parameter on the proxy listing (e.g. `draft=true`) that lists the config's version instead of the release. It requires `view` permission; execute-only and public-access callers are refused. The default stays the released listing.
- **Studio**: the management detail page and the drawer outside market mode pass it; catalog and market views do not.
- **Order**: Studio first — the handler ignores unknown query parameters, so it is a no-op until the backend lands — then the backend. No window in which the authoring page regresses.

## Acceptance

- Editing server with release [A] and draft [A, B]: default listing → [A]; `draft=true` with view permission → [A, B]; `draft=true` with execute permission only → 403.
- Published, unpublished and offline servers: `draft=true` returns the same as the default.
- Studio detail page on an editing server lists the draft tools and debugs a draft-only tool.

sh00tg0a1:
# Pull Request

## What Changed

- [x] `GET /mcp/proxy/{mcp_id}/tools` takes an optional `draft` query flag that lists the config's own version instead of the served one.
- [x] On the public face the draft listing takes `view` permission; `execute` or `public_access` stays enough for the default listing, which is unchanged.
- [x] The listing handler binds the query string (it bound only the URI and headers).
- [x] The public and private MCP API docs describe the served version and the new flag, and the public doc no longer says an editing server refuses tool calls.

---

## Why

- Related Issue: Closes #1523
- Background: #1518 serves an editing MCP Server from its release, which is what every runtime caller runs. Studio's MCP detail page builds its tool rail, input form and Debug button from the same listing while Debug runs the draft, so since #1518 an author editing a server sees the released tools: a tool added in the draft cannot be selected or debugged, and a removed one stays listed. The Studio side (passing `draft=true` from the management views) is a separate bkn-studio PR.

---

## How

- `GetMCPTools` builds the request from the config when `draft` is set and from `servingListToolsRequest` otherwise; the config-based construction is shared through `configListToolsRequest`. The flag changes the answer only while the server is editing: in every other status the config is what is served.
- The permission set is chosen before the check: `view` alone for the draft, `view` / `public_access` / `execute` for the default.
- The handler adds `ShouldBindQuery`. Query parameters the request does not declare are still ignored, so the paging parameters Studio already sends keep working; an invalid `draft` value (e.g. `draft=yes`) now answers 400.
- The unit tests match the variadic operation list as a whole with a slice matcher. A bare value in that position is compared with the first operation only — gomock does not check the remaining arguments when the matcher count equals the method's parameter count — which let the check against the wider permission set pass on the unfixed code.
- Breaking changes: none. The flag is opt-in; the default listing, the call path and the internal face behave as before.

---

## Testing

- [x] Unit tests passed locally: every package CI runs (`go list ./... | grep -v /server/tests | grep -v /server/mocks`, `-gcflags="all=-N -l"`). The new tests (`TestGetMCPToolsListsTheDraftOnRequest`, `TestGetMCPToolsBindsTheDraftFlag`) fail on the code before this change.

Test notes: patched binary swapped into `agent-operator-integration` on the test server (md5 of `/proc/1/exe` matched the local build), then the image was set back and every test server removed. `docs/api/execution-factory/mcp.yaml` passes `redocly lint`.

| Server | Query | Result |
|---|---|---|
| editing, release [A], draft [A, B] | (none) | 200 [A] |
| | `draft=true` | 200 [A, B] |
| | `draft=false` | 200 [A] |
| | `draft=yes` | 400 `Public.BadRequest` |
| | `page=1&page_size=100&draft=true` (Studio's params) | 200 [A, B] |
| published [A, B] | (none) / `draft=true` | 200 [A, B] / 200 [A, B] |
| unpublished [A] | (none) / `draft=true` | 200 [A] / 200 [A] |

With the Studio change served by a local dev server against the test server: the management detail page of the editing server lists [A, B] (before: [A]), the catalog view lists [A], and debugging the draft-only tool B from the page reaches the draft instance (`POST .../tool/list_knowledge_networks/debug` → 200). The 403 for an execute-only caller is covered by the unit test only; no execute-only account was set up on the test server.

---

## Risk & Rollback

- Potential risks: an execute-only caller that opens Studio's management detail page gets 403 on the tool listing (shown as the page's tool-load error); such a caller cannot read the management record either.
- Rollback plan: revert the PR; Studio falls back to the served listing because the handler ignores the flag again.

---

## Additional Notes

- Merge order: the Studio PR can land first — the handler before this PR ignores unknown query parameters, so it is a no-op until this lands.


sh00tg0a1:
Studio side: openbkn-ai/bkn-studio#630. It can merge first — this handler ignored unknown query parameters before this PR, so the flag is a no-op until this lands. The red `Analyze (go)` here failed while uploading its SARIF (no findings reported; the analysis step completed), and that run cannot be retried; it is not a required check.

PR #1528, opened 2026-09-13
feat(ontology-query): store action execution results in their own index

sh00tg0a1:
# Pull Request

## What Changed

This is PR2 of the three-PR plan for #790 ([plan and owner confirmation](https://github.com/openbkn-ai/bkn-foundry/issues/790#issuecomment-5650683091), label `owner-confirmed`). Per-instance action results move out of the execution document into their own index.

- [x] `OpenSearchAccess.BulkIndexDocuments`: bulk index under explicit ids, with a clean source and item-level error check
- [x] Results store: new shared index `ontology_action_execution_results`, one document per result, paged and filtered in OpenSearch
- [x] Executor appends only new results per batch; progress and terminal writes carry counters only
- [x] New `GET /api/ontology-query/{v1,in/v1}/knowledge-networks/:kn_id/action-logs/:log_id/results?offset&limit&status`
- [x] Executions written before this change (results embedded in the document) stay readable, with no migration
- [x] OpenAPI updated

---

## Why

- Related Issue: #790, items 1, 2 and 3 plus the historical-document compatibility. PR1 (#1522) covered items 4, 5 and 6; PR3 (retention) follows.
- Design: openbkn-ai/bkn-docs#126
- Background: after #1522 the execution document is no longer read-modify-written, but results were still embedded. Three problems remained:
  - Every progress batch re-sent all results so far: O(N²/batch) data per execution.
  - The detail API read every result to return one page.
  - `results` was a `nested` field mapped dynamically from arbitrary tool output, so it hit the 10000 nested-object cap and field-type conflicts.

---

## How

**Storage**

- Index `ontology_action_execution_results` is created on first use:
  ```json
  {"settings": {"number_of_shards": 1, "number_of_replicas": 0},
   "mappings": {"dynamic": false, "properties": {
     "kn_id": {"type": "keyword"}, "execution_id": {"type": "keyword"},
     "seq": {"type": "integer"}, "status": {"type": "keyword"}}}}
  ```
  Everything else in a result (parameters, tool output, identity, targets, error) stays in `_source` unmapped, so no tool output can add fields to the mapping or conflict with another result's types.
- One index for all knowledge networks, as confirmed on the issue. It adds one shard in total instead of one per knowledge network.
- Documents use the id `{execution_id}_{seq}`, where `seq` is the instance's position in the execution (0 for once mode). A retried batch therefore overwrites instead of duplicating.
- Existing `ontology_action_executions_*` indices are untouched: no mapping change, no reindex, no migration.

**Write path (executor)**

- A `resultBuffer` keeps only the results produced since the last flush.
- At each progress point, the new results are bulk-indexed (500 per request) at their positions, then only `success_count`/`failed_count` are merged into the execution document.
- A batch that fails to store stays pending and is retried at the same positions with the next flush. The last flush retries with backoff. Bulk writes use `refresh=wait_for`, so they return once searchable without forcing a refresh of the shared shard.
- The last partial batch, and every instance skipped by a cancel, are stored before `FinishExecution`, which now writes status, counters and timing only.

**Read path**

- Detail, `action-executions` and the new endpoint all go through `pageResults`:
  - If the execution document still embeds results, it is a pre-#790 execution and is paged in memory, as before.
  - Otherwise results come from the results index: `term` filters on `kn_id`, `execution_id` and optional `status`, sorted by `seq`, with `from/size` and a `_count` for the total.
  - A missing results index reads as no results.
- Response shapes of the existing endpoints are unchanged. `action-executions` still returns the first 1000 results, so agent-retrieval's `get_action_execution` is unaffected.
- Legacy detection is "embedded results non-empty", not a marker field. During a rolling upgrade an old pod may still rewrite a document and would drop a field it does not know; an empty embedded array still routes to the index.

**New endpoint and limits**

- `GET .../action-logs/:log_id/results` returns `{entries, total_count}`.
  - `limit` defaults to 100 and is capped at 1000.
  - `offset` must be ≥ 0 and `offset + limit` ≤ 10000 (OpenSearch's default `max_result_window`); otherwise 400.
  - `status` is one of `success | failed | cancelled`.
  - An execution that does not belong to the knowledge network returns 404.
- The detail endpoint now returns 400 instead of 404 for a results page that ends beyond 10000.

**Breaking changes:** none for API consumers. `ActionLogsService` gains `AppendResults`/`QueryResults`; `ExecutionProgress`/`ExecutionOutcome` lose `Results`. Both are internal to ontology-query.

---

## Testing

- [x] Integration tests passed locally (real OpenSearch, see below)

**Unit.** `I18N_MODE_UT=true go test -gcflags=all=-l ./...` passes in all 25 packages. The hard-coded Chinese check passes (0 new). `redocly lint` passes on the spec, with the warning count unchanged from `main`. New tests:
- **`action_results_store_test.go`**: bulk ids and positions across three 500-document requests; the index is created once, with `dynamic: false` and no mapping for `result`; OpenSearch-side filter, sort, page and count; legacy embedded paging; missing index; window limit; detail reads the index total.
- **`action_scheduler_service_test.go`**: `AppendsOnlyNewResults` with 250 instances. The first batch write fails, then positions (0,100) → (0,200) → (200,50) show the retry at the same positions; progress carries counters only. The existing executor tests now read results from `AppendResults`.
- **`action_log_results_test.go`** (handler): binding, defaults and caps, 400s for window, negative offset and bad status, 404/500 mapping, and 400 on the detail window.
- **`opensearch_access_test.go`**: `BulkIndexDocuments` metadata and body, empty batch, and item or response errors.

**Integration against the test server's OpenSearch 2.19.4.** This used the real executor and action logs service from this branch over an SSH tunnel, with only the tool call stubbed (every 10th call fails on purpose). It is not committed.

| Scenario | Result |
|---|---|
| per_instance, 250 instances, runs to completion | `completed`, 225/25; 250 result docs, 25 `failed`; page at offset 245 ends at `inst_249` |
| per_instance, 250 instances, cancelled 2.5 s in | cancel landed at call 114; executor stopped at call 146 (within about 0.6 s); 250 docs = 146 run + 104 `cancelled`; final `cancelled` |
| once, 30 targets | 1 tool call, 1 result doc at seq 0 carrying 30 targets |
| cancel fired inside the very last invocation, after the last check | final status stays `cancelled`; all 150 results stored |

**Test server (14.103.77.23).** The binary was built from the deployed commit `69af2c1` plus #1522 and these commits. It was swapped into `ontology-query` (md5 checked in the pod); the original image was restored afterwards and all scratch data deleted.
- Before the swap, the results index did not exist.
- Reads, with 10000 synthetic results for one execution and 8000 embedded results for a legacy one:

| Read | Latency |
|---|---|
| detail, first page / last page (offset 9900) / `status=failed` | 13 ms / 9 ms / 5 ms |
| `/results?offset=5000&limit=1000` | 18 ms |
| legacy execution, detail or `/results` (in-memory compatibility path) | ~48 ms |
| list endpoint on three knowledge networks | 5–22 ms (acceptance: < 3 s) |
| `/results?offset=9950&limit=100` | 400 `results offset + limit must not exceed 10000` |

- The failed filter on the integration execution returned its 25 failures with their error messages.
- **Rollback.** With the original image back, reading the new-format executions returns 200 with 0 results and no error, and the legacy execution still returns its 8000 results. This matches the rollback plan on the issue.
- **Not exercised end-to-end through a real tool.** On the test server every action execution currently fails before invoking anything: `OntologyQuery.Proxy.Unavailable: knowledge network proxy mapping is unavailable`. The original `main` image (`69af2c1`) fails the same way, including the `execute/check` endpoint (503), so this PR is not the cause. The integration run above covers the same code path with the tool stubbed.

---

## Risk & Rollback

- **Potential risks**
  - If the results index is deleted by hand while the service runs, the next bulk write lets OpenSearch auto-create it with default dynamic mapping, because the "index exists" state is cached per process. Restarting ontology-query after such a manual deletion avoids this.
  - If the last flush of a run still fails after its backoff retries (0.5 s, 1 s, 2 s), those results are missing and an error log names the positions; the terminal write still records the counters. Before this PR, the equivalent failure lost the whole terminal write and left the execution `running`.
  - During a rolling upgrade, old pods show 0 results for new-format executions until the rollout completes.
- **Rollback plan:** roll back the ontology-query image. Nothing is deleted; results written meanwhile stay in `ontology_action_execution_results` and are visible again after re-upgrading. To remove the feature entirely, delete that index once no rollback-forward is planned.


sh00tg0a1:
Both notes are addressed in the latest commit:

1. **Both final flushes failing.** The last flush of an execution now retries with backoff (0.5 s, 1 s, 2 s) before the terminal write. If results still cannot be stored, an error-level log records the execution and the unstored positions; the terminal write still records the counters of what ran. For comparison, before this PR the equivalent storage failure made the whole terminal write fail, so the execution stayed `running` forever with its results missing. Now the status and counters are right, and any gap in results is logged instead of silent. `Test_executeAsync_RetriesFinalResultsFlush` covers both paths: stored on the third attempt, and giving up after the schedule while the terminal write still happens.

2. **Forced refresh on the shared single-shard index.** Bulk result writes now use `refresh=wait_for`. They return once the documents are searchable, without forcing a refresh, so concurrent executions no longer each refresh the same shard. I checked this on the test server's OpenSearch 2.19.4. After letting a temporary index go search-idle (40 s without searches), a `wait_for` bulk returned in 0.68 s and its documents were immediately countable. The cost is at most about one refresh interval per 100-instance batch on the executor's own goroutine.

PR #1538, opened 2026-09-13
fix(execution-factory, context-loader): keep an editing MCP Server searchable and executable

### LINKED ISSUE #1524: bug(mcp): capability index drops an editing MCP Server that the proxy still serves
## Problem

After #1518 an editing MCP Server keeps serving its release, but `syncMCPCapabilities` purges every server whose status is not `published` (`logics/mcp/capability_index.go:46`, from #1443), and it runs after both `UpdateMCPStatus` and `UpdateMCPServer`. Editing a published server — any edit moves it to `editing` — therefore removes its tools from `search_capabilities` while they stay callable through the proxy.

Skills already index the release while editing (`skillIndexPayload` in `logics/skill/index_sync.go`).

## Expected

An editing server is indexed from its release. Since #1518 `GetMCPTools` already resolves an editing server to its release, so the gate can admit `editing` alongside `published`. Offline and unpublished servers are still purged.

## Acceptance

- Publish → tools searchable.
- Edit (auto `editing`) → still searchable with the released tools; a draft-only tool is not searchable.
- Republish → the draft's tools become searchable.
- Offline → purged.

sh00tg0a1:
# Pull Request

## What Changed

- [x] Execution factory: the capability index keeps an editing MCP Server's tools, indexed from its release, instead of purging them.
- [x] Context Loader: `MCPServerIsUsable` accepts `editing` as well as `published`, so `search_capabilities` stops withholding an editing server's tools and `execute_tool` stops refusing them.
- [x] A comment in Context Loader's `execute_tool` that said the MCP proxy performs no status check of its own is corrected.
- [x] An editing MCP Server with no release (what importing an exported editing server produces) is treated as not served: the proxy refuses to call it and the index purges it, instead of both falling back to the draft.

---

## Why

- Related Issue: Closes #1524
- Background: since #1518 an editing MCP Server is a published server with a draft beside it and is served from its release. Two places still treated it as unavailable, so editing a mounted server — any edit moves it to `editing` — took it away from agents until it was republished:
  - the capability index purged every server that was not `published` (the #1443 rule, `capability_index.go`);
  - Context Loader asked the execution factory whether the server was usable and accepted only `published`, both when filtering `search_capabilities` hits and before `execute_tool` ran. The second means #1518 did not reach agents calling MCP tools through Context Loader.

---

## How

- `syncMCPCapabilities` admits `editing` alongside `published`. It lists the version the proxy serves — the release for an editing server — so the index holds what callers run: a tool added in the draft is not indexed until the draft is published, when the publish sync writes it. Servers purged while editing under the old rule come back on the next reconcile pass (which also runs at startup). Unpublished and offline servers are still purged.
- `MCPServerIsUsable` returns true for `published` or `editing`; offline, unpublished and unreadable servers are still refused.
- `servingListToolsRequest` now also reports whether the server is served at all: published, or editing with a release. The call path and the index sync both decide from it, so they cannot disagree; the sync lists the version it resolves instead of going through `GetMCPTools` (one config read instead of two). An editing server with no release — the status machine never produces one, but the importer writes the exported status and records a release only for published servers — is refused on call (400 `MCPServerNotPublished`, before connecting) and purged from the index. Its default listing still shows the config, as drafts always have. This answers the review's open question: before, such a server's draft was called through the proxy (since #1518) and would have been through Context Loader too.
- Mounting is unchanged on purpose: bkn-backend still refuses a new mount of a server that is not `published`, so an editing server cannot be newly mounted; existing mounts keep working while it is edited.
- Breaking changes: none in API, config or schema. Behaviour change: an editing MCP Server's released tools stay searchable and executable through Context Loader.

---

## Testing

  - operator-integration: every package CI runs (`go list ./... | grep -v /server/tests | grep -v /server/mocks`, `-gcflags="all=-N -l"`); the new `TestSyncMCPCapabilitiesIndexesAnEditingServerFromItsRelease` lists a real in-process MCP server per version and asserts only the released tool is upserted and nothing is purged. It fails before the fix (the server is purged).
  - agent-retrieval: the CI commands (`go build ./...`, `go build -tags ee_dev ./...`, vet and tests of `./server/extension/...` and `./server/driveradapters/mcp/`, plus `-tags ee_dev`) and `go test ./server/...`. The new `TestMCPServerIsUsable` covers published / editing / offline / unpublish / unreadable; editing fails before the fix. CI does not run `drivenadapters` tests, so it only runs locally.

Test notes: both patched binaries were swapped into the test server (md5 of `/proc/1/exe` matched the local builds), a scratch knowledge network was created, and a `tool_imported` server with tool A was published and mounted on it; afterwards both images were set back to their main tags and the network and servers deleted.

| Step | `search_capabilities` (A) | `execute_tool` (A) | New mount |
|---|---|---|---|
| before, published | found | 200 | — |
| before, edited (release [A], draft [A, B]) | not found | 400 `Public.BadRequest` | 400 `TargetNotAvailable` |
| before, republished | found | 200 | — |
| after, published | found | 200 | — |
| after, edited | found | 200 | 400 `TargetNotAvailable` |
| after, republished | found | 200 | — |

The index itself, read through `internal-v1/capabilities/search` with refs for both tools: published [A] → `get_operator_schema`; editing with draft [A, B] → `get_operator_schema` only; republished → both.

Editing without a release, simulated by deleting the server's `t_mcp_server_release` row while it was editing (then saving once to re-sync); the normal scenario above was rerun on this build with the same results:

| | index | default listing | proxy call | CL `search_capabilities` | CL `execute_tool` |
|---|---|---|---|---|---|
| editing, with release | [A] | [A] | 200 | found | 200 |
| editing, release deleted | [] | [A, B] (config) | 400 `MCPServerNotPublished` | not found | 400 |

---

## Risk & Rollback

- Potential risks: Context Loader now depends on the execution factory serving editing servers (#1518). Against an execution factory older than #1518, `execute_tool` on an editing server passes Context Loader's check and is refused by the proxy (400 `MCPServerNotPublished`) instead of by Context Loader — still a refusal, with a different code.
- Rollback plan: revert the PR; editing servers are purged from the index again on the next sync.

---

## Additional Notes

- The importer writing `editing` for a server that has no release in the target environment is the source of that data; it is not changed here.

- Not changed here: bkn-backend's capability listing marks an editing server's mounted tools as `disabled` when it backfills their metadata (`capability_binding/backfill.go`), although they are now callable. It is display-only — Context Loader does not read that status.


sh00tg0a1:
On the review's open question (an editing server with no release falling back to the draft): confirmed reachable — importing an exported editing server writes `editing` and records a release only for published servers. Fixed in 0b4eb590c and a623d3fb7: `servingListToolsRequest` now reports whether the server is served, the call path refuses such a server with 400 `MCPServerNotPublished` before connecting, and the index purges it instead of listing the draft; its default listing still shows the config. Verified on the test server by deleting a test server's release row while it was editing (table in the description). The importer itself is left as is.

Issue #1176, opened 2026-08-27
perf(bkn-backend,context-loader): search_schema 两次串行拉 Vega 资源，同一份富化做两遍


sh00tg0a1:
## 现象

有用户反馈升到 0.1.4 后 `search_schema` 很慢。在测试环境（全栈 0.1.4）实测，一次 `search_schema` 墙钟约 3.4s（KN 规模：18 个对象类 / 24 个关系类 / 14 个不同 Vega 资源）。

先说结论：**这不是 0.1.4 引入的回归**。0.1.3→0.1.4 在 schema 路径上唯一新增的同步开销是 #1090 把对象类塞进原本只放关系类的那次 rerank 调用（上限 `ObjectRerankCandidateLimit=200`），实测代价只有 ~2.5ms/doc，小网上 0.05–0.3s。真正的大头在 0.1.3 就已经存在，只是没人量过。

## 耗时拆解（实测）

| 阶段 | 耗时 | 说明 |
|---|---|---|
| `GET /in/v1/knowledge-networks/{kn}?include_detail=true&mode=export` | 0.55s / 167KB | 取全量 schema |
| `POST /in/v1/knowledge-networks/{kn}/object-types`（粗召回打分） | **1.15s** / 120KB | 见下 |
| rerank（远端 cross-encoder） | ~0.4s | 20 doc 0.62s / 100 doc 0.79s / 200 doc 1.10s |
| `GET .../object-types/{15 个 id}?include_detail=true`（condition_operations 回填） | **0.92s** / 96KB | 见下 |
| metric 召回 ×2 | 0.08s ×2 | 不是瓶颈 |

3.4s 里有 **2.1s** 落在标粗的两行上，而且这两行做的是**同一件事**。

## 根因一：逐个对象类串行拉 Vega 资源

`processObjectTypeDetails`（`adp/bkn/bkn-backend/server/logics/object_type/object_type_service.go:1769`）对每个对象类发一次 `GetResourceByID` 来富化 `condition_operations` / `mapped_field` 显示名。单次 Vega 调用实测 ~80ms，调用方是个串行 for 循环：

- 概念搜索路径：18 个对象类 → 14 个不同资源 × 80ms ≈ **1.2s**
- 对象类详情路径：15 个 id → 同一个循环再跑一遍 ≈ **0.9s**

[...]

PR #1561, opened 2026-09-14
perf(bkn-safe): stop serializing authorization decisions on the policy read

### LINKED ISSUE #1554: perf(bkn-safe): authorization decisions are serialized by an exclusive casbin lock
## Description

bkn-safe answers authorization decisions one at a time. Every decision takes an exclusive lock on the global casbin enforcer and, while holding it, scans every policy rule. Concurrent callers queue behind each other, so any caller that issues decisions in parallel gets no speedup, and every service that asks bkn-safe many small questions pays for each one in sequence.

Found while investigating #1176 (`search_schema` latency).

## Where

- `bkn-safe/server/internal/authz/filter.go:235`: `grantIndex` calls `en.e.GetImplicitPermissionsForUser(accessorID)`.
- `casbin/v2@v2.135.0/rbac_api_synced.go:170-173`: `SyncedEnforcer.GetImplicitPermissionsForUser` takes `e.m.Lock()` (exclusive), not `RLock()`. The other synced read methods (`GetImplicitRolesForUser`, `GetFilteredPolicy`) take `RLock()`, but a waiting writer blocks new readers, so they queue behind it too.
- Inside the lock, `GetNamedImplicitPermissionsForUser` loops over every p-rule and copies the matches, so each call is O(total rules).
- Every decision path reaches `grantIndex`: `POST /api/safe/v1/authz/resource-filter` (via `filterResourceOps`), `/check` (via `OperationDecision` / `LocalDecision`, unified in #1457), `/operations`, plus `EffectivePermissions` and `accessibleResources`.
- `grantIndex` (added in #705) put every decision on this path; #1457 moved `/check` onto it as well. #1540 (fixing #1511) reworked the write side only.

## Measured on the test deployment

| Call | One at a time | Concurrently |
|---|---|---|
| bkn-safe `resource-filter`, 10 calls | 0.45s total (25-60ms each) | 0.54s total, latencies stack 0.15s → 0.53s |
| Vega `GET /resources/{id}`, 14 resources (one decision each) | 1.34s | 1.47s, latencies stack 0.85s → 1.46s |
| Vega `GET /resources/{14 ids}` batch (one decision) | 0.107s | - |

Knock-on effect: ontology-query `GET /object-types/{id}/schema` costs ~0.45s per object type (several decisions each), and context-loader calls it once per object type. Parallelizing those calls in context-loader was tried for #1176 and gave no improvement on the test deployment (search_schema 8.2s → 8.5s), because the calls only queue in bkn-safe.

## Expected behavior

Read-only decisions from different requests proceed in parallel, and a decision does not cost a full scan of the policy set.

## Possible directions (not decided)

1. Take the enforcer's read lock (`SyncedEnforcer.GetLock().RLock()`) and call the unsynchronized `Enforcer.GetImplicitPermissionsForUser`. Small change, but it first needs confirmation that the whole call chain, including the role manager, is free of writes (lazy caches, link building). Each decision would still scan all rules, but decisions would overlap.
2. Cache the grant index per accessor, keyed by the published model version and invalidated in `publish`. Removes the scan, but invalidation must be exactly right: a stale entry is either a privilege escalation or a false denial.

- [ ] N concurrent `resource-filter` calls for different accessors complete in clearly less than N × the single-call latency on the test deployment.
- [ ] Decisions stay identical before and after for the existing authz test suites, and for grants and revocations made while requests are in flight (no stale allow after a revoke).
- [ ] `go test -race` for `internal/authz` passes with a concurrent read/write test.
- [ ] Re-measure `search_schema` with context-loader schema lookups parallelized (patch kept from #1176).

sh00tg0a1:
# Pull Request

## What Changed

- [x] `grantIndex` no longer loads policy rows for a super-admin. Its decisions never read them.
- [x] Policy reads for decisions, `EffectivePermissions` and `AccessibleResources` take the enforcer's shared lock instead of the exclusive one.

---

## Why

- Related Issue: Closes #1554 (also the remaining part of #1176)
- Background: every decision built its grant index through casbin's `SyncedEnforcer.GetImplicitPermissionsForUser`, which takes the enforcer's **exclusive** lock and copies out every policy row of the accessor. On the test deployment:
  - 133,169 of the 134,068 p rules belong to one accessor, the super-admin. 131,942 of them are per-resource grants from the ISF migration (`legacy`/`migration`, 22,087 resources × 6 operations).
  - A super-admin decision therefore cost 25-70ms, against 3.3ms for an ordinary user. The copy was the cost, not the scan.
  - Because of the exclusive lock, concurrent decisions queued: 20 in parallel took about 1s.
  - None of those rows affect a super-admin's decision: `localParts` allows every operation before reading them. `grantIndex` copied them all before it checked for the super-admin role.
- Every Vega resource read, object type fetch and ontology-query schema lookup made by a super-admin paid this cost, which is what made `search_schema` slow (#1176).

---

## How

- `grantIndex` resolves the accessor's roles first. For a super-admin it returns an index with no rows (only the subjects). For everyone else it reads the rows as before.
  - Every consumer reads the rows only inside `localParts`, which returns before touching them for a super-admin. `explain` lists grants from the store via `PolicyRecords`, not from the index, so it is unchanged.
- New `implicitPermissions` helper: takes `SyncedEnforcer.GetLock().RLock()` and calls the unsynchronized `Enforcer.GetImplicitPermissionsForUser`. It falls back to the plain call for the transaction-scoped enforcer.
  - Why this is safe (casbin v2.135.0 source): the read touches only the model, plus the role manager's temporary role for an accessor absent from the role graph. `Role`'s fields are all `sync.Map`, and bkn-safe registers no matching function on `g`, so that temporary create/remove touches only other `sync.Map`s. casbin's own synced `GetImplicitRolesForUser` walks the same path under `RLock`. `GetFieldIndex` has its own mutex.
  - Writers are unchanged and still take the exclusive lock (`publish`, casbin's synced mutators), so a read always sees a whole model.
- No new data structure, no data or schema change, no config change.

---

## Testing

  - `go test ./...` for bkn-safe/server; `go test -race ./internal/authz/`.
  - New `grant_index_test.go`. Each test is named below with what it proves:
    - `TestPolicyReadsDoNotTakeTheExclusiveLock`: holds the shared lock and runs filter / effective permissions / accessible resources. All three timed out before the change.
    - `TestGrantIndexDoesNotLoadRowsForSuperAdmin`: the super-admin index loads no rows, and decisions are unchanged, including allow over an explicit deny. It failed before the change (20 exact + 1 wildcard rows).
    - `TestExplainStillListsSuperAdminGrants`: explain still lists the super-admin's direct grants.
    - `TestConcurrentDecisionsAndWrites`: parallel decisions (super-admin, ordinary user, accessors absent from the role graph) alongside grant and role writes, under `-race`.

Test notes: the test server runs the enterprise image, which pins core `bkn-safe/server@f8031aaf`. The change was cherry-picked onto `f8031aaf` (clean) and built through the enterprise module with a local `replace`. The binary's md5 matched `/proc/1/exe`. The image was set back to its original tag afterwards.

| Measurement | Before | After |
|---|---|---|
| super-admin decision (`resource-filter`), p50 | 41-50ms | 0.7ms |
| 20 concurrent super-admin decisions, total | ~1000ms | 20ms |
| ordinary-user decision, p50 / 20 concurrent | 3.4ms / 55ms | 3.4ms / 33ms |
| Vega `GET /resources/{id}` × 14, serial / parallel | 1400ms / 1460ms | 80ms / 21ms |
| `search_schema`, 6 object types | 3.49s | 0.34s |
| `search_schema`, 18 object types | 8.33s | 0.73s / 0.57s |

Decision equivalence: 1,672 `resource-filter` queries run against the old and the new build and diffed. They covered 38 accessors (every role-bound subject, the top 25 policy holders, one with no grants), 22 resource types with up to 30 granted ids plus an ungranted id each, and both evaluation scopes. **0 differences.** The `search_schema` response was identical before and after.

---

## Risk & Rollback

- Potential risks:
  - The shared-lock read relies on casbin's read path staying write-free apart from the `sync.Map`s. A casbin upgrade should re-check `GetImplicitPermissionsForUser` / `RoleManagerImpl.getRole`. `TestConcurrentDecisionsAndWrites` under `-race` is the guard.
  - A non-super-admin that directly holds a very large number of grants still pays to copy its own rows, proportional to its own grant count rather than the total.
- Rollback plan: revert this PR; no data or config change.

---

## Additional Notes

- The enterprise image pins core by pseudo-version; it picks this up once openbkn-ee bumps its `bkn-safe/server` requirement.
- The 131,942 migrated per-resource grants of the super-admin are redundant with its `*`/`*` role grant. They no longer cost anything after this change; cleaning them up would be a separate data change.


sh00tg0a1:
For the reviewer, on the one point the automated review could not check: the casbin v2.135.0 read path this PR now runs under the shared lock.

- `rbac_api_synced.go:170` `SyncedEnforcer.GetImplicitPermissionsForUser` takes `e.m.Lock()`; `:155` `GetImplicitRolesForUser` takes `e.m.RLock()` and calls the same role-manager method below.
- `rbac_api.go:320` `GetNamedImplicitPermissionsForUser`: calls `rm.GetImplicitRoles`, `GetFieldIndex`, then only reads `e.model["p"][ptype].Policy` and deep-copies matches.
- `model/model.go:425` `GetFieldIndex`: its cache write is guarded by `assertion.FieldIndexMutex`.
- `rbac/default-role-manager/role_manager.go`:
  - `:410` `GetImplicitRoles` → `:269` `getRole`: for a name not in the graph it `Store`s a new role in `allRoles` (a `sync.Map`). Only when `matchingFunc != nil` does it also `addMatch` on other roles. bkn-safe registers no matching function on `g` (`keyMatch` is used on `obj` only).
  - `:301` `removeRole` deletes it again.
  - `:31` `Role`: every field is a `*sync.Map`.

So the only writes on the path are `sync.Map` operations and a mutex-guarded cache, the same path casbin itself already runs under `RLock` via `GetImplicitRolesForUser`. `TestConcurrentDecisionsAndWrites` runs decisions for accessors absent from the role graph concurrently with writes under `-race`.

PR #1562, opened 2026-09-14
fix(bkn-backend): hide references to object types the caller holds nothing on

sh00tg0a1:
## What Changed

- [x] bkn-backend: a reference to an object type the caller holds no effective operation on is no longer shown through relation types, action types, concept groups or navigation shells.

## Why

- Related Issue: #1532 (part 2; part 1 is #1557 in Context Loader)
- Background: relation type, action type and concept group reads checked only the resource being read. A caller allowed to read a relation type saw its endpoints — ids everywhere, names/icons and mapping-rule property display names in lists and details — even when they held nothing on an endpoint object type. Concept groups listed and counted every member; navigation shells were opened and counted by such rows. Studio shows this today, and after #1557 Context Loader's `get_kn_detail` would too.

## How

Rule: an object type may be named to a caller through another resource only if the caller holds at least one effective operation on it (`permission.VisibleReferencedObjectTypes`) — the same projection parent navigation already uses.

- **Relation types** (`ListRelationTypes`, `GetRelationTypesByIDs`, `SearchRelationTypes`): readable only if both endpoints pass the rule. Lists and search drop them before paging, so `total_count` and pages count only what the caller sees; a by-id read is refused (403), all or nothing like the existing permission check. Concept group details, the export view, object type details and the Cypher schema inherit this through the list.
- **Action types** (same three methods): readable only if the **bound** object type passes the rule. Affected object types and impact contracts are deliberately not checked: ontology-query reads action types with the caller's identity to run them, and an action is also how a caller is let write to data they cannot read.
- **Concept groups**: `object_type_ids` and statistics count only members passing the rule, and only relation/action types the caller may read. The group list resolves one member scope per page (one relation-type read, one action-type read, three batched authorization checks), independent of the number of groups.
- **Navigation**: relation and action type rows hidden by the rule no longer open a navigation shell or count in its statistics.
- "Any effective operation" rather than `view_detail`: ontology-query reads relation and action types with the caller's identity for subgraph, Cypher and action execution; a caller granted `query_data` alone on an endpoint keeps working.
- Authorization failures are returned as errors, never read as "nothing visible".
- Breaking changes: none for callers with access to the whole network. Behaviour change: a relation type granted without its endpoints (or an action type without its bound object type) is no longer visible to that caller; with Professional deny rules, relation/action types touching a denied object type disappear with it.

## Testing

- [x] Unit tests passed locally (`go test ./...`, 47 packages; `-race` on the touched ones). Each new test fails with its filter disabled.

Test notes (test deployment; binary built from the deployed commit + these four commits):

Temporary user with no roles, granted on one network: `view_detail` on object type *customer*, `query_data` only on *lead*, relation types *lead→customer* and *customer→contact*, action types bound to *lead* (affects another type) and to *activity*, and one concept group of 7 members.

| Read | Before | After |
|---|---|---|
| Relation type list | both, incl. *contact* name | only *lead→customer* (`query_data` endpoint kept) |
| Relation *customer→contact* by id | 200 | 403 |
| Action type list | both | only the one bound to *lead* (delegated affect kept) |
| Action bound to *activity* by id | 200 | 403 |
| Concept group | all 7 member ids, stats 7/13/2 | 2 member ids, stats 2/1/1 |
| Navigation statistics | rt 2, at 2 | rt 1, at 1 |

- Admin on 5 networks: relation / action / concept group lists, statistics and navigation statistics byte-identical before and after.
- A pre-existing restricted account with mixed child grants: its concept group now lists only the member it can see (1/0/0 instead of 3/2/0); a network where it can see both endpoints and bound types is unchanged.
- End to end with #1557: `get_kn_detail` (REST and MCP, summary/full) returns only the permitted object types, relation types, action types and group members.

## Risk & Rollback

- Potential risks: one extra batched authorization call per list/detail/search of relation and action types (object types, any-operation projection), plus one definition read per restricted network with visible relation/action rows during navigation.
- Rollback plan: revert the four commits.

## Additional Notes

- Not verifiable on the test deployment: `query_data`-only execution through ontology-query (actions are unavailable there, `OntologyQuery.Proxy.Unavailable`); covered by unit tests on the by-id reads ontology-query uses.


sh00tg0a1:
On the two open review items:

1. Navigation authorization growing with the number of restricted networks — addressed in #1563: definitions are still read per network, but the referenced object types are now checked for all networks in one call.
2. Concept group members counted with any effective operation while the group detail lists object types by `view_detail` — intentional. Member ids and `OtTotal` are references and follow the reference rule; the detail's object types are full definitions and follow `view_detail`. The network level already works this way: navigation statistics count object types with any operation, the object type list filters by `view_detail`. Only a `query_data`-only member is counted but not listed.

PR #1563, opened 2026-09-14
perf(bkn-backend): check navigation references for every network in one call

sh00tg0a1:
## What Changed

- [x] bkn-backend: the navigation reference check introduced in #1562 makes one authorization call for all restricted networks instead of one per network.

## Why

- Related Issue: #1532 (follow-up to #1562, raised in its review)
- Background: when listing networks, every network the caller reaches only through its children has its visible relation and action types checked against the object types they reference. #1562 did that check network by network — one bkn-safe call each — so the network list (Studio's network page, Context Loader's `list_knowledge_networks`) got slower with every network a caller had scattered grants in.

## How

- Relation and action type definitions are still read per network, where they live (local database reads).
- The object types they reference are collected across networks and checked in a single `FilterResources` call through the new `permission.VisibleReferencedObjectTypesByKN`.
- `VisibleReferencedObjectTypes` is now the single-network case of it, so the rule has one implementation.
- Breaking changes: none; results are unchanged.

## Testing

- [x] Unit tests passed locally (`go test ./...`, 47 packages; `-race` on the touched packages).
  - `TestNavigation_ReferencesAcrossNetworksCostOneAuthorizationCall` lists two restricted networks and asserts one object type authorization call (it failed with 2 before this change) and the same filtering result.
  - `TestVisibleReferencedObjectTypesByKN_OneCallForEveryNetwork` covers ids shared across networks, duplicates, empty references and networks without references.
- [ ] Verified in test environment — not deployed: the test deployment's images are being switched to the official main build separately; this change will ride along with the next main image.

## Risk & Rollback

- Potential risks: the single call carries the object type ids of all restricted networks; it goes through the same chunked filter as the other navigation checks.


sh00tg0a1:
On the open item about holding every restricted network's definitions until the function returns: the rows kept are only the relation and action types the caller can already see in those networks, so their number is bounded by the caller's grants rather than by network size. The same function already holds the child resource candidates of all restricted networks at once (`ListKNChildResourceCandidates` is one call for all of them), and these definitions are a subset of those rows, so this adds no new order of memory. No change planned.

PR #1572, opened 2026-09-14
fix(execution-factory): require a tool to belong to the toolbox in the route

sh00tg0a1:
# Pull Request

## What Changed

- [x] Execution factory: every toolbox route that names a tool now loads it through one helper, `selectBoxTool`, which requires the tool to belong to the toolbox named in the route. This covers tool detail, update, debug and proxy execution on the public, internal and caller-scoped faces, plus `ExecuteToolCore`, the entry MCP servers built from toolbox tools use.
- [x] A tool stored under another toolbox is answered exactly like a missing one: same status, code, description and details.

---

## Why

- Related Issue: none
- Background: permissions on these routes are checked on the toolbox in the route, so the tool addressed through that route should be one of that toolbox's own. Batch delete and batch status changes already select by `(box_id, tool_id)`. The single-tool routes selected by tool id alone. This aligns them, the same way the definition read route in #1567 already checks the tool against its toolbox.

---

## How

- `selectBoxTool(ctx, boxID, toolID)` loads the tool by id (`t_tool.f_tool_id` is unique, and a tool never changes toolbox). It reports the tool as absent when its `BoxID` differs from the route's. The five call sites keep their existing not-found handling unchanged.
- Audited and already correct, so unchanged: batch delete, batch status change, tool lists, release info, export, capability index reconcile, `SearchTools` (filters by `(box_id, tool_id)` pairs), the MCP call and debug routes, and `/app/:mcp_id/*` (tools resolved by name within the addressed server).
- Breaking changes: none for requests that address a toolbox's own tools.
  - An MCP server created with `creation_type=tool_imported` must name each tool with its own toolbox. That is what the UI and API produce.
  - Optional pre-upgrade check for existing records:
    `SELECT m.f_mcp_id, m.f_box_id, m.f_tool_id, t.f_box_id FROM t_mcp_tool m JOIN t_tool t ON t.f_tool_id = m.f_tool_id WHERE m.f_box_id <> t.f_box_id;`

---

## Testing

- [x] Unit tests passed locally: `go vet ./...` and the CI package set of the execution factory. The new `tool_box_scope_test.go` covers all five paths, same toolbox and other toolbox. It asserts the not-found answer and that nothing is read, written or proxied for a tool of another toolbox. The tests fail with the toolbox comparison removed.

Test notes (test deployment). The execution factory deployment ran this branch for the run and was then rolled back. The temporary account held `view`, `execute` and `modify` on one toolbox only.

| Request by the temporary account | Result |
|---|---|
| Detail, execute (`proxy`), debug and update of a tool in its toolbox | detail 200; execute returns the tool's result |
| Same four requests naming a tool id stored under another toolbox | 400 `ToolNotFound` "工具不存在", byte-identical to a request with a non-existent id; the other toolbox's tool is unchanged afterwards |

The temporary account was removed afterwards.

---

## Risk & Rollback

- Potential risks: an MCP server whose stored `(box_id, tool_id)` pair does not match would stop resolving that tool (see the query above). Normal creation paths do not produce such pairs.


sh00tg0a1:
On the open item about stored MCP tool pairs: the test deployment has 33 rows in `t_mcp_tool` and none whose `(f_box_id, f_tool_id)` pair disagrees with `t_tool` (the query from the PR description returns 0). Registration already builds each pair from a toolbox tool read through `GetBoxTool`, and after this change that read refuses a mismatched pair, so no new ones can be written. I would keep configuration generation failing loudly on such a row rather than skipping the tool silently: a server that quietly loses a tool is harder to notice than one that refuses to deploy. The pre-upgrade query stays in the description for installations that want to check first.

PR #1574, opened 2026-09-14
feat(trace): expose authorized interaction artifacts

leecky:
## What Changed

- [x] Add an internal, optional reader capability for full interaction question/result artifacts.
- [x] Verify caller scope, interaction ownership, artifact reference, and artifact type before returning text.
- [x] Add focused regression coverage for complete result retrieval and unrelated-artifact rejection.

## Why

- Related Feature: BKN Trace business provenance source-text fidelity.
- Background: the EE provenance panel must render a recorded original only when Core can retrieve the authorized artifact, rather than presenting a shortened summary preview as original text.

## How

- The existing enterprise interaction reader exposes an optional in-process capability only; no public HTTP, MCP, or SDK contract changes.
- Artifact references are first matched to the selected interaction summary, then loaded under the trusted technical scope.

## Testing

- [x] Unit tests passed locally: `go test ./src/driveradapter/api/httphandler -count=1`
- [ ] Verified in test environment

## Risk & Rollback

- Potential risks: EE must run with this Core capability to show complete source text; otherwise it correctly retains a preview without claiming it is original.
- Rollback plan: revert this commit; the optional capability is additive and has no public contract effect.

## Additional Notes

- Supports EE PR #75 follow-up for lossless historical provenance snapshots.


leecky:
/review

Fixed the interaction-vs-artifact authorization mismatch identified in review. The optional source-text reader now resolves a referenced artifact through `GetArtifactForInteraction`, retaining the already-authorized interaction scope. The regression test proves it no longer takes the direct artifact path.

PR #1577, opened 2026-09-14
fix(trace): omit nil source artifact content

leecky:
## What Changed

- Treat a referenced artifact with no content as unavailable source text.
- Add a regression test that prevents the literal string `"null"` from appearing as interaction source text.

## Why

Business provenance must not present a missing recorded artifact as a user input or business answer.

## How

`fullArtifactText` now returns an empty value before serializing a nil artifact content value. The existing authorized reader therefore omits the optional source-text fields.

## Testing

- [x] `go test ./src/driveradapter/api/httphandler -count=1`

## Risk & Rollback

- Low risk: affects only nil content, which now follows the existing unavailable-source behavior.
- Rollback: revert this single commit.

## Additional Notes

- Follow-up to #1574 and EE #75.


leecky:
/review

Please review the nil-content source-artifact guard. It is a focused follow-up to #1574; the new test first reproduces the erroneous `"null"` text and passes after the guard.

PR #1588, opened 2026-09-15
fix(trace): avoid mixed collations in agent exclusion

leecky:
## Summary
- separate non-ASCII agent display names from ASCII identity exclusions
- retain exact ascii_bin identity semantics without converting indexed columns
- add MariaDB regression coverage that verifies the actual UTF-8 connection collation

## Verification
- sessionstore package tests pass
- targeted MariaDB integration test compiles and skips locally when no DSN is configured


leecky (commented):
Review conclusion: approved. The exclusion predicate correctly confines non-ASCII values to agent_name while retaining exact ascii_bin semantics for identity fields. Unit coverage is included; local sessionstore tests pass, and the MariaDB integration path is covered by CI.

rongwei-liu (approved):

PR #1603, opened 2026-09-16
fix(authz): preserve delegated object grants

rongwei-liu:
## Summary

- attribute new delegated object grants to the authenticated grantor
- preserve delegated permissions after the grantor loses authorization, while preventing future unauthorized changes
- add atomic source-level batch revocation with matching audit semantics
- keep official 0.1.4 legacy grant migration behavior

## Testing

- go test ./internal/authz ./internal/httpapi -count=1
- git diff --check

Related to openbkn-ai/bkn-studio#691


rongwei-liu:
Superseded by #1604 because this branch name did not satisfy the repository branch-name policy.

PR #1604, opened 2026-09-16
fix(authz): preserve delegated object grants

rongwei-liu:
## Summary

- attribute new delegated object grants to the authenticated grantor
- preserve delegated permissions after the grantor loses authorization, while preventing future unauthorized changes
- add atomic source-level batch revocation with matching audit semantics
- keep official 0.1.4 legacy grant migration behavior

## Testing

- go test ./internal/authz ./internal/httpapi -count=1
- git diff --check

Related to openbkn-ai/bkn-studio#691


rongwei-liu:
Addressed the audit-provenance review finding in 5e8de10d5.

- single revocation now retains policy_source, authority_source, created_by, operation, effect, and via in _outcome
- batch revocation now records each selected source under grant_sources before deletion
- added single and batch audit-detail regression coverage

Validated with: go test ./internal/authz ./internal/httpapi -count=1

rongwei-liu:
Addressed the remaining P1 audit-capacity review in commit a6de16a41.

Changes:
- Fan out successful batch revokes into one audit row per grant, correlated by the same request ID.
- Preserve grant_id, policy_source, authority_source, created_by, operation, effect, via, and the exact removed status for every grant.
- Pre-encode and size-check both possible per-grant outcomes before any permission mutation.
- Append up to 500 audit entries atomically in one transaction while preserving the tamper-evident hash chain.
- Keep oversized request details as valid bounded JSON instead of truncating raw JSON bytes.
- Return exact per-ID removal results from RevokePolicies.

Verification:
- go test -p=1 ./internal/authz ./internal/audit ./internal/httpapi -count=1
- go vet ./internal/audit ./internal/httpapi
- golangci-lint run ./internal/audit ./internal/authz ./internal/httpapi: 0 issues
- go test -race ./internal/audit -count=1

kenlee1988 (approved):

PR #1643, opened 2026-09-17
feat(authz): enforce grantable operations

kenlee1988:
## What Changed

- [x] Add the `grantable` operation declaration to bkn-safe’s persisted authorization registry.
- [x] Reject `grantable: false` operations in authorization write paths while preserving the existing `*/*` break-glass grant.
- [x] Return every operation, including its `grantable` flag, from the authorization registry API.
- [x] Declare `catalog:view_summary` as non-grantable.

---

## Why

- Related Issue: openbkn-ai/bkn-studio#709
- Related Feature: Resource catalog authorization contract
- Background: authorization editors and write APIs need one backend-owned declaration of operations that may be persisted as grants.

---

## How

- Add a backward-compatible `operations.grantable` database field; omitted seed values default to `true`.
- Enforce the declaration in central policy persistence, HTTP/admin/object/proxy write paths, and registry consumers.
- Breaking changes (API, config, database, etc.): additive `grantable` response property and an auto-migrated non-null boolean column with default `true`.

---

## Testing

- [ ] Integration tests passed locally (not run; no external database required for this change)
- [ ] Verified in test environment (not run)

Test notes:

- `go test ./internal/seed ./internal/authz ./internal/httpapi ./internal/proxygrant ./internal/database`
- `golangci-lint run ./internal/seed/... ./internal/authz/...`
- `make api-docs-lint` (completed; repository has pre-existing warnings in unrelated API specifications)

---

## Risk & Rollback

- Potential risks: a caller attempting to persist a newly non-grantable operation receives a validation failure.
- Rollback plan: revert this PR; the database column is additive and defaults to `true`.

---

## Additional Notes

- `view_summary` is declared and write-gated here; its derived authorization decision is intentionally out of scope for this PR.


kenlee1988:
Confirmed scope: `grantable:false` is used only for newly introduced operations, currently `catalog:view_summary`; existing operations are never converted from `true` to `false`. No declared grantable operation requires a non-grantable operation.

Therefore the hypothetical existing-policy cleanup and post-normalization dependency cases cannot occur under this contract, so no code change is required for this PR.

/review

PR #1658, opened 2026-09-18
fix(deploy): publish proxy grant snapshots in migration

rongwei-liu:
## Summary
- Persist the published knowledge-network proxy grant-source snapshot during the 0.1.4 to 0.1.5 permission transition.
- Mark a proxy mapping ready only after its snapshot is written and verified against the synchronized authorization sources.
- Bundle the precompiled static Linux amd64 authorization migration executable for environments without Go.

## Tests
- python3 -m unittest -v test_bkn_data.py test_migrate.py
- GOWORK=off go test -p=1 ./...


rongwei-liu:
Addressed the review findings in 9d68d77c9:

- record the bundled Linux migrator as executable (100755) and add a regression test;
- report the persisted proxy snapshot version, matching f_published_model_version and f_synced_model_version;
- align the documentation and ignore rules with the checked-in executable.

Validated with Python migration tests (44 tests), Go tests, and the bundled executable help command. The authz_migrate source is unchanged since the binary's recorded build revision.

/review

PR #1680, opened 2026-09-20
docs(api): sync OpenAPI specs with the shipped responses and request fields

sh00tg0a1:
# Pull Request

## What Changed

Brief description of changes:

- [x] `docs/api/bkn/**` — knowledge network, capabilities and concept group specs
- [x] `docs/api/context-loader/**` — kn-explore and object-instance specs
- [x] `docs/api/bkn-safe/audit.yaml`, `docs/api/execution-factory/function.yaml`

Specs only. No code, config or schema changes.

---

## Why

- Related Issue: none — this came out of an ad-hoc contract audit rather than a tracked issue. Happy to open one retroactively if the module owners prefer.
- Related Feature: n/a
- Background: `docs/api/tools/api_contract_diff.py` run against a test deployment reported nine endpoints returning fields no spec declares; a static pass over the gin handlers found two request fields in the same state. Every item was then confirmed against the handler or the response struct.

---

## How

- Key implementation points:
  - **Responses**: `skill_content` on the knowledge network list/detail; `sources_available` on the capabilities list (prose described it, the schema did not); the capabilities **internal face** documented as its own path with `CapabilityReferenceList`, since it answers with bare references and the shared schema made `total_count` / `boxes` / `metadata_available` look mandatory there; `skills_total` / `functions_total` / `apis_total` / `mcp_tools_total` on concept group statistics (the knowledge-network copy of `Statistics` also lacked the last two, and its `functions_total` text predates the function/API split); `operations` on the group, object types and relation types, `module_type` on relation types and `branch` on `SimpleObjectType`; `chain_state` on audit logs; `effective_permissions` and `data_source_metadata_unavailable` on a Context Loader object type, `tags` and `comment` on a Context Loader concept group; `effective_permissions` on object instances.
  - **Requests**: `owner_id` accepted as a spelling of `box_id` on the capabilities list; `working_directory` on function execute, which also scopes artifact collection.
  - **Correction, not an addition**: `query_object_instance` takes and returns an opaque `cursor` string, while the spec still described a `search_after` array. The binding ignores unknown fields, so a caller following the old spec silently received page one again. `search_after` stays documented where it is still used (instance subgraph, action logs).
  - Dropped a duplicated `updater` entry from the knowledge-network `required` list.
- Breaking changes (API, config, database, etc.): none. Documentation catches up with behaviour that already ships; no endpoint changes.

---

## Testing

- [ ] Unit tests passed locally — n/a, no code changed
- [ ] Integration tests passed locally — n/a, no code changed

Test notes:

- `npx @redocly/cli lint --config .redocly.yaml` passes (exit 0) on all seven changed files with the pinned CLI from `npm ci`.
- `api_contract_diff.py --face ex --include-probe-post` against a test deployment: **0 gaps** over 92 compared read-only endpoints, down from 9 before this change.
- That run skipped three Context Loader endpoints, so they were probed by hand: `get_kn_detail` returns concept groups with `comment` / `tags` and object types with `effective_permissions`; `get_object_types` likewise; `query_object_instance` returns `cursor`, `datas`, `effective_permissions`, `total_count`.
- Paging correction verified directly: passing `cursor` returns the next rows; passing `search_after` returns the first page unchanged, which is what the old spec told callers to do.

Coverage gaps worth knowing: the checker never sends writes (110 operations), ten endpoints returned empty arrays so their element fields are still unobserved, and the execution-factory and bkn-safe surfaces are largely outside its probe scope. `working_directory` and the concept-group `operations` / `module_type` / `branch` additions were confirmed statically against the structs rather than observed live.

---

## Risk & Rollback

- Potential risks: documentation-only. The one behavioural claim a reader could act on is the `cursor` paging correction, which was verified live.

---

## Additional Notes

- A companion PR applies the same commit to `release/0.1.5`; `docs/api` is identical on both branches and every field documented here exists in the release line.
- Known and out of scope: `/api/capabilities-lab/v1` (35 routes) remains undocumented, as `docs/api/execution-factory/README.md` already states.


sh00tg0a1:
Fixed in the pushed commit — the blocking item was real, confirmed in the code and then against a live response.

`memberScope.apply()` (`adp/bkn/bkn-backend/server/logics/concept_group/member_visibility.go:115-127`) only assigns `OtTotal` / `RtTotal` / `AtTotal`, and the remaining counters have no `omitempty`, so they always serialize as `0` on this face. On the test deployment:

```
concept group g_retrieval  object_types_total 4, relation_types_total 4, action_types_total 0,
                           skills_total 0, functions_total 0, apis_total 0, mcp_tools_total 0
knowledge network detail   concept_groups_total 5, object_types_total 15, relation_types_total 45
```

The counters stay declared — they are on the wire, and dropping them would just move the problem to "returned but undeclared" — but each now states it is not calculated here and points at the knowledge-network detail, which owns those resources. The same note went onto `concept_groups_total`, `risk_types_total` and `metrics_total`: they have always been zero on this face, which predates this branch. The three counters that *are* calculated now describe what they count (visible members, relation types with both endpoints inside the group, action types bound to a visible member).

Also addressed from the same round:

- **Mount response flags** (the non-blocking note on #1681): confirmed. `AttachCapabilities` fills only `Entries` / `TotalCount` / `Boxes` (`capability_binding_handler.go:130-134`), so `metadata_available` and `sources_available` are always `false` there. Both fields now say they are meaningful on the list response only, and the mount response documents what it leaves unset.
- **`SimpleObjectType.branch`**: correct, `json:"branch"` has no `omitempty` (`interfaces/object_type.go:140`), so it is always present and belongs in `required`. Applied to all three copies of that schema — `concept-group.yaml`, `relation-type.yaml`, `action-type.yaml` — rather than only the one this branch had touched.

redocly lint passes on every changed file.

/review

PR #1684, opened 2026-09-20
docs(api): say that SimpleObjectType.branch always comes back empty

sh00tg0a1:
# Pull Request

## What Changed

Brief description of changes:

- [x] `docs/api/bkn/concept-group.yaml`, `relation-type.yaml`, `action-type.yaml` — one description, the three copies of `SimpleObjectType`

Specs only. One field description, no schema shape change.

---

## Why

- Related Issue: none — this closes the one remaining note from the review of #1680 / #1681, which were merged with it open as non-blocking.
- Related Feature: n/a
- Background: #1680 moved `SimpleObjectType.branch` into `required`, which is right at the level of "the key is always there" — the field has no `omitempty`. The review then pointed out that no producer ever assigns it, so a client generated from the spec treats it as a branch id it can use and receives an empty string.

---

## How

- Key implementation points: every construction site fills only `OTID`, `OTName`, `Icon` and `Color` (`relation_type_service.go:344,352,456,466,498,514,521,529` and `action_type_service.go:477,559,568`). The description now says the field is always present but unfilled today, and points at the enclosing resource for the branch. All three copies of the schema carry the same note.
- Breaking changes (API, config, database, etc.): none.

---

## Testing

- [ ] Unit tests passed locally — n/a, no code changed
- [ ] Integration tests passed locally — n/a, no code changed

Test notes:

- `npx @redocly/cli lint --config .redocly.yaml` passes (exit 0) on all three changed files with the pinned CLI.
- On a test deployment, a relation type inside a concept group answers with `"source_object_type": {"id": "document", "name": "文档", "branch": "", "icon": "", "color": ""}` — the key is present, the value is empty.

---

## Risk & Rollback

- Potential risks: documentation-only.

---

## Additional Notes

- The alternative would be to fill `branch` at the producers, which is a behaviour change and belongs in its own issue. This PR only stops the spec from promising a value that never arrives.


sh00tg0a1:
Both notes from the last round are fixed in the pushed commit — each one checked against the file and the wire.

- **Examples missing `branch`** (inline on `action-type.yaml:784`): correct. The two list-response examples printed `object_type` with only `id` / `name` / `icon` / `color`, which contradicted both the "always present" wording and the `required` list one schema below. They now carry `branch: ""`, which is what the service actually returns.
- **`AffectDetail` has no `branch` to go up to**: correct, and the old wording ("read it from the enclosing resource") dead-ends there. The description now names the carriers — the relation type or action type, or the request's `branch` query parameter — and says explicitly that a nested `affect.object_type` is two levels down, so go up to the action type.

redocly lint passes on all three changed files. Both branches carry the identical commit.

/review

PR #1692, opened 2026-09-21
test(row-filter): cover acceptance safety boundaries

rongwei-liu:
## Summary
- verify that a changed effective row-filter digest invalidates an existing object cursor before a downstream Vega read
- verify that a no-rows decision performs no resource read and exposes no total-count signal
- reject row-filter responses whose value set exceeds the documented 100-value maximum at both Foundry resolver boundaries

## Validation
- GOWORK=off go test ./logics/permission ./logics/cypher ./logics/object_data_stats ./driveradapters
- GOWORK=off go test ./logics/object_type ./logics/metric ./logics/knowledge_network ./logics/permission

Part of #1679. The remaining release-gate work requires EE management API completion and target-environment performance validation.


rongwei-liu:
/review

Rebased onto the current fixed-condition implementation and resolved the blocking finding. The consumer-side 100-value cap and its tests were removed, so a normalized bkn-safe plan is no longer rejected as a 503. The remaining service-level regressions now assert the precise invalid-cursor response and that Vega is not called; the FALSE-plan test also verifies no resource read or count signal.

rongwei-liu:
Superseded by #1754, rebased onto current main with the blocking consumer-side value cap removed.

PR #1696, opened 2026-09-21
feat(row-filter): add policy management socket

rongwei-liu:
## Summary
- add the Core-owned row-filter management socket and guarded route
- validate published exact-field capabilities through ontology-query
- provide chart configuration for the ontology-query validation upstream

## Validation
- go test ./extension/rowfilter ./internal/httpapi ./config
- go test ./logics/object_type -run TestCompileRowFilter
- go test ./driveradapters -run TestRowFilterCapability

Related to openbkn-ai/openbkn-ee#92


rongwei-liu:
Addressed the review findings in a2abb7442:

- corrected the ontology-query defaults, chart values, ConfigMap fallback, and local example to port 13018;
- made ConditionOperations mandatory only for string fields, because the existing model emits it only for string properties; mapped integer and boolean predicates retain exact scalar semantics;
- normalized the management caller department limit to the same Core default as the runtime read route;
- made the capability decoder forward-compatible with additive upstream fields.

Validation:
- go test ./logics/object_type -run TestCompileRowFilter
- go test ./driveradapters -run TestRowFilterCapability
- go test ./internal/httpapi ./config

/review

PR #1707, opened 2026-09-21
fix(context-loader): return committed interaction when question evidence cannot be stored

sh00tg0a1:
# Pull Request

## What Changed

Brief description of changes:

- [x] context-loader / MCP lifecycle: `bkn_start_interaction` returns the committed interaction ids when only the question artifact write fails with a retryable error, instead of reporting `trace_core_unavailable`
- [x] context-loader / MCP lifecycle: every Trace Core availability failure on the MCP path is logged with fixed `tool` / `stage` / `code` fields (new helper `lifecycleUnavailable`)
- [x] context-loader / evidence ingest: `postBatchWithRetry` resumes from the first undelivered event and stops retrying non-retryable rejections
- [ ] Docs: not applicable — no API, schema or configuration change

---

## Why

- Related Issue: Refs #1691 (this PR implements the Context Loader follow-up listed under the issue's "supplement" section; it does not close the issue, whose main scope is Trace Core)
- Related Feature: BKN Trace managed interaction lifecycle
- Background:
  `bkn_start_interaction` asks Trace Core to create the interaction and **then** writes the question artifact to the evidence store. When the evidence store was overloaded, the artifact write failed and the tool answered `trace_core_unavailable` / `retry_later` — although Core had already committed the interaction. The caller never received the ids. The retry it was told to make generates a new idempotency key, so:
  - in a continued conversation Core refuses it with `interaction_in_progress` (pointing at the orphan the caller cannot address);
  - in a new conversation it leaves one more orphaned active interaction per retry until the 5-minute lease expires.

  Agents reported managed calls stalling for 20–60 s in retry loops while the evidence store was overloaded. The same failures left no trace in Context Loader logs, because the MCP path returned lifecycle availability errors without logging them.

---

## How

- Key implementation points:
  - `lifecycle_adapter.go`: when `RecordInteractionArtifact` fails for the question, classify the error with the existing `lifecycleAvailabilityError`. Retryable (timeout, 5xx, transport) → log `[BKN Trace] lifecycle degraded: tool=… stage=question_artifact code=… interaction_id=…` and return the committed ids. Non-retryable (deployment defects such as `INGEST_AUTH_NOT_CONFIGURED`) → unchanged, still fails with `contact_platform_operator`.
  - The question artifact only feeds the evidence summary; nothing in the lifecycle (sessionsvc) reads it. The cost of degrading is that the trace view lacks the question text for that interaction.
  - `session_guard.go`: new `lifecycleUnavailable(ctx, tool, stage, err)` logs `[BKN Trace] lifecycle unavailable: tool=… stage=… code=…` and returns the same error payload as before. Used at every availability-error site on the MCP path (`ensure_operation`, `trace_correlation`, `ensure_conversation`, `read_interaction`, `answer_artifact`, `core_call`, `question_artifact`). Context Loader has no metrics pipeline (no MeterProvider), so these fixed-shape log lines are the countable signal.
  - `evidence.go`: `postBatch` delegates to a new `postEventsFrom(…, start)` that returns the index of the first event Core did not accept; `postBatchWithRetry` resumes from that index, and returns immediately on a non-retryable `CoreHTTPError` (4xx other than 429), matching `postArtifactWithRetry`.
- Breaking changes (API, config, database, etc.): none. Error payloads are unchanged; the only visible difference is that a start which previously failed with a retryable availability error after Core committed the interaction now succeeds.

---

## Testing

- [ ] Integration tests passed locally — not applicable: the module has no integration suite for the MCP lifecycle; covered by the environment verification below

Test notes:

- New regression tests, each confirmed to fail with the fix reverted and pass with it:
  - `TestStartInteractionReturnsCommittedIDsWhenQuestionArtifactStoreIsUnavailable`
  - `TestPostBatchWithRetryResumesAfterLastAcceptedEvent` (without the fix: `evt-1 evt-2 evt-1 evt-2 evt-3`)
  - `TestPostBatchWithRetryStopsOnNonRetryableRejection` (without the fix: 3 calls)
- Existing `TestStartInteractionReportsIngestConfigurationFailureWithoutRetryGuidance` still passes (non-retryable path unchanged).
- `go test ./...` in `adp/context-loader/agent-retrieval`: all packages pass. CI steps reproduced locally: `go build ./...`, `go build -tags ee_dev ./...`, `go vet`, `go test` and `go test -tags ee_dev` for `./server/extension/...` and `./server/driveradapters/mcp/`.
- Dev VM (v0.1.5 + this patch, arm64 k3s). Evidence store made unavailable by pointing `BKN_TRACE_EVIDENCE_INGEST_URL` at a closed local port:

  | step | released v0.1.5 image | patched image |
  |---|---|---|
  | `bkn_start_interaction` (continue) | `trace_core_unavailable`, interaction already committed | succeeds, returns ids |
  | agent retries start | `interaction_in_progress`, `current_interaction_id` = orphan | not needed |
  | business tool with the returned context | — | succeeds |
  | Context Loader log | nothing | `lifecycle degraded … stage=question_artifact`, `lifecycle unavailable … stage=answer_artifact` |

  After restoring the normal ingest URL, the full start → tool → finish flow on the patched image succeeds with no degradation lines. The patched binary's md5 was checked against the running process path. The VM was restored to the released image and configuration afterwards.

---

## Risk & Rollback

- Potential risks:
  - During an evidence-store outage, interactions are created without a question artifact, so the trace view shows no question text for them. This is logged per interaction.
  - Evidence batches rejected with a non-retryable 4xx are no longer retried (previously 3 identical attempts with the same result).
  - `bkn_finish_interaction` with `outcome=completed` still fails when the answer artifact cannot be written — Core requires the answer artifact reference. Out of scope here; tracked in #1691.
- Rollback plan: revert this PR or redeploy the previous agent-retrieval image. No schema, configuration or data changes.

---

## Additional Notes

- Screenshots, logs, documentation links, etc.:
  - EE picks this up through its regular upstream merge; the three changed source files are identical between core and EE.
  - Found during verification, unrelated to this change and reproducible on the released image: schema-snapshot evidence from `get_kn_detail` is rejected by Trace Core with `invalid_evidence_event` (payload hash mismatch), so any managed interaction that calls it ends `partial`. Will be tracked separately.


sh00tg0a1:
Thanks — both points are about where the retryable / non-retryable line sits. Addressed as follows:

**2. Plain errors classified as retryable (fixed)**

An ingest URL that does not end in `/events` leaves no artifact endpoint to derive, and `postArtifactWithRetry` returned a plain error, which the mapping treats as a transient outage. With this PR that meant every start silently degraded and the question artifact was never recorded.

`postArtifactWithRetry` now returns the sentinel `bkntrace.ErrEvidenceArtifactURLNotConfigured`, and `lifecycleAvailabilityError` maps it to `evidence_capture_failed` / `contact_platform_operator` (not retryable), the same class as `INGEST_AUTH_NOT_CONFIGURED`. New tests:

- `TestLifecycleAvailabilityErrorReportsMissingArtifactEndpointAsDeploymentDefect`
- `TestStartInteractionReportsUnderivableArtifactEndpointWithoutDegrading`

Both fail with the mapping removed and pass with it.

**1. Non-retryable branch still leaves the committed interaction (kept, by design)**

A non-retryable failure here is a deployment or code defect: missing ingest credentials, a denied ingest identity, an underivable artifact endpoint, or an event Core rejects. Every start fails the same way until an operator fixes it, and the caller is told `contact_platform_operator` with `retryable: false`, so it does not retry into the loop this PR removes. Degrading here instead would hide the defect from everyone except the logs, which reverses the existing decision covered by `TestStartInteractionReportsIngestConfigurationFailureWithoutRetryGuidance`.

The interaction left active in that case expires with its lease. Removing it entirely needs Core to accept a missing question artifact as a recorded partial reason, which is tracked in #1691.

